PR #25, opened 2017-04-27
Send custom headers with segment downloads

### LINKED ISSUE #20: [Feature Request] Pass mpd URL custom headers to actual video files?
I have a .strm file like this:

```
#KODIPROP:inputstream.adaptive.license_key=https://wvlic.brightcove.com/proxy/5375730767001|Content-Type=application%2Fx-www-form-urlencoded&X-Forwarded-For=202.89.4.222|A{SSM}|
http://dashtvnz-a.akamaihd.net/963482467001/963482467001_5375755467001_5375730767001.mpd?pubId=963482467001&videoId=5375730767001|X-Forwarded-For=202.89.4.222
```

For a GEO blocked service.

Here's the log:
https://pastebin.com/raw/T8jsHLXV

You can see it get's the license all OK etc.
However, it fails to open http://dashtvnz-a.akamaihd.net/963482467001/5375755266001/video/1/init.mp4

This will be due to CURL not setting the X-Forwarded-For (it only sets keep alive)

Would it be possible to update the code to pass on any custom stream headers to the actual video files?

I guess somewhere around here it would happen:
https://github.com/peak3d/inputstream.adaptive/blob/master/src/main.cpp#L217

glennguy:
Hi @peak3d 

This fixes #20. I'm *very* new to c++ so I don't know if this implementation is a good way of going about it particularly because it relies on a global variable, but it works nonetheless.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -191,12 +191,10 @@ bool adaptive::AdaptiveTree::download(const char* url, const std::map<std::strin
     return false;
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "seekable", "0");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip");
-  if (!manifestHeaders.empty())
+ 
+  for (const auto &entry : manifestHeaders)
   {
-    for (std::map<std::string, std::string>::const_iterator b(manifestHeaders.begin()), e(manifestHeaders.end()); b != e; ++b)
-    {
-      xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, b->first.c_str(), b->second.c_str());
-    }
+    xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, entry.first.c_str(), entry.second.c_str());
   }
 
   xbmc->CURLOpen(file, XFILE::READ_CHUNKED | XFILE::READ_NO_CACHE);
@@ -225,12 +223,10 @@ bool KodiAdaptiveStream::download(const char* url, const std::map<std::string, s
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "seekable" , "0");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Connection", "keep-alive");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip, deflate");
-  if (!mediaHeaders.empty())
+  
+  for (const auto &entry : mediaHeaders)
   {
-    for (std::map<std::string, std::string>::const_iterator b(mediaHeaders.begin()), e(mediaHeaders.end()); b != e; ++b)
-    {
-      xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, b->first.c_str(), b->second.c_str());
-    }
+    xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, entry.first.c_str(), entry.second.c_str());
   }
 
   xbmc->CURLOpen(file, XFILE::READ_CHUNKED | XFILE::READ_NO_CACHE | XFILE::READ_AUDIO_VIDEO);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -183,7 +183,7 @@ protected:
 Kodi Streams implementation
 ********************************************************/
 
-bool adaptive::AdaptiveTree::download(const char* url)
+bool adaptive::AdaptiveTree::download(const char* url, const char* manifestHeaders)
 {
   // open the file
   void* file = xbmc->CURLCreate(url);
@@ -191,6 +191,16 @@ bool adaptive::AdaptiveTree::download(const char* url)
     return false;
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "seekable", "0");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip");
+  if (strcmp(manifestHeaders, "") != 0)
+  {
+    std::vector<std::string> header, headers = split(manifestHeaders, '&');
+    for (std::vector<std::string>::iterator b(headers.begin()), e(headers.end()); b != e; ++b)
+    {
+      header = split(*b, '=');
+      xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, trim(header[0]).c_str(), header.size() > 1 ? url_decode(trim(header[1])).c_str() : "");
+    }
+  }
+
   xbmc->CURLOpen(file, XFILE::READ_CHUNKED | XFILE::READ_NO_CACHE);
 
   // read the file
@@ -208,7 +218,7 @@ bool adaptive::AdaptiveTree::download(const char* url)
   return nbRead == 0;
 }
 
-bool KodiAdaptiveStream::download(const char* url, const char* rangeHeader, const char* segmentHeaders)
+bool KodiAdaptiveStream::download(const char* url, const char* rangeHeader, const char* mediaHeaders)
 {
   // open the file
   void* file = xbmc->CURLCreate(url);
@@ -219,9 +229,9 @@ bool KodiAdaptiveStream::download(const char* url, const char* rangeHeader, cons
     xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Range", rangeHeader);
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Connection", "keep-alive");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip, deflate");
-  if (strcmp(segmentHeaders, "") != 0)
+  if (strcmp(mediaHeaders, "") != 0)
   {
-    std::vector<std::string> header, headers = split(segmentHeaders, '&');
+    std::vector<std::string> header, headers = split(mediaHeaders, '&');
     for (std::vector<std::string>::iterator b(headers.begin()), e(headers.end()); b != e; ++b)
     {
       header = split(*b, '=');
@@ -840,7 +850,7 @@ void Session::STREAM::disable()
   }
 }
 
-Session::Session(MANIFEST_TYPE manifestType, const char *strURL, const char *strLicType, const char* strLicKey, const char* strLicData, const char* strCert, const char* strSegmentHeaders, const char* profile_path)
+Session::Session(MANIFEST_TYPE manifestType, const char *strURL, const char *strLicType, const char* strLicKey, const char* strLicData, const char* strCert, const char* strManifestHeaders, const char* strMediaHeaders, const char* profile_path)
   : manifest_type_(manifestType)
   , mpdFileURL_(strURL)
   , license_key_(strLicKey)
@@ -928,8 +938,9 @@ Session::Session(MANIFEST_TYPE manifestType, const char *strURL, const char *str
     b64_decode(strCert, sz, server_certificate_.UseData(), dstsz);
     server_certificate_.SetDataSize(dstsz);
   }
-  if (*strSegmentHeaders)
-    adaptiveTree_->segment_headers_ = strSegmentHeaders;
+  
+  adaptiveTree_->manifest_headers_ = strManifestHeaders;
+  adaptiveTree_->media_headers_ = strMediaHeaders;
 
 }
 
@@ -1446,7 +1457,8 @@ extern "C" {
   {
     xbmc->Log(ADDON::LOG_DEBUG, "Open()");
 
-    const char *lt(""), *lk(""), *ld(""), *lsc(""), *sh("");
+    const char *lt(""), *lk(""), *ld(""), *lsc(""), *manh(""), *medh(""), *m_strURL;
+    m_strURL = props.m_strURL;
     MANIFEST_TYPE manifest(MANIFEST_TYPE_UNKNOWN);
     for (unsigned int i(0); i < props.m_nCountInfoValues; ++i)
     {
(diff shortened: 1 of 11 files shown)


src/main.cpp
```diff
@@ -35,6 +35,7 @@
 
 ADDON::CHelper_libXBMC_addon *xbmc = 0;
 std::uint16_t kodiDisplayWidth(0), kodiDisplayHeight(0);
+std::string pipe;
```
peak3d: For me it feels wrong to use header fields from manifest requestl for the underlying stream files.
Its great that it works for the example in issue #20, but there could be some other constellations in the future.

Additional Headers should be passed using a new listItem property and has to be kept in session or adaptive_tree_ instance. They should then be passed to AdaptvieStream::download as a parameter value.

Beside this singletons on global scope (pipe) are not longer supported in future binary addons because multiple instances of the addon can be opened (pls. see agile branch)

matthuisman:
@peak3d 
Your thinking something like

#KODIPROP:inputstream.adaptive.custom_headers=X-Forwarded-For=202.89.4.222

I assume adding a new parameter requires changing KODI code?

matthuisman:
Oh, or is it as easy as adding a new listitemprops to addon.xml.in
listitemprops="license_type|license_key|license_data|manifest_type|server_certificate**|segment_headers**"

Then adding a new argument to:

`Session::Session(MANIFEST_TYPE manifestType, const char *strURL, const char *strLicType, const char* strLicKey, const char* strLicData, const char* strCert, _const char* strHeaders_, const char* profile_path)`

matthuisman:
OK.
Did initial adding of new parameter

See here:
https://github.com/peak3d/inputstream.adaptive/compare/master...matthuisman:master

So, we have the segment_headers_ now in Session::Session

Now, will bool AdaptiveStream::download_segment()  be able to access this variable?
And pass it here: return download(strURL.c_str(), rangeHeader);

matthuisman:
I'm still not sure what is best name for all the variables.
It could be more than just headers - it's more like extra Curl parameters...

peak3d:
@matthuisman yes, exactly this way, there are no kodi changes needed as you already fugured out.
I'll look closer at it today.

matthuisman:
Do you think maybe KodiAdaptiveStream::download could be changed to just accept a URL & a map of headers.

`bool KodiAdaptiveStream::download(const char* url, std::map<string, string> &headers)`

The range header and segment_curl_option_headers could be set in the AdaptiveStream::download_segment() function.

Is there an existing method that takes a curl header string and converts to a map?
eg. "Content-Type=application%2Fx-www-form-urlencoded%2FX-Forwarded-For=202.89.4.222"
and returns {'Content-Type': 'application/x-www-form-urlencoded', 'X-Forwarded-For':202.89.4.222}

If we don't want to append the string to the URL and let KODI handle it - then I guess we need a way of parsing it our-self?

Unless curl addheader can accept that string of multiple headers?

src/main.cpp
glennguy: Sure thing. I've slowly managed to figure it out - I've got it to take the info from the listitem property, store in the session, pass the info to the adaptivetree_ instance so it can be passed to the download function in main. 

Working now on how to parse the curl options string

glennguy:
>My initial idea was allowing multiple headers to be added using the & to separate them (like the | method)

@peak3d Should we just have a string of header options that are delimited by '&' and the header values url encoded, just like kodi accepts after the pipe?

Also, should there be a 'pre-approved' list of headers eg. as there is in the Kodi source [here](https://github.com/xbmc/xbmc/blob/2966375f853bb01c3be571a892190f4a128c224b/xbmc/filesystem/CurlFile.cpp#L792) and [here](https://github.com/xbmc/xbmc/blob/c5eb848739f0bb34d2ef74c3354e27e78f69567b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp#L680)?

peak3d:
@glennguy yes, pass headers somehow separated through the new listitem property, pass the string at initialization time to the AdaptiveTree. AdaptiveStream has access to the AdaptiveTree and can pass it as string (const std::string &) to the download override methode.

In download you have to split (there is a split function in helpers.h/cpp) and loop over the list and call CURL_AddOption (making another split for key value there).

Depending if you want  to solve it by yourself or not, I have a little time tomorrow and can put it in if you want.

glennguy:
@peak3d Thanks for the help,  I think I should be able to solve it now. I'll push some changes, later today hopefully.

glennguy:
@peak3d @matthuisman Changes made - tested with the following .strm file:
```
#KODIPROP:inputstream.adaptive.segment_curl_option_headers=X-Forwarded-For=202.89.4.222&User-Agent=Mozilla/5.0%20%28X11%3B%20Ubuntu%3B%20Linux%20x86_64%3B%20rv%3A53.0%29%20Gecko/20100101%20Firefox/53.0
#KODIPROP:inputstream.adaptive.license_key=https://wvlic.brightcove.com/proxy/5375730767001|Content-Type=application%2Fx-www-form-urlencoded&X-Forwarded-For=202.89.4.222|A{SSM}|
http://dashtvnz-a.akamaihd.net/963482467001/963482467001_5375755467001_5375730767001.mpd?pubId=963482467001&videoId=5375730767001|X-Forwarded-For=202.89.4.222
```
I followed the example in wvdecrypter.cpp which made it quite easy in the end.

src/main.cpp
```diff
@@ -232,6 +219,15 @@ bool KodiAdaptiveStream::download(const char* urlPointer, const char* rangeHeade
     xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Range", rangeHeader);
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Connection", "keep-alive");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip, deflate");
+  if (strcmp(segmentHeaders, "") != 0)
+  {
+    std::vector<std::string> header, headers = split(segmentHeaders, '&');
```
matthuisman: If the string is encoded.. would you not be splitting on %26?

Maybe best to decode the entire string before this first split?
Then you don't need the decode in the for loop

src/main.cpp
```diff
@@ -831,12 +840,13 @@ void Session::STREAM::disable()
...
 }
 
-Session::Session(MANIFEST_TYPE manifestType, const char *strURL, const char *strLicType, const char* strLicKey, const char* strLicData, const char* strCert, const char* profile_path)
+Session::Session(MANIFEST_TYPE manifestType, const char *strURL, const char *strLicType, const char* strLicKey, const char* strLicData, const char* strCert, const char* strSegmentHeaders, const char* profile_path)
   : manifest_type_(manifestType)
   , mpdFileURL_(strURL)
   , license_key_(strLicKey)
   , license_type_(strLicType)
   , license_data_(strLicData)
+  , segment_headers_(strSegmentHeaders)
```
matthuisman: Do we need to set the variable segment_headers_ at all? (It's not actually used anywhere.)

Could we not just pass strSegmentHeaders direct to adaptiveTree?
adaptiveTree_->segment_headers_ = strSegmentHeaders;

matthuisman:
Still seems a bit odd having to define the same headers 3x times...

@peak3d 
Are you sure we couldn't maybe swap it to a inputstream.adaptive.stream_headers (or session_headers)
And it applies to the "whole" inputstream session (every curl call for that stream- Key, MPD, Segments).
So, no longer required on the the stream URL, or key URL.

Individual specific headers could still be used by the pipe method on the key (eg. form encoded) / mpd if needed.

JinRonin:
the motivation to allow custom headers should not be to bypass geo restrictions...

src/main.cpp
glennguy: Right you are. Removed.

src/main.cpp
glennguy: No. '&' is used as a delimiter to separate the keys. The values are what is encoded. 

CurlAddOption seems to take care of the decoding itself. The convention in the above example I posted is the same as what Kodi requires after the pipe from add-ons.

matthuisman: Oh. OK.
Sorry, i thought it was following the pipe convention (whole string URL encoded)

matthuisman:
@JinRonin 
Your right. However, there may be other good uses - eg: a required User-agent header.

glennguy:
@JinRonin up until recently Netflix has filtered certain User-Agent headers. Seeing that large amount of work that has recently gone into adding support to play Netflix content, and once the kodi-agile fork is merged and Kodi 18 is released the assumed following large amount of users taking advantage of it, there's a high likelihood that that Kodi User-Agent headers will be filtered on Netflix's end. 

The stream file posted by @matthuisman in issue #20 has merely been a convenient way to test the implementation of this feature. I don't believe there has been any explicit or implicit statement at least from myself that the motivation behind this PR has been to bypass geoblocking.

matthuisman:
I often find users who are in the correct country but use a VPN.
Therefore their IPs will be shown as another country.
By them setting the X-Forward-For header to their actual IP, they may be able to view the content for their actual country.

matthuisman:
Off-topic.
But does anyone else get Github notifications really late? 
Mine seem to come in 24-48 hours after the actual activity.

peak3d:
I'll look today over the changes, sry for late response...

peak3d:
> Are you sure we couldn't maybe swap it to a inputstream.adaptive.stream_headers (or session_headers)
@matthuisman: stream-key is already in the license_key template, therefore only manifest / media has to be decided.

Ok, lets start with an general Header key, wich is valid for both manifest + media (not license_key!!)
If we / you need a different control for headers between manifest and media, we'll define new keys for them wich overrides the general headers described above.

matthuisman:
@peak3d 

In the MPD download function, the URL might have headers in it (using |)
We would want these to over-ride any custom headers?

Therefore, we would need to set our custom headers before the URL is parsed.
I think we can do file.CURLAddOption before a file.CURLCreate.

DashTree and SmoothTree call download directly when getting the MPD file.
Feels strange editing these and adding stream_headers_ to their download call.

I suggest we have an override download function in AdapativeTree.cpp that accepts just a url and then calls download with the url and streamheaders?
That way, we don't need to modify the individual trees?

Here is some changes to look at (I havn't done the h files)
I also don't know how to compile the addons so it's just to maybe help out @glennguy  :)

https://github.com/peak3d/inputstream.adaptive/compare/agile...matthuisman:agile

Thoughts?

peak3d:
> In the MPD download function, the URL might have headers in it (using |)
> We would want these to over-ride any custom headers?

It was not expected that someone uses | for the mpd call, but for sure you are rugth, this is possible.
My opinion how we should proceed with the headers (on session initialization)

- if stream_headers listitem key is given:
 a.) parse this key value into mpd_header pairs, make a copy (media_header) 
 b.) remove the | separated part from mpd_url (mix is not supported)

- if stream_headers listitem key is NOT given:
a.) get the | separated header from mpd_url (if any) and parse it into mpd_header. media_header remains empty.

After that you have 2 parsed maps, one for tree, one for stream.
Pass those during the initialization to Tree and Stream so they can pass the header fields to download.

download then ONLY works with the extra passed headers, the | separated has been removed earlier.
IMO its always a good idea to choose one concept on a very early stage instead passing around multiple options.
After this change it should not feel strange to pass the "map" to the download function.

glennguy:
@peak3d sounds good, thanks. I'll get some changes made tonight.

glennguy:
@peak3d I think I've satisfied your comment above. 

mpd_url is always stripped of pipe and anything following it.

If there is no stream_headers listitem set, then if there is pipe/headers then it is used for the manifest download only

If there is a stream_headers listitem, then it is used for both manifest and media.

I've left the headers in string format as it seemed a lot easier passing it around, and the parsing is done in the main/download functions.

src/main.cpp
```diff
@@ -1468,6 +1501,15 @@ extern "C" {
       }
     }
 
+    if (strcmp(manh, "") == 0)
```
peak3d: This is very confusing for me, after 5 times reading I still don't understand why need 3 strings for this task.

IMO this is the place where:
- make a copy of props.m_strURL
- look if | is in string, if so:
- if inputstream.adaptive.stream_headers is not given, parse it into mahd
- remove the | part from the copy.

Parsing of a string for header parts should be done in a function (parseheader(std::map, const char *)

Doing the parsing stuff should be only done once (here!!), having it in download() means that we parse the same stuff every segment again, this is not very resource friendly.

src/main.cpp
```diff
@@ -1434,7 +1457,8 @@ extern "C" {
   {
     xbmc->Log(ADDON::LOG_DEBUG, "Open()");
 
-    const char *lt(""), *lk(""), *ld(""), *lsc("");
+    const char *lt(""), *lk(""), *ld(""), *lsc(""), *manh(""), *medh(""), *m_strURL;
```
peak3d: manh / medh are std::map<std::string, std::string>

glennguy:
@peak3d Thank you so much for the feedback, not being picky at all. My apologies that it seems as I don't really know what I'm doing with regards to a lot of the changes, I'm very much a beginner (less than a month) at c++.  I am keen to contribute but please say if it is unhelpful at my skill/experience level.

I think I've addressed the issues now. We're now using maps to pass the headers around, and there is a separate function in helpers.cpp to parse them.

JinRonin:
really, please dont add this feature unless there is a real issue.
if so it would be nice if you block the x-forwarded-for header.

matthuisman:
@glennguy @peak3d 

Maybe it can drop down to a single download function in main.cpp now?

peak3d:
@glennguy way of programming is different from dev to dev, its like starting a friendship where each other has to figure out how the other one's character is. I really appreciate that you go the way with me, the future is important and the more people contribute the better.
Thanx for your patience!

Pls. leave 2 download methds so far as we do not know what's happening in the future.
My feeling is, that we will be happy to have them separated in the future.

I'll look over the changes today, keep head up :-)

src/main.cpp
```diff
@@ -183,14 +183,22 @@ class AP4_DASHStream : public AP4_ByteStream
...
-bool adaptive::AdaptiveTree::download(const char* url)
+bool adaptive::AdaptiveTree::download(const char* url, const std::map<std::string, std::string> &manifestHeaders)
 {
   // open the file
   void* file = xbmc->CURLCreate(url);
   if (!file)
     return false;
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "seekable", "0");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip");
+  if (!manifestHeaders.empty())
```
peak3d: Not needed to request for empty() here.

If you are lazy in writing you could also do:
for (const auto &entry :  manifestHeaders)
{

src/main.cpp
```diff
@@ -208,17 +216,22 @@ bool adaptive::AdaptiveTree::download(const char* url)
...
   // open the file
   void* file = xbmc->CURLCreate(url);
   if (!file)
     return false;
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "seekable" , "0");
-  if (rangeHeader)
-    xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Range", rangeHeader);
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_HEADER, "Connection", "keep-alive");
   xbmc->CURLAddOption(file, XFILE::CURL_OPTION_PROTOCOL, "acceptencoding", "gzip, deflate");
+  if (!mediaHeaders.empty())
```
peak3d: pls. see previous comment

matthuisman:
Great work guys!
Looking forward to this :)

I'm guessing this change will only be in KODI 18.0 builds as this is on agile branch?
(Not back-ported to current 17.1?)

glennguy:
@peak3d all sorted, and changed to 'lazy' for loop. Seems a lot easier to read that way. Thanks for your patience and spoon feeding me some of it, it has helped me learn a lot.

@matthuisman this PR is against the master branch so its for 17.1. I imagine it will get pulled into the agile branch but @peak3d will decide that. I see you mentioned you haven't got a build environment set up for this, let me know if you need some help - I can build for some platforms.

peak3d:
Thanx @glennguy !

priyavadan:
sorry for reviving an old thread, so do we just need to define the stream_headers and that will apply the headers to both media and mpd? I am trying out the X-Forwarded-For passed to the stream headers.

PR #28, opened 2017-05-13
Fix seeking in timeshift

rbuehlma:
In case of re-aligning the segments, use the first PTS as starting
point.

Use double instead of float


rbuehlma:
Need to use different branch -> close this PR

PR #30, opened 2017-05-14
Do not seek past live stream

rbuehlma:
With this patch, Kodi will continue playing the live stream from its current position (minus ~12 seconds) instead of freezing the stream in case of seeking too far.

In my opinion,
caps.m_supportsSeek = session && !session->IsLive();
could with this patch be changed to
caps.m_supportsSeek = session;

Due to a conflict, this PR includes:
https://github.com/peak3d/inputstream.adaptive/pull/29


rbuehlma:
Rebased

rbuehlma:
Does not work anymore with current master. -> Closing PR

PR #32, opened 2017-05-16
change to automatic depends set on addon.xml

AlwinEsch:
Related to kodi Pull Request xbmc/xbmc#12029

Not merge in before is in Kodi!


AlwinEsch:
Can be merged in, if OK. Kodi include now related changes.

PR #35, opened 2017-05-22
Update on kodi to addon interface

AlwinEsch:
Related to xbmc/xbmc#12143


AlwinEsch:
What is with this and #33?

PR #62, opened 2017-09-06
Unload wv

GTechAlpha:
WV is not unloaded on exit.


peak3d:
There was a reason why I commented this part out, have to investigate.

PR #63, opened 2017-09-06
Recursively search for decrypters

GTechAlpha:
Does not find libssd_wv in the lib sub-directory of addon directory.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1486,15 +1486,20 @@ void Session::GetSupportedDecrypterURN(std::string &key_system)
 
   std::vector<kodi::vfs::CDirEntry> items;
 
-  for (std::vector<std::string>::const_iterator path(searchPaths.begin()); !decrypter_ && path != searchPaths.end(); ++path)
+  for (std::size_t p(0); !decrypter_ && p < searchPaths.size(); ++p)
   {
-    kodi::Log(ADDON_LOG_DEBUG, "Searching for decrypters in: %s", path->c_str());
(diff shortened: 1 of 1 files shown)


MilhouseVH:
Why not just move `libssd_wv` into the root where it is now expected (since about version 2.0.16), and ensure the addon is being packaged using the correct locations? There's no need to search recursively - if `libssd_wv` is where it's expected to be it either doesn't exist, or the addon isn't packaged correctly.

GTechAlpha:
~~LibreELEC package still places it in lib sub-directory~~.

I see that you updated the package - thanks for that. I think recursively searching would still be beneficial for possible future inconsistencies.

MilhouseVH:
I can see why you'd suggest that, but it's unnecessary complexity to resolve what would be a packaging error.

If you were to upgrade to the latest version of this addon, then `libssd_wv` should be found in two possible locations (`/usr/lib` on Linux, or the addon root) and nowhere else, in which case searching recursively is unnecessary complexity and overhead.

GTechAlpha:
The code is fairly simple. If the lib is found in one of the default search locations, the search stops, taking no extra time. If not, the search attempts to find it recursively.

peak3d:
Sorry, I see no reason for this and it could break things for users who store old libssd_wv's in a temp child folder

GTechAlpha:
That's fine. For the record, an old lib would only be used if a lib was not first found one of the default search directories.

PR #66, opened 2017-09-08
[RFC] [tryfix] segmented ttml

stefansaraev:
while working on an addon for local VOD provider, I noticed subtitles were not working at all.

they are referenced in an ism manifest as
```
<StreamIndex Type="text" Language="BUL" Name="textstream_BUL_CAPT" Subtype="CAPT" Chunks="61" QualityLevels="1" Url="QualityLevels({bitrate})/Fragments(textstream_BUL_CAPT={start time})">
```

fragment sample: http://sprunge.us/TeDJ

this is a quick (incomplete but works for me (tm)) attempt to adapt the TTML parser.

### DIFF
--- a/src/parser/TTML.cpp
+++ b/src/parser/TTML.cpp
@@ -74,6 +74,8 @@ start(void *data, const char *el, const char **attr)
           }
           if (strcmp(el, "br") == 0)
             ttml->m_strXMLText += "\n";
+          if (strcmp(el, "p7:br") == 0)
+            ttml->m_strXMLText += "\n";
         }
         else if (strcmp(el, "p") == 0)
         {
@@ -84,7 +86,11 @@ start(void *data, const char *el, const char **attr)
           {
             if (strcmp((const char*)*attr, "begin") == 0)
               b = (const char*)*(attr + 1);
+            if (strcmp((const char*)*attr, "p7:begin") == 0)
+              b = (const char*)*(attr + 1);
             else if (strcmp((const char*)*attr, "end") == 0)
+              e = (const char*)*(attr + 1);
+            else if (strcmp((const char*)*attr, "p7:end") == 0)
               e = (const char*)*(attr + 1);
             else if (strcmp((const char*)*attr, "xml:id") == 0)
               id = (const char*)*(attr + 1);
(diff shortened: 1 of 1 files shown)


stefansaraev:
removed an early reurn in TTML2SRT::Parse which caused subtitles to be rendered all over the screen.

src/parser/TTML.cpp
```diff
@@ -208,12 +208,9 @@ bool TTML2SRT::Parse(const void *buffer, size_t buffer_size, uint64_t timescale,
   XML_SetElementHandler(parser, start, end);
   XML_SetCharacterDataHandler(parser, text);
 
-  XML_Status retval = XML_Parse(parser, (const char*)buffer, buffer_size, done);
+  XML_Parse(parser, (const char*)buffer, buffer_size, done);
   XML_ParserFree(parser);
 
-  if (retval == XML_STATUS_ERROR)
```
peak3d: I don't understand this removal

stefansaraev: without this removal, (non-english) subtitles are rendered once they are parsed like this: http://imgur.com/a/wFkcO

I could not find the root cause, but removing the return fixes the issue

here is a [bulgarian](https://test.net.in/Fragments(textstream_BUL_CAPT=14400000000)) fragment sample (does not match the screenshot above but you get the idea) that broke

another [english](https://test.net.in/Fragments(textstream_eng_SUBT=14400000000)) sample that worked

stefansaraev: just to add, as a workaround, we adapted our addon to download external .dfxp subtitle file, convert it to srt and and tell the player to use load external subs. so this is not high prio. would be nice if you have an idea what's the root cause :)

peak3d: ok, thanks for providing samples!

stefansaraev: first commit is still needed. it makes (at least) english subs work. I can drop the second one for now if you want me to.

peak3d:
let me read first a little bit about expat / namespaces

stefansaraev:
@peak3d updated. moved the return where it belongs. should be ok now ?

stefansaraev:
better fix was pushed to master, I'll close this. thanks @peak3d

PR #68, opened 2017-09-09
[fix] set default TimeScale

stefansaraev:
I've got ism manifests where TimeScale parameter is omitted,
according to https://msdn.microsoft.com/en-us/library/ee673438 it is optional
and should default to 10000000 if omitted

this fixes an issue where it's impossible to seek with such manifests


peak3d:
thx @stefansaraev

PR #125, opened 2018-03-04
[appveyor] add WindowsStore 

Rechi:
@afedchin @notspiff FYI


afedchin:
x64-uwp and arm-uwp are enough.

Rechi:
removed x86-uwp

PR #127, opened 2018-03-16
added .gitattributes to avoid the issue with LE for patches and diffs.

afedchin:
see title.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# never modify line endings for patchs and diffs
+*.patch text eol=lf
+*.diff text eol=lf


peak3d:
What does this PR do exactly? I like to see in diff if I have CRLF line endings because I remove them.
If this PR hides the line endings issues, it would not match to my working behaviour.

afedchin:
This makes checkout diffs with LF ending on all platforms

Rechi:
@afedchin the issue is that expat's CMakeLists.txt has windows line endings, but the patch has linux line endings.

afedchin:
```
[ 14%] Performing patch step for 'expat'
patching file CMakeLists.txt
Assertion failed: p_base <= file_pos, file ../patch-2.5.9-src/pch.c, line 772

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
[ 14%] Performing configure step for 'expat'
```
As I saw configuring is done well, expat build fails because path step fails.

afedchin:
nvm, I got you

afedchin:
I've upload expat archive with fixed CMakeLists.txt line endings

Rechi:
> I've upload expat archive with fixed CMakeLists.txt line endings
If you want to upload a fixed version don't override the original file, have reverted it.

Rechi:
Just check in the second patch with windows line endings, like it was done for the first one.

afedchin:
@Rechi may be we should update the patch utility to avoid the similar issues in the future?  I have v2.7.5 installed with msys-git and it works well.

Rechi:
If you can find a patch executable which isn't linked against `msys-2.0.dll` or similar shared libs, do it.

afedchin:
What's wrong with shared libs?

Rechi:
I would prefer a version without shared libs, but as long as you package all required libs it's fine for me.

PR #133, opened 2018-03-31
Align streams after opening and seeking

rbuehlma:
This prevents a delay in the video or audio stream on playback start or after seeking

The issue is discussed here:
https://www.kodinerds.net/index.php/Thread/58635-PVR-Addon-f%C3%BCr-Zattoo-und-Reseller/?postID=454271#post454271

Is this a good way to solve this issue? Should it be possible to enable/disable this aligning in the settings?

While testing with Zattoo, this patch solved the issue and the stream start time was only marginally longer if noticeable at all.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -47,6 +47,7 @@
 #endif
 
 #define DVD_TIME_BASE 1000000
+#define MAX_STREAM_ALIGN_SECONDS 5
 
 #undef CreateDirectory
 
@@ -2133,6 +2134,43 @@ AP4_Movie *Session::PrepareStream(STREAM *stream)
   return nullptr;
 }
 
+void Session::AlignStreams()
+{
+  uint64_t maxDts = 0;
+  STREAM *res(0), *waiting(0);
+  for (std::vector<STREAM*>::const_iterator b(streams_.begin()), e(streams_.end()); b != e; ++b)
+  {
+    bool bStarted(false);
+    if ((*b)->enabled && (*b)->reader_ && !(*b)->reader_->EOS()
+    && AP4_SUCCEEDED((*b)->reader_->Start(bStarted))
+    && (*b)->reader_->DTS() > maxDts)
+      maxDts = (*b)->reader_->DTS();
+  }
+  
+  for (std::vector<STREAM*>::const_iterator b(streams_.begin()), e(streams_.end()); b != e; ++b)
+  {
+    bool bStarted(false);
+    if ((*b)->enabled && (*b)->reader_ && !(*b)->reader_->EOS()
+        && AP4_SUCCEEDED((*b)->reader_->Start(bStarted)) && (*b)->reader_->DTS() < maxDts) {
+      if (maxDts - (*b)->reader_->DTS() > (MAX_STREAM_ALIGN_SECONDS * DVD_TIME_BASE)) {
+        //Dont try to align more than MAX_STREAM_ALIGN_SECONDS seconds:
+        kodi::Log(ADDON_LOG_INFO, "Skip aligning stream %d due to more than %d seconds difference.",(*b)->info_.m_streamType, MAX_STREAM_ALIGN_SECONDS);
+        continue;
+      }
+      kodi::Log(ADDON_LOG_INFO, "Align stream %d %" PRIu64 " to %" PRIu64,(*b)->info_.m_streamType, (*b)->reader_->DTS(), maxDts);
+      int maxAlignIterations(1000);
+      while ((*b)->enabled && (*b)->reader_ && !(*b)->reader_->EOS()
+        && AP4_SUCCEEDED((*b)->reader_->Start(bStarted)) && (*b)->reader_->DTS() < maxDts) {
+          (*b)->reader_->ReadSample();
+          if (maxAlignIterations-- <= 0) {
+            break;
+          }
+      }
+    }
+  }
+
+}
+
 SampleReader *Session::GetNextSample()
 {
   STREAM *res(0), *waiting(0);
@@ -2708,7 +2746,11 @@ bool CInputStreamAdaptive::OpenStream(int streamid)
         stream->reader_->GetInformation(m_session->GetStream(m_IncludedStreams[i])->info_);
       }
   }
-  return stream->reader_->GetInformation(stream->info_);
+  bool ret = stream->reader_->GetInformation(stream->info_);
+  if (ret) {
+    m_session->AlignStreams();
+  }
+  return ret;
 }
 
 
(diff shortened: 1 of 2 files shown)


peak3d:
This PR workarounds an issue, it does not hit the root cause.

peak3d:
Can someone pls. provide (link) a debug log here?

rbuehlma:
Here you are:

https://gist.github.com/rbuehlma/f73d2c663e80a9439269ab2f6146f4e6

The log starts just before selecting the channel. In this case, the audio had a delay of about 2 seconds compared to the video stream.

rbuehlma:
Hi @peak3d do you have any news on this one? Unfortunately, the issue is still there.
Just to be clear:
Its only about the audio starting 1-3 seconds after the video. As soon as the audio is there, it is in sync with the video.

dagwieers:
@rbuehlma I can confirm this issue.

mafiwa2167:
@rbuehlma Maybe it's time to close this pr. @peak3d has no interest in fixing this because it doesn't concern him. It's his software so we should just accept his decision.

basrieter:
@mafiwa2167 I really think you should change your attitude! 

Many of the Kodi Devs do essential work to the product and related add-ons. However, we have not been cloned so there is only one of each of us to do work. And sometimes we prioritize other things. Without @peak3d  there would not be a lot of online streaming at all. So keep your negativity to your self.

mafiwa2167:
@basrieter Funny you consider yourself a "Kodi Dev" because you're the developer of an addon but that's off topic

I did not mean to be negative or disrespectful in any way. As I said, it's @peak3d's software so he can do whatever he want's with it and if he has no interest in fixing this, then we must accept that and be thankful for all the work he put into the addon and that he shares it with us for free.

I wasn't even complaining or anything so please stop putting words in my mouth.

rbuehlma:
While this change does fix the (minor) issue for me, it most probably is not the right way to solve it and probably more of a workaround. The idea of this PR is to help @peak3d to find a proper way of solving the issue.
Since the issue is not yet solved, there is no point of closing it from my side and probably the only one else allowed to do so, is @peak3d. Since he did not do it yet, there is still hope ;)

peak3d:
@rbuehlma can you pls. provide some steps for reproduction? Is maybe a .strm file already enough?
Or do I need to install one of your zattoo addons? If so, which one?

mafiwa2167:
You can use any stream hosted on nxload.com (hls) for example. Only to reproduce it.

That will also reveal an even bigger issue that the addon can lose the audio stream due to seeking and then the video stream starts to skip randomly by itself. If you get lucky, the audio might come back after a few skips but it also happens that the stream needs to be completely restarted.

peak3d:
@mafiwa2167 Can you try to create a .strm file with one of the not working streams pls?
Or does it need auth tokens when requesting it?

```
[HLS master playlist URL]
```

mafiwa2167:
Unfortunately this is the only example I could find right now. Usually I don't approve this kind of stuff and neither do you I guess but at least it's a safe and easy way to reproduce the issue.

You'll probably know how to get the m3u8 via the network browser. I think there's something like a session id in the URL else I'd get it for you.

https://nxload.com/embed/J2NID7rCiTI2/x.mkv

You can delete this commend after you saw the link.

peak3d:
@mafiwa2167 inputstream.adaptive cannot handle mkv files. Seems that you arrived another issue, which should be placed in forum.kodi.tv

mafiwa2167:
Thanks for taking the time to look into it.
Have you been able to reproduce the aforementioned issues with the example provided?

peak3d:
Just looked into the streams, issue is that video segment 1 is declared as 20s. duration, but has in reality 25 seconds included. This affects all subsequent segments leads to the fact, that seeking / starting at a position > 0 shows the described behaviour.
I'll look if something cam be changed here, but its quite clear that the hls is not correct written.

mafiwa2167:
Good to know. I wasn't aware of that. At least in Chrome and Firefox there are no issues at all with seeking. Audio and video will start with no delay right away from any seek position. Not sure though how they make it work in spite of the broken hls / segment lengths.

peak3d:
@mafiwa2167 your seek issues should be better now with the latest commits in master branch.
Same for @rbuehlma, but only when seeking it is fine now, startup is a different issue I'll have to look for separately.

mafiwa2167:
@peak3d Yes, no more issues there. I'm not sure if this is related to any of the previous now fixed issues, but inputstream won't even start playinig this file at all

https://jetload.net/p/vcxtgL6JTvdW (BigBuckBunny)

If it's not related, I can open a new issue.

peak3d:
> if it's not related, I can open a new issue.

@mafiwa2167 inputstream.adaptive is only capable playing master playlists, which contain multiple sub-playlists. The one you linked seems to ba a normal playlist and should be played with ffmpeg (not using inputstream.adaptive)

mafiwa2167:
@peak3d Sorry for the false alarm then. Even without inputstream.adaptive Kodi is not playing it but guess there's no point reporting it at the Kodi repo when ffmpeg is the cause.

rbuehlma:
@peak3d thank you very much. Seeking is indeed great now. Would be great if the startup works equally :) And would it be possible to cherry-pick the change to Leia branch or is the diff already too big?

rbuehlma:
Closing this PR since audio delay on channel startup and seek seems to be gone. Thank you

larena1:
@CastagnaIT do you think this can be fixed? Very often when seeing audio resumes quickly but video takes another 5 seconds or more to also resume. Maybe this occurs if audio and video segments are of different lengths but I'm not aware of any other player having these kinds of issues. Happens on YouTube for example 

Maybe seeking could also be made more accurate in that turn? Seems it's currently just starting playback from the closest segment which is not exactly accurate when you seek back 10 seconds because you missed something but segments are 15+ seconds.

CastagnaIT:
> do you think this can be fixed? Very often when seeing audio resumes quickly but video takes another 5 seconds or more to also resume. Maybe this occurs if audio and video segments are of different lengths but I'm not aware of any other player having these kinds of issues. Happens on YouTube for example

i dont think its right revive this Issue thread,
the symptom mentioned is part of a problem already known to Kodi core VP
which sometimes occurs because ISA manage packages as segments and when segments duration between a/v are different causes a misalignment
In the past, an issue had already been opened on xbmc repo about it and a patch had already been applied to mitigate the problem, but it cannot be completely solved

Furthermore, the current "seek" code in ISA needs to be reworked, there are some TODO in code,
which should take precedence over other tasks
atm i don't have time to recheck all this
if you wish, you can open a new Issue by providing a stream with which to replicate the problem

PR #136, opened 2018-04-02
[Win] Fix CDM library loading error if path contains unicode characters

AkariDN:
This PR fixes CDM library loading error on Windows if path contains unicode characters:
```
ERROR: AddOnLog: InputStream Adaptive: Unable to load widevine shared library (C:\Users\ユーザー\AppData\Roaming\Kodi\cdm\widevinecdm.dll)
```
Use UTF8->Unicode conversion.


peak3d:
Thanx!

PR #166, opened 2018-06-24
Moved the retry check

slaverman:


peak3d:
We're using blanks, and no tabs in Kodi and inputstream.adaptive.
Because of the blanks I cannot see what you have changed.

But pls. wait with spending more time correcting this, I'm already working at this place.

PR #217, opened 2018-12-27
HLS Streams can now use cookies returned from the manifest get reques…

ThorAsgardDev:
…t to get the decrypt keys

Useful because some HLS streams has the following scenario:
get request to retrieve the manifest => returns cookies (cookies returned change at each get manifest call)
get requests to retrieve decrypt keys => must use the cookies returned by the previous call to get the manifest

To use this new feature, the user has to set the "inputstream.adaptive.license_key" property of the playItem to something like "||manifest_cookies"

The implementation of this new feature add optional parameters to the download function that give the ability to return the cookies.
The HLS parser is modified to set the cookies to retrieve the keys if the user has set the property "inputstream.adaptive.license_key" to something like "||manifest_cookies"

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -444,6 +444,7 @@ namespace adaptive
 
 protected:
   virtual bool download(const char* url, const std::map<std::string, std::string> &manifestHeaders, void *opaque = nullptr, bool scanEffectiveURL = true);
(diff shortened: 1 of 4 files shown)


ThorAsgardDev:
Other addons GET manifest, then retrieves cookies, then pass these cookies (as manifest_headers) and the manifest URL to inputstream.adaptive. Then inputstream.adaptive get the manifest (again) and use the manifest_headers given by the user to get the keys.
In my use case this system doesn't work because the cookies returned by the GET request to get the manifest must be used to get the keys. If there is another call to get the manifest then the previous cookies are obsolet and the new ones must be used to get the keys.

PR #245, opened 2019-03-17
Simplify decision tree

dagwieers:
While reading this part of the code I was wondering why it wasn't made somewhat easier.

Also, why do we return an empty string for Video, and default to AAC for Audio ?

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -56,16 +56,16 @@ static void parseResolution(std::uint16_t &width, std::uint16_t &height, const s
 
 static std::string getVideoCodec(const std::string &codecs)
 {
-  if (codecs.empty() || codecs.find("avc1.") != std::string::npos)
+  if (codecs.empty())
     return "h264";
-  else if (!codecs.empty())
-  {
-    if (codecs.find("hvc1.") != std::string::npos)
-      return "hvc1";
-    else if (codecs.find("hev1.") != std::string::npos)
-      return "hev1";
-  }
-  return "";
+  else if (codecs.find("avc1.") != std::string::npos)
+    return "h264";
+  else if (codecs.find("hvc1.") != std::string::npos)
+    return "hvc1";
+  else if (codecs.find("hev1.") != std::string::npos)
+    return "hev1";
+  else
+    return "";
 }
 
 static std::string getAudioCodec(const std::string &codecs)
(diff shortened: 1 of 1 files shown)


gizmocuz:
I don't see any changes here ? The code was already quite simple.
Besides cleaning up some white spaces, the code already defaulted to AAC as well ?

dagwieers:
@gizmocuz Maybe check the diff in split-screen ? These are not just whitespace changes.

And I guestioned why we return AAC as default for audio, but an empty string as default for video. I did not say I made a change there.

gizmocuz:
@dagwieers , regarding the 'aac' change, it's just a whitespace.

Add ?w=1 behind the split diff like:

https://github.com/peak3d/inputstream.adaptive/pull/245/files?w=1

dagwieers:
@gizmocuz This patch is not about whitespaces. Sure, since we changed some structure we had to reindent one block, and sure I fixed one whitespace because there were two. But that is all relevant to what this patch DOES improve:
1. readability/complexity of something that is intrinsically more simple
2. avoid a second call to `empty()` in various cases

No need to show me my patch as I made those changes :-)

peak3d:
Audio currently has to return something != "" to support stream containers which include audio + video.
The code is somewhat grown, there will cleaner solutions but my time is too limited to retest all the different sream representation which arrived until today. Thx for the cleanup nevertheless, I'm fine with this.

dagwieers:
@peak3d If we do not have a test-suite with all possible examples, does such a project exist ? I would think of different video streams with video's showing the resolution/codec, and audio repeating the audio codec used. This way at least a visual/audible verification could be made to test end-user applications (as well as libraries).

peak3d:
@dagwieers no, it doesn't exist because many streams have only limited expiration date.
The only possibility I see would be completely downloading the underlying content.

Remaining issue: license expiration dates / license server accessibility. I have a bunch of streams here for testing, but oly a few are still working.

dagwieers:
@peak3d No, I mean a project that offers this for testing purposes, not actual content, test content. So if this doesn't exist it could be useful to create it. It could even help in automated testing if the streams would contain metadata that could be verified.

dagwieers:
I found these for now:
- https://bitmovin.com/mpeg-dash-hls-examples-sample-streams/
- http://ampdemo.azureedge.net/ams_samplestreams.html
- https://www.bbc.co.uk/rd/blog/2013-09-mpeg-dash-test-streams
- https://openidconnectweb.azurewebsites.net/Doc/Index?topic=test_streams
- http://ronallo.com/blog/testing-dash-and-hls-streams-on-linux/
- https://stackoverflow.com/questions/34343956/live-video-streaming-server-available-for-testing
- https://www.bento4.com/developers/dash/dash-sample-content/
- https://g33ktricks.blogspot.com/2016/04/list-of-hls-streaming-video-sample-test.html

PR #250, opened 2019-04-12
fixed ttml subtitles problem, peak3d/inpustream.adaptive#247

mediaminister:
fixed ttml subtitles problem on windows, fixes https://github.com/peak3d/inputstream.adaptive/issues/247:
- fixed sscanf problem on Windows with fraction after delimiter
- added support for frames in timeExpression


peak3d:
thx @mediaminister

PR #254, opened 2019-04-25
[cmake] remove usage of kodi-platform

olafhering:
The inputstream.adaptive binary addon does not use kodi-platform.
Remove existing references from debian/control.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,8 +1,7 @@
 Source: kodi-inputstream-adaptive
 Priority: extra
 Maintainer: peak3d <info@peak3d.de>
-Build-Depends: debhelper (>= 9.0.0), cmake, libkodiplatform-dev (>= 17.0.0),
-               kodi-addon-dev, pkg-config, libexpat1-dev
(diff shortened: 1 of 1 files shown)


peak3d:
Why p8 platform?

olafhering:
> Why p8 platform?

My grep can not know that `p8-platform/windows/dlfcn-win32.h` is only used on win32...

peak3d:
The dependency of p8 platform is only used to unique the dlXXX calls over platforms.
Same topic is already implemented in wvdecrypter, but without any additional dependency.

I see 2 ways to solve this topic properly (remove kodilatform and p8 platform):
- option 1: implement an own wrapper for win32 / dlopen, do not use any deps for this
- option 2: use https://github.com/dlfcn-win32/dlfcn-win32 instead p8 platform

Both options would require adding libtinyxml b.t.w.

What do you think?

olafhering:
> What do you think?

I am not familiar with all that, so can not really help with further changes.

peak3d:
kodi-platform was removed in a recent change

PR #256, opened 2019-04-27
Fix SegFault on SegmentURL handling

MisterD81:
While trying to create a Kodi addon for Rakuten I got several times a segmentation fault for some videos. It turned out that the MPD of the crashing ones was different then for the working ones.
In the crashing ones they use in SegmentURL elements both "media" and "mediaRange" attribute at the same time. 
Therefore I adjusted Segment structure to keep both attributes at the same time. In addition to that the range header handling was added for that case too.

Additional side-effect is that it fixes this issue with the Youtube addon:
https://github.com/jdf76/plugin.video.youtube/issues/595

I don't know if those audio streams are using the same structure like Rakuten but at least no I have sound there as well.

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -268,15 +268,13 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment *seg)
           download_url_ = current_rep_->url_ + download_url_;
       }
       else
-      {
         download_url_ = current_rep_->url_;
-        uint64_t fileOffset = seg != &current_rep_->initialization_ ? m_segmentFileOffset : 0;
-        if (~seg->range_end_)
-          sprintf(rangebuf, "bytes=%" PRIu64 "-%" PRIu64, seg->range_begin_ + fileOffset, seg->range_end_ + fileOffset);
-        else
-          sprintf(rangebuf, "bytes=%" PRIu64 "-", seg->range_begin_ + fileOffset);
-        rangeHeader = rangebuf;
-      }
+      uint64_t fileOffset = seg != &current_rep_->initialization_ ? m_segmentFileOffset : 0;
+      if (~seg->range_end_)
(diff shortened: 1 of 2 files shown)


peak3d:
@MisterD81 can you pls. attach the mpd? In general different mpd representations should be separated by flag (if possible) to reduce amount of memory kept by internal segment list.

MisterD81:
Sure. Please find attached one of the crashing MPD files:
[Sample-MPD.txt](https://github.com/peak3d/inputstream.adaptive/files/3132070/Sample-MPD.txt)

MisterD81:
@peak3d : Do you need further information for merging this PR or should I rework parts? Please just let me know. I really would like to see this integrated as it prevents Kodi from crashing when trying to play such streams.

peak3d:
PR makes sense looking at the manifest, I'll take a final look on in tomorrow

src/common/AdaptiveStream.cpp
```diff
@@ -268,15 +268,13 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment *seg)
...
-        if (~seg->range_end_)
-          sprintf(rangebuf, "bytes=%" PRIu64 "-%" PRIu64, seg->range_begin_ + fileOffset, seg->range_end_ + fileOffset);
-        else
-          sprintf(rangebuf, "bytes=%" PRIu64 "-", seg->range_begin_ + fileOffset);
-        rangeHeader = rangebuf;
-      }
+      uint64_t fileOffset = seg != &current_rep_->initialization_ ? m_segmentFileOffset : 0;
+      if (~seg->range_end_)
+        sprintf(rangebuf, "bytes=%" PRIu64 "-%" PRIu64, seg->range_begin_ + fileOffset, seg->range_end_ + fileOffset);
+      else
```
peak3d: This does not work for URLSegment representations with no byte range given. The complete block must be placed inside a condition if (~seg->range_begin_)

MisterD81: @peak3d : For sure this could happen. I added the requested changes.

PR #259, opened 2019-05-07
Fix issue where 1080p means 720p

### LINKED ISSUE #237: Auto stream selection always chooses 720p for 1080p Youtube streams
Fedora 29
InputStream Adaptive 2.3.12

Youtube streams are always choosing 720p even though 1080p is available.

Example video: https://www.youtube.com/watch?v=UA3jI71gy8E
720p60 is always chosen with "Auto" but using "Manual" I can choose the 1080p60 stream and it plays fine.

Netflix streams always pick 1080p with "Auto" set and play fine.

Debug log from playing example video using "Auto": https://gist.github.com/mooninite/90d5b30821a80f13db41cbdc4b68c941
MPD of example video: https://gist.github.com/mooninite/1864f7ccd2e08d1247f14383b2ce48f2

dagwieers:
It appears that the enumeration was not 100% the same in the settings causing 720p to be 640p, and 1080p to be 720p.

This fixes #237 and pietje666/plugin.video.vrt.nu#195

### DIFF
--- a/inputstream.adaptive/resources/settings.xml.in
+++ b/inputstream.adaptive/resources/settings.xml.in
@@ -20,8 +20,9 @@
             <options>
               <option label="30150">0</option> <!-- Max -->
               <option label="30151">1</option>  <!-- 480p -->
-              <option label="30153">2</option>  <!-- 720p -->
-              <option label="30154">3</option> <!-- 1080p -->
+              <option label="30152">2</option>  <!-- 640p -->
+              <option label="30153">3</option>  <!-- 720p -->
+              <option label="30154">4</option> <!-- 1080p -->
             </options>
           </constraints>
           <control type="spinner" format="string" />


peak3d:
very cool, thx @dagwieers !

dagwieers:
@peak3d Not to put any pressure on you, but is there a timeline when to expect this fix in the Kodi repositories?

Looking at the feedback we received wrt. 720p not working, and the issues I found that could also be in part related to this issue (or at least troubled users in understanding what is going on), I think we shouldn't hold onto this too long.

dagwieers:
I found these for now:
- https://bitmovin.com/mpeg-dash-hls-examples-sample-streams/
- http://ampdemo.azureedge.net/ams_samplestreams.html
- https://www.bbc.co.uk/rd/blog/2013-09-mpeg-dash-test-streams
- https://openidconnectweb.azurewebsites.net/Doc/Index?topic=test_streams
- http://ronallo.com/blog/testing-dash-and-hls-streams-on-linux/
- https://stackoverflow.com/questions/34343956/live-video-streaming-server-available-for-testing
- https://www.bento4.com/developers/dash/dash-sample-content/

dagwieers:
This fix is still not available on Raspberry Pi (i.e. LibreELEC). Not sure where to report it ?
Are binary modules the responsibility of the OS ?

dagwieers:
A new inputstream.adaptive v2.3.22.1 was released for ARM (Raspberry Pi) however it requires a reboot/restart of Kodi.

PR #260, opened 2019-05-08
Allow for intermittient failing of HLS playlist refresh download

glennguy:
Hi @peak3d 

Myself and a lot of other uses of @matthuisman addon https://github.com/matthuisman/plugin.video.kayo.sports have been having intermittient issues on typically a Friday and Saturday evening where the service will return HTTP 404 for the playlist, typically for up to 10 seconds. The service is using Akamai CDN but it's hard to know exactly what's going on but it would appear to be some sort of bottleneck between Kayo and Akamai.

At the moment when a playlist 404 happens the stream stalls and doesn't recover (you are just looking at a still image), attempting to seek will cause playback to end.

This fix stores a string copy of the playlist in the representation to pass in during prepareRepresentation in case of AdaptiveTree::download failing.

I've tested using a mitmproxy script to return 404 for the playlist (first 15 seconds of each minute) and had success.  Segment downloads and playback continue while the playlist is 404.

Happy for all/any feedback. I know this is somewhat of a workaround for (in this case) a content provider's buggy service but my reasoning is that if we retry on segment 404 then shouldn't that also be applied to a playlist refresh?

Thanks

Glenn

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -137,7 +137,8 @@ namespace adaptive
     {
       Representation() :bandwidth_(0), samplingRate_(0), width_(0), height_(0), fpsRate_(0), fpsScale_(1), aspect_(0.0f),
         flags_(0), hdcpVersion_(0), indexRangeMin_(0), indexRangeMax_(0), channelCount_(0), nalLengthSize_(0), pssh_set_(0), expired_segments_(0),
-        containerType_(AdaptiveTree::CONTAINERTYPE_MP4), startNumber_(1), nextPts_(0), duration_(0), timescale_(0), current_segment_(nullptr) {};
+        containerType_(AdaptiveTree::CONTAINERTYPE_MP4), startNumber_(1), nextPts_(0), duration_(0), timescale_(0), current_segment_(nullptr),
+	      unparsed_hls_rep_(""){};
       ~Representation() {
         if (flags_ & Representation::URLSEGMENTS)
         {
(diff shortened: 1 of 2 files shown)


src/common/AdaptiveTree.h
```diff
@@ -137,7 +137,8 @@ namespace adaptive
     {
       Representation() :bandwidth_(0), samplingRate_(0), width_(0), height_(0), fpsRate_(0), fpsScale_(1), aspect_(0.0f),
         flags_(0), hdcpVersion_(0), indexRangeMin_(0), indexRangeMax_(0), channelCount_(0), nalLengthSize_(0), pssh_set_(0), expired_segments_(0),
-        containerType_(AdaptiveTree::CONTAINERTYPE_MP4), startNumber_(1), nextPts_(0), duration_(0), timescale_(0), current_segment_(nullptr) {};
+        containerType_(AdaptiveTree::CONTAINERTYPE_MP4), startNumber_(1), nextPts_(0), duration_(0), timescale_(0), current_segment_(nullptr),
+	      unparsed_hls_rep_(""){};
```
gizmocuz: a std::string is automatically initialized by it's default constructor and can be removed here

peak3d:
@glennguy thx for bringing up this issue, but it feels wrong to copy / parse the previous playlist again as there has nothing changed in the existing playlist representation so far. It will be most likely only a flag what has to be set to re - fetch the failed update.

Pls. provide a full debug log, so I can figure out whats happening here

glennguy:
[kodi.log](https://github.com/peak3d/inputstream.adaptive/files/3157033/kodi.log)
@peak3d

peak3d:
@glennguy pls move this bracket:
https://github.com/peak3d/inputstream.adaptive/blob/master/src/parser/HLSTree.cpp#L453

down to line 465 and try again, thx

glennguy:
@peak3d that's perfect, thanks!

Shall I close this PR and allow you to make the change?

peak3d:
Already in inputstream.adaptive master branch

PR #297, opened 2019-08-08
add support for mpd location redirects

mediaminister:
Some streaming servers use MPD Location tag redirects for statistics or tracking purposes.
For instance:
```xml
<?xml version="1.0" encoding="utf-8"?>
<MPD xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="urn:mpeg:dash:schema:mpd:2011"
    xmlns:xlink="http://www.w3.org/1999/xlink"
    xsi:schemaLocation="urn:mpeg:DASH:schema:MPD:2011 http://standards.iso.org/ittf/PubliclyAvailableStandards/MPEG-DASH_schema_files/DASH-MPD.xsd"
    profiles="urn:mpeg:dash:profile:isoff-live:2011"
    type="dynamic"
    availabilityStartTime="1970-01-01T00:00:00Z"
    publishTime="2019-08-08T07:57:22"
    minimumUpdatePeriod="PT10S"
    suggestedPresentationDelay="PT10S"
    minBufferTime="PT10S">
    <Location>https://dcs-live.apis.anvato.net/server/play/kZEoafO6xdBCB8JJd/manifest.mpd?anvsid=xxxxxxxxxx-xxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx</Location>
</MPD>
```
This pull request implements this kind of redirect by reading the Location tag and using this value as mpdFileURL_.

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -33,6 +33,7 @@ using namespace adaptive;
 enum
 {
   MPDNODE_MPD = 1 << 0,
+  MPDNODE_LOCATION = 1 << 0,
   MPDNODE_PERIOD = 1 << 1,
   MPDNODE_ADAPTIONSET = 1 << 2,
   MPDNODE_CONTENTPROTECTION = 1 << 3,
@@ -245,6 +246,9 @@ start(void *data, const char *el, const char **attr)
 
   if (dash->currentNode_ & MPDNODE_MPD)
   {
+    if (dash->currentNode_ & MPDNODE_LOCATION)
+    {
+    }
     if (dash->currentNode_ & MPDNODE_PERIOD)
     {
       if (dash->currentNode_ & MPDNODE_ADAPTIONSET)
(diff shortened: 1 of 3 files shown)


dagwieers:
It would be nice, to see this one also in the upcoming release.

gizmocuz:
Just double checking this... 

```
  MPDNODE_MPD = 1 << 0,
  MPDNODE_LOCATION = 1 << 0,
```

Is this meant to be ? Both share the same value...

peak3d:
@gizmocuz oh, thats an issue and must be fixed, @dagwieers can you pls. add a PR with the next fr bit set?

dagwieers:
cc @mediaminister :-)

src/parser/DASHTree.cpp
```diff
@@ -1304,6 +1313,15 @@ end(void *data, const char *el)
         dash->currentNode_ &= ~MPDNODE_BASEURL;
       }
     }
+    else if (strcmp(el, "Location") == 0)
```
peak3d: Also this part should be checked identically as other parts, first check for set bit, and then for the string

peak3d:
:-) right, sorry @dagwieers

dagwieers:
I can do it, but I cannot test it. (no means of ARM/Windows development)

peak3d:
would be great. otherwise I would revert the merge

src/parser/DASHTree.cpp
mediaminister: Okay, I think I fixed this in https://github.com/peak3d/inputstream.adaptive/pull/303

peak3d:
@dagwieers it is fixed with #303

PR #298, opened 2019-08-08
add support for anvato livestreams

mediaminister:
This includes:
- add presentationTimeOffset and startNumber SegmentList attributes
- startPTS_ handling for SegmentList based streams
- This fixes https://github.com/peak3d/inputstream.adaptive/issues/288

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -134,7 +134,7 @@ namespace adaptive
     {
       Representation() :bandwidth_(0), samplingRate_(0), width_(0), height_(0), fpsRate_(0), fpsScale_(1), aspect_(0.0f),
         flags_(0), hdcpVersion_(0), indexRangeMin_(0), indexRangeMax_(0), channelCount_(0), nalLengthSize_(0), pssh_set_(0), expired_segments_(0),
-        containerType_(AdaptiveTree::CONTAINERTYPE_MP4), startNumber_(1), nextPts_(0), duration_(0), timescale_(0), current_segment_(nullptr) {};
+        containerType_(AdaptiveTree::CONTAINERTYPE_MP4), startNumber_(1), ptsOffset_(0), nextPts_(0), duration_(0), timescale_(0), current_segment_(nullptr) {};
       ~Representation() {
         if (flags_ & Representation::URLSEGMENTS)
         {
@@ -180,6 +180,7 @@ namespace adaptive
       unsigned int startNumber_;
       uint64_t nextPts_;
       //SegmentList
+      uint64_t ptsOffset_;
       uint32_t duration_, timescale_;
       uint32_t timescale_ext_, timescale_int_;
       Segment initialization_;
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -270,6 +270,13 @@ start(void *data, const char *el, const char **attr)
                 }
                 else if (strcmp((const char*)*attr, "media") == 0)
                 {
+                  if (dash->current_representation_->segments_.data.empty())
+                  {
+                    seg.startPTS_ = dash->base_time_ + dash->current_representation_->ptsOffset_;
+                    seg.range_end_ = dash->current_representation_->startNumber_;
+                  }
+                  else
+                    seg.startPTS_ = dash->current_representation_->nextPts_ + dash->current_representation_->duration_;
                   dash->current_representation_->flags_ |= DASHTree::Representation::URLSEGMENTS;
                   size_t sz(strlen((const char*)*(attr + 1)) + 1);
                   seg.url = new char[sz];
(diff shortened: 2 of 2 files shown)


dagwieers:
@peak3d Can we get this shipped in a release ? It is essential to get VTM GO live streams working in the add-on.

bigretromike:
It would be awesome to have.
That would resolve issue where you had to `seek(0)` player after stream start playing that is live-on to start it from beginning.

peak3d:
I'll look later on it, just back from holiday

peak3d:
I merged this PR but I also expect some regressions with DAZN / Eurosport live streams.
Because I have no account for any of them, we'll wait for comments from real users.

dagwieers:
@peak3d How would real users be able to test this before release?
Do we have binaries available, and a network of people we can swing into action?

peak3d:
@dagwieers There are builds available for the latest commit in github:
Click on the grean hook (or red cross) in commits section, if jencins was successful click on details.

This leads you here: https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/master/94/pipeline

and in Artifacts you'll find the build binaries.

dagwieers:
@peak3d Should the android/armv7 work on RPi3 ? It fails to install.

peak3d:
Not sure, I only tried for windows so far, and I know that it is not working for android.

mediaminister:
I'm beginning to think that this last "fix" broke Android. I got "Could not open / parse mpdURL" error.
But my original pull request seems to work. Can anyone confirm if 93 works on Android: https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/master/93/artifacts

dagwieers:
cc @michaelarnauts

dagwieers:
So what I feared would happen (https://github.com/peak3d/inputstream.adaptive/pull/298#issuecomment-522030541), is happening. We have no means to test the current master branch (android may be broken, rpi is untested) and if there is no guarantee that it works on all platforms, we cannot release this.

The is only a Windows binary that currently can be tested. So how do we get out of this situation?
And how can we prevent getting into this situation in the future? Can we set up a group of people that would like to help in testing before every new release so we can be release with confidence?

mediaminister:
I know, we have also no guarantee that all streams will work.
Maybe it's better to revert https://github.com/peak3d/inputstream.adaptive/commit/2e3aa2e132a2a84f3c3f333875a271a1670a7154 and https://github.com/peak3d/inputstream.adaptive/commit/55087d9da146bb4ddaacf3e69da250cadc255a56 and remove `Location` tag support for now because it breaks Android.
We have a workaround in Python for the Location tag.
Only releasing Anvato livestream support is also a nice step forward for add-ons that needs this.

michaelarnauts:
That's fine for me. We can look at the redirect later. We have a workaround anyway.

peak3d:
What is the issue with "Android". I use the current inputstream.adaptive master on many android devices without issues.

dagwieers:
> Not sure, I only tried for windows so far, and I know that it is not working for android.

I don't know, you tell me ;-)

mediaminister:
With MPD files with a Location tag I get "Could not open / parse mpdURL" error on Android.
I also found out that the MPEG-DASH spec defines multiple use cases for the Location tag. I only implemented a simple Location tag redirect which most likely breaks stream that contain a Location tag for other purposes.

dagwieers:
I think there is a confusion. When we were discussing that there was no (working) RPi build ad the Android ARM doesn't work for RPi, the remark was that it was not working for Android *(either)*.

So I assumed that the default Android builds simply did not (no longer) work on Android. So not just *this* build, but all builds.

peak3d:
@mediaminister I'll leave the PR's in until someone provides an URL / mpd which fails playing.

mediaminister:
I think I found something with a `Location` tag that's not intended to use for a simple redirect.
https://content.uplynk.com/channel/e11a05356cc44198977436418ad71832.mpd
license url: https://content.uplynk.com/wv

peak3d:
But the URL directs at least to the same / matching mpd

mediaminister:
It keeps redirecting in an endless loop without playing the actual content.

peak3d:
@mediaminister what was again the initial reason to use Location as redirect?

peak3d:
From what I read Location should be used to fetch the next mpd (for live stream manifet update), but in specification nowhere is written that it has to be used for media retrieval.

michaelarnauts:
@peak3d How does this interact with the `inputstream.adaptive.manifest_update_parameter` parameter? Since now that parameter specifies that the same manifest should be reloaded.

peak3d:
@michaelarnauts this is exactly what I thoght about in my latest commit, I guess that Location is exactly doing what the manifest_update parameter is thought for. This means that only "Full" is working correctly when the stream includes the Location tag. Otherwise there would be interferences between these two ways.

mediaminister:
@peak3d Streaming providers use the Location tag for statistics or tracking, by adding a query parameter to the original mpd url.

peak3d:
@mediaminister The query parameter could be also used to reduce the amount of segments when performing a manifest update.

PR #301, opened 2019-08-14
Fallback, assume TS container also on unknown extensions

L2501:
Currently HLS assumes TS container only for segments that have no file extension. Also assuming TS container for segments with unknown extension improves compability with more streams that have random extensions instead of .ts.  
Maybe this name check should be removed and container detected based on the file content.

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -379,8 +379,8 @@ bool HLSTree::prepareRepresentation(Representation *rep, bool update)
                 rep->containerType_ = CONTAINERTYPE_MP4;
               else
               {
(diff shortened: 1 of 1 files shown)


michaelarnauts:
Or based on mime type of the download headers?

peak3d:
@michaelarnauts  yes, a HEAD request was the intended solution but never found its way into the sources. I'll try to make an implementation for this soon before merging this PR.

L2501:
headers/playlists/filenames are not always accurate, and probing is used to make sure.  
 eg: https://github.com/video-dev/hls.js/blob/e7f9c9e656ac97b517b95016909ff6ce8ecf4553/src/demux/tsdemuxer.js#L66

PR #344, opened 2019-11-02
Add french translate

brunoduc:
Add french translate


peak3d:
Thx for contributing @brunoduc !

PR #358, opened 2019-11-30
Fixed and improved subtitle properties

CastagnaIT:
With the last changes made in branch, the properties of the subtitles have been broken, caused by a double identical condition, fixed this, i also introduced two new things:
- added "forced" property to adaptationset
this allow to set more than one property flag to a single subtitle stream (see screenshot, more properties splitted by comma)
- support to "impaired" property
allow to set a subtitle as for impaired people

image

Tested on Windows with latest Kodi 18 and NF add-on

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -596,8 +596,16 @@ start(void *data, const char *el, const char **attr)
               value = (const char*)*(attr + 1);
             attr += 2;
           }
-          if (schemeOk && value && strcmp(value, "subtitle") == 0)
-            dash->current_adaptationset_->type_ = DASHTree::SUBTITLE;
+          if (schemeOk && value)
+          {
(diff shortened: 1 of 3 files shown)


CastagnaIT:
What do you think about this PR? @peak3d

src/parser/DASHTree.cpp
```diff
@@ -719,24 +727,6 @@ start(void *data, const char *el, const char **attr)
           dash->strXMLText_.clear();
           dash->currentNode_ |= MPDNODE_BASEURL;
         }
-        else if (strcmp(el, "Role") == 0)
```
peak3d: What is the reason removing the Role element?

CastagnaIT: it is double, look line 584

peak3d: Yes, thx for finding / contributing

Issue #353, opened 2019-11-21
HLS with Widevine support? (Disney+)


matthuisman:
@peak3d 

Trying to playback some Disney+ streams and it appears they use HLS with widevine.

In the playlist:

`#EXT-X-SESSION-KEY:METHOD=SAMPLE-AES-CTR,KEYFORMAT="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed",KEYFORMATVERSIONS="1",URI="data:text/plain;base64,AAAAMnBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABISEGlJ4uAaTEQRh7SKUAR6QAY="`

SAMPLE-AES-CTR is apparently just CENC.
The full playlist & substream list is attached.

Kodi seems to download the segments but never plays.
I never see it even try to obtain the license from the license server
Get "Unsupported encryption method: "
So maybe first thing is to not error on EXT-X-KEY:METHOD=SAMPLE-AES-CTR.
then maybe I could at least manually first download the playlist, get the pssh data and send that as license_data.

[master_playlist.txt](https://github.com/peak3d/inputstream.adaptive/files/3872627/playlist.txt)
[substream_playlist.txt](https://github.com/peak3d/inputstream.adaptive/files/3872730/substream.txt)

Another thing may be looking into also supporting EXT-X-SESSION-KEY (in master playlist) in case the substream doesn't contain EXT-X-KEY
However, it appears Disney has both so not an issue here

matthuisman:
This may help
https://github.com/google/shaka-player/blob/c0f4f9da7b7ad5e8149cf8910ec62ab1cc1e0647/lib/hls/hls_parser.js#L1848

michaelarnauts:
I've read that Disney+ requires Widevine L1: https://www.androidpolice.com/2019/10/22/disney-will-only-work-on-devices-that-support-the-strictest-widevine-l3-drm/?amp (the URL indicated L3, but the article says L1.)

The Kodi implementation is L3 afaik, so I guess Widevine will refuse to continue.

yoshimo:
Is this a clone of https://github.com/peak3d/inputstream.adaptive/issues/96?

matthuisman:
Sample aes is different. This is cenc widevine encryption

glennguy:
@matthuisman https://github.com/peak3d/inputstream.adaptive/blob/master/src/parser/HLSTree.cpp#L429
[...]

PR #361, opened 2019-12-11
Non-byteranged fmp4 HLS

glennguy:
Allows for HLS playlists that have individual mp4 segments

Example stream here: https://bitmovin-a.akamaihd.net/content/MI201109210084_1/m3u8s-fmp4/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8

FFMPEG command to produce a playlist for testing:
`ffmpeg -i input.mp4 -b:v:0 1000k -b:v:1 640k -b:a:0 64k -map 0:v -map 0:a -map 0:v  -f hls -hls_playlist_type vod -hls_segment_type fmp4 -master_pl_name master.m3u8 -hls_segment_filename stream_%v/seg_%03d.mp4 -var_stream_map "a:0,agroup:aud_low,default:yes v:0,agroup:aud_low v:1,agroup:aud_low" stream_%v/out.m3u8`

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -86,9 +86,6 @@ namespace adaptive
       if (rep->flags_ & Representation::URLSEGMENTS)
         delete[] bs->url;
     }
-    if ((rep->flags_ & (Representation::INITIALIZATION | Representation::URLSEGMENTS))
-      == (Representation::INITIALIZATION | Representation::URLSEGMENTS))
-      delete[]rep->initialization_.url;
     rep->segments_.clear();
     rep->current_segment_ = nullptr;
   }
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -86,6 +86,9 @@ namespace adaptive
       if (rep->flags_ & Representation::URLSEGMENTS)
         delete[] bs->url;
     }
+    if ((rep->flags_ & (Representation::INITIALIZATION | Representation::URLSEGMENTS))
+      == (Representation::INITIALIZATION | Representation::URLSEGMENTS))
+      delete[]rep->initialization_.url;
     rep->segments_.clear();
     rep->current_segment_ = nullptr;
   }
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -265,14 +265,9 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment *seg)
     {
       if (current_rep_->flags_ & AdaptiveTree::Representation::URLSEGMENTS)
       {
-        if (seg != &current_rep_->initialization_)
-        {
-          download_url_ = seg->url;
-          if (download_url_.find("://", 0) == std::string::npos)
-            download_url_ = current_rep_->url_ + download_url_;
-        }
-        else
-          download_url_ = current_rep_->url_;
+        download_url_ = seg->url;
+        if (download_url_.find("://", 0) == std::string::npos)
+          download_url_ = current_rep_->url_ + download_url_;
       }
       else
         download_url_ = current_rep_->url_;
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -265,9 +265,14 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment *seg)
     {
       if (current_rep_->flags_ & AdaptiveTree::Representation::URLSEGMENTS)
       {
-        download_url_ = seg->url;
-        if (download_url_.find("://", 0) == std::string::npos)
-          download_url_ = current_rep_->url_ + download_url_;
+        if (seg != &current_rep_->initialization_)
+        {
+          download_url_ = seg->url;
+          if (download_url_.find("://", 0) == std::string::npos)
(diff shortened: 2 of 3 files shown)


glennguy:
@peak3d 
On a side note the HLS spec allows for more than one initialization segment to be specified in a playlist, usually used with embedded ads etc. This can be seen as an example in substream_playlist.txt in the OP of issue #353. 

Do you have any thoughts on how to restructure for this or how you would want to see it done?
Logically if you were to compare to a mpd with the same program material you would see the main content/ads in separate periods, however with this situation in HLS you're effectively getting different periods inside the representations, which I think won't work for us here.

I haven't looked in detail and not sure if it would be too complex for my abilities but my first thoughts are in this case to use a flag for each segment to signify if it's an initialization segment. Still treat the manifest as one continuous segment list but for the purposes of playback/seeking use the most recent initialization segment in the case of HLS.

gizmocuz:
@glennguy , thank you for the PR.
I have some streams with strange chunk names like .ts, but it is fragmented, then i have a stream that uses .m4s
Now in HLSTree.cpp there is a automatic detection is the containerType_ == CONTAINERTYPE_NOTYPE, but here it goes wrong for sure with .ts and there is no translation from .m4s to CONTAINERTYPE_MP4

To solve this, could you add at the bottom of the new "#EXT-X-MAP:" block the following line:

rep->containerType_ = CONTAINERTYPE_MP4;

I think it is safe to assume it will always be fragmented correct ?

glennguy:
@gizmocuz thanks for the heads up, I ran into a test stream with .m4s but completely forgot about it!

Yes seeing as you can specify whatever file extension you like when producing the content it's a great idea to force the container type, according to HLS spec the EXT-X-MAP tag guarantees fragmented segments. 

added in 27fa857

glennguy:
I've discovered my changes in AdaptiveStream.cpp have broken SegmentList based dash manifests. I'll find another way....

peak3d:
@glennguy if you're talking about VOD I would prefer multiple periods for the multi - init - issue.
Implementing other solutions would lead to lot of side - implementations (for example seek)

src/common/AdaptiveTree.cpp
```diff
@@ -86,9 +86,6 @@ namespace adaptive
       if (rep->flags_ & Representation::URLSEGMENTS)
         delete[] bs->url;
     }
-    if ((rep->flags_ & (Representation::INITIALIZATION | Representation::URLSEGMENTS))
-      == (Representation::INITIALIZATION | Representation::URLSEGMENTS))
-      delete[]rep->initialization_.url;
```
peak3d: Why is this removed?

glennguy: With only byteranged initialization support beforehand (no URLSEGMENTS) this code block would have never been executed (as far as I can see it is only called from HLSTree.cpp). Now with fmp4 urlsegments it deletes the initialization url from the representation causing the stream to fail to play.
Would you prefer instead to override the function in HLSTree?

glennguy:
@peak3d I have solved the multi init issue using multiple periods, I'll provide a PR soon :)

src/common/AdaptiveTree.cpp
glennguy: Also once the initialization url is set there should be no need to change or delete it, it should survive for the life of the representation. Further EXT-X-MAP tags will also be accompanied by EXT-X-DISCONTINUITY tags which mean new period/new representation.

peak3d: First this is the general implementation for all parser, and therefore should never modified because it is currently called only from HLS tree, but from what I read in your code (see new block #EXT-X-MAP)  the HLS initialization segment uses _url and allocates data, so it must be deleted[] to avoid memory leaks.
What is the real issue if the code abouve is not deleted?

glennguy: Ok I understand I shouldn't modify the general implementation to suit only one of dash/hls/ss, once we figure a way forward I will adjust accordingly.

From what I understand FreeSegments is there to clean up any allocated data on playlist refreshes before newSegments is swapped into rep->segments_. The real issue with the code in FreeSegments is that rep->initialization->url_ is being deleted pretty much straight after being set and then shortly after in [restart_stream()](https://github.com/peak3d/inputstream.adaptive/blob/master/src/common/AdaptiveStream.cpp#L224) the init download fails e.g in windows the log shows `2020-01-17 23:31:56.284 T:17816   ERROR: CWIN32Util::ConvertPathToWin32Form: Error converting path "ޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝޝa" to Win32 form` - we're reading garbage from memory for the url_.

Another way to put it - there's no 'newInitialization' that's being swapped into the rep after FreeSegments is called like in the case of newSegments. Is that a solution that you'd prefer to see btw?

It's not an issue for byteranged fmp4 as there is no rep->initialization->url_ needed or set.

peak3d: Thx for clarification @glennguy . I would prefer the newInitialization in this case.
If you assign it after the FreeSegments call you dont have to care about freeing the existing memory (in case you revert your before mentioned changes in AdaptiveTree)

glennguy: Thanks for your guidance, I think it's good now. Could you review the latest changes @peak3d?

PR #374, opened 2020-01-22
[HLS] EXT-X-DISCONTINUITY (multiple periods)

glennguy:
Hi @peak3d 

The #EXT-X-DISCONTINUITY tags are handled this way:
* Close off the current representation (freeSegments, swap newsegments etc.) and set period duration
* Add a new period (re-use master manifest to set up) if there isn't one already from preparation of other representations
* use the index of our current rep/adp to ensure we keep working on the same rep in the new period

Apologies for the large diff.

With HLS a 'period' differs from DASH since it is defined within a representation/adaptationset, therefore each period should contain an identical amount of adaptationsets with an identical amount of representations. So my solution is to copy the adaptationset/representations to each new period.

I thought the easiest and safest way to deep-copy the periods is to just re-parse the master manifest. To avoid downloading the manifest each time I split the functions, producing the pretty sizable diff. 

To help you read it, in processManifest the changes are at the top of the funtion (no changes after `std::map<std::string, std::string> map;`)

A small test strm file:
```
https://stream.test.theplatform.com/nbc/cmaf/master-24fps.m3u8
```
Seeking works well with fmp4, however seeking in TS seems to seek to the PTS of the TS stream i.e if there's a 25 minute program with a 5 minute ad inserted at 10:00, seeking to 17:00 will land the player at 22:00 (17:00 of the show that we're watching but not what would be expected behavior I assume)

Thanks for your time

Glenn

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -179,6 +179,7 @@ public:
     static const uint16_t ENABLED = 256;
     static const uint16_t WAITFORSEGMENT = 512;
     static const uint16_t INITIALIZATION_PREFIXED = 1024;
+    static const uint16_t DOWNLOADED = 2048;
 
     uint16_t flags_;
     uint16_t hdcpVersion_;
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -315,7 +315,9 @@ bool HLSTree::prepareRepresentation(Representation *rep, bool update)
     if (!effective_url_.empty() && download_url.find(base_url_) == 0)
       download_url.replace(0, base_url_.size(), effective_url_);
 
-    if (download(download_url.c_str(), manifest_headers_, &stream, false))
+    if (rep->flags_ & Representation::DOWNLOADED)
+      ;
+    else if (download(download_url.c_str(), manifest_headers_, &stream, false))
     {
 #if FILEDEBUG
       FILE *f = fopen("inputstream_adaptive_sub.m3u8", "w");
@@ -545,16 +547,6 @@ bool HLSTree::prepareRepresentation(Representation *rep, bool update)
           }
         }
       }
-      if (discont_count)
-      {
-        periods_[discont_count]->duration_ = pts - newSegments[0]->startPTS_;
-        overallSeconds_ = 0;
-        for (auto p : periods_)
-          overallSeconds_ += p->duration_ / p->timescale_;
-      }
-      else
-        overallSeconds_ = newSegments[0] ? (pts - newSegments[0]->startPTS_) / rep->timescale_ : 0;
-
       if (!byteRange)
         rep->flags_ |= Representation::URLSEGMENTS;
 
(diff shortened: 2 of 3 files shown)


peak3d:
Seeking logic for .ts streams should work also for .ts streams if period durations are fine.
Session::SeekTime should subtract 10+5 minutes from requested seek time (what leeds to 2 minutes seek in third period). Have you checked that duration in period is really set properly?

glennguy:
The seeking logic is fine - Session::SeekTime is working correctly. I've checked the duration_ in all periods and it's correct.

I think the issue is in TSReader::SeekTime where m_pkt.pts of the main program material is always going to be the pts we're seeking to less the duration of any ads that have been inserted into the playlist. This leads to inputstream.adaptive downloading a whole bunch of segments trying to find the pts value it's after (all segments from 17:00 to 22:00)

This also makes it impossible to seek into an ad period (but I can't imagine too many complaints)

glennguy:
I've also discovered an issue with my code with calling freesegments when there's psshsets added to periods (aes-128 streams). I'll try and get that fixed.

peak3d:
do you have a .ts sample strm file?

peak3d:
> I've also discovered an issue with my code with calling freesegments when there's psshsets added to periods (aes-128 streams). I'll try and get that fixed.

ok, just take care that refcount is updated properly....

glennguy:
I think I made an error when cleaning things up for the PR - it was working correctly before. I'll fix then supply a strm file as it affects it

glennguy:
aes-128 crash is now resolved, FreeSegments was being called on the incorrect period.

Here is a sample .ts stream. The link will be valid for ~24 hrs, it will 404 after that. I'll post another tomorrow if you haven't had time to look at it. You'll likely need a VPN to Australia.
```
https://sbsvoddai-vh.akamaihd.net/i/svod/SBS/managed/2020/01/31/667805_,1500K,1000K,512K,128K,.mp4.csmil/master.m3u8?hdnea=st=1581505159~exp=1581508759~acl=/i/svod/SBS/managed/2020/01/31/667805_,1500K,1000K,512K,128K,.mp4.csmil/*~id=7a70e2fb-7484-4a87-9a94-87c25f2d5b37~hmac=2968961e9531a5f8fc6874271c6970b1a993eced269ce48641c2f6c3fc249c4c&originpath=/ondemand/hls/content/2488267/vid/1686006851742/SIN/streams/e1bb3220-8cca-46a6-8783-a8e1f63b549c/master.m3u8
```

Another thing - with this stream the ad segments have no file extension and don't play (as in issue #301) so I modified in the containertype block:
```
            else
              {
                rep->containerType_ = CONTAINERTYPE_INVALID;
                continue;
              }
```
becomes
```
             else
                rep->containerType_ = CONTAINERTYPE_TS;
```

Thanks
Glenn

glennguy:
@peak3d new strm file:
```
https://sbsvoddai-vh.akamaihd.net/i/svod/SBS/managed/2020/01/12/665001_,128K,1000K,512K,1500K,.mp4.csmil/master.m3u8?hdnea=st=1581566423~exp=1581570023~acl=/i/svod/SBS/managed/2020/01/12/665001_,128K,1000K,512K,1500K,.mp4.csmil/*~id=b5d678fd-6c11-4ec4-8fdc-c4c2c4434e61~hmac=8f542b2b7ee30b60ec99c91baab7fc7ea756687d0d299f66640f92fb10c65053&originpath=/ondemand/hls/content/2488267/vid/1666166851665/SIN/streams/6562828b-872c-4532-b7ff-30ab1bcff34f/master.m3u8```

PR #375, opened 2020-01-22
Fix for manual stream changing

glennguy:
Hi @peak3d

I noticed in my testing for #374 that manually choosing another stream during playback causes a crash (m_IncludedStreams[2] for eg. would be 1007) - this PR addresses that.


glennguy:
To be clear this only affects streams with an includedstream

And here is a strm file to test: ( .ts segments)
```
https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_16x9/bipbop_16x9_variant.m3u8
```

peak3d:
Yes, this one makes absolutely sense, thx much!

PR #382, opened 2020-01-31
Ensure seeking to zero does not wrap around

glennguy:
Hi @peak3d 

reader_->GetPTSDiff can return a negative number and seek to or near zero will cause the adaptivestream to download every segment available in the tree (I think it must be looking for the right PTS?), playback pauses until this has finished and hangs on pressing the stop key.

I'm not sure this is an issue for fmp4 but definitely for ts streams.

Stream used for testing:
```
https://bitmovin-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8
```

You can reproduce by starting the stream and pressing down (seeking backwards by pressing left doesn't allow you to easily seek to zero)

Thanks to @matthuisman for finding and investigating this issue

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2754,10 +2754,10 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
   {
     seekTimeCorrected += timing_stream_->stream_.GetAbsolutePTSOffset();
     int64_t ptsDiff = timing_stream_->reader_->GetPTSDiff();
-    if (ptsDiff < 0 && seekTimeCorrected + ptsDiff < seekTimeCorrected)
-      seekTimeCorrected += ptsDiff;
-    else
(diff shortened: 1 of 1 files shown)


matthuisman:
Test builds here (requires latest kodi nightly):
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/PR-382/1/artifacts

Will try hopefully today or tomorrow :)

matthuisman:
I think this may have broke seeking forward... Every-time I try to seek forward, it keeps restarting at 00:00

glennguy:
Whoops, fixed now, thanks @matthuisman

matthuisman:
Nice! just tested and seems rock solid :D

peak3d:
Thx much @glennguy

PR #401, opened 2020-03-17
Filled #30161 (greek)

Twilight0:
& fill blank strings with the same values to avoid errors on check scripts


Twilight0:
Sorry for those mistakes in the last PRs, I had rebased my fork on the wrong commit.

dagwieers:
You don't have to close a PR. If you push your changes to the branch that is tracked by the PR, these updates will automatically be visible in this PR. So simply make the changes, force-push and refresh your page.

dagwieers:
There is one translation missing at the end. Apparently it was only added to the English translation recently.

Twilight0:
> There is one translation missing at the end. Apparently it was only added to the English translation recently.
Indeed, just filled it and I reverted that last string to its former translation. I got confused.

PR #402, opened 2020-03-18
Various translation fixes

dagwieers:
This PR includes:
- Translations for strings that were expected to be identical
- Add missing strings to most languages
- Translate untranslated strings for Dutch
- Remove BOM from Russian translation
- Fix syntax errors in French translation
- Remove unused/incorrect English `msgstr` entries


dagwieers:
@brundoc @peak3d There seems to be an issue related to the English source to contain a different `msgid` and `msgstr`.

Where most translations have translated the `msgid` (as is custom), the French translation has translated the `msgstr` instead. This is problematic. I have fixed the English translation file, but we need to re-translate the French translation as well.

dagwieers:
This one needs to be merged last (after all translations) as it will cause merge conflicts because of untranslated (and missing) strings.

PR #421, opened 2020-04-15
FIX: HLS fails on first EXT-X-KEY if not supported and doesn't check any others

matthuisman:
The below playlist will fail because the first EXT-X-KEY is SAMPLE-AES
Even though it has a valid key method (widevine) also.

This PR changes the behavior to instead look for VALID methods and fallback to invalid.
If a valid isn't found, it logs invalid and keeps trying.

It also makes a lot easier to add more supported methods.
Much easier checking for what you support, than checking for ever possible case you dont.

eg.
INFO: Unsupported encryption method: SAMPLE-AES
INFO: Unsupported encryption method: SAMPLE-AES
INFO: Supported encryption method found:SAMPLE-AES-CTR

Without the supported message, it appears it's failed.
IT also makes it easy to spot in a log when a stream is encrypted but no method is found
INFO: Unsupported encryption method: SAMPLE-AES
INFO: Unsupported encryption method: SAMPLE-AES
(no supported message)

It also means in the future, you just add in supported methods.
Not have to keep trying to add in non-supported methods.

There is also a commit to check KEYFORMAT before METHOD as the actual below playlist uses SAMPLE-AES instead of SAMPLE-AES-CTR for the widevine. KEYFORMAT is therefore a more reliable way to check.
```
#EXT-X-KEY:URI="skd://itunes.apple.com/p65435462/c1",KEYFORMAT="com.apple.streamingkeydelivery",KEYFORMATVERSIONS="1",METHOD=SAMPLE-AES
#EXT-X-KEY:URI="data:text/plain;charset=UTF-16;base64,vgEAAAEAAQC0ATwAVwBSAE0ASABFAEEARABFAFIAIAB4AG0AbABuAHMAPQAiAGgAdAB0AHAAOgAvAC8AcwBjAGgAZQBtAGEAcwAuAG0AaQBjAHIAbwBzAG8AZgB0AC4AYwBvAG0ALwBEAFIATQAvADIAMAAwADcALwAwADMALwBQAGwAYQB5AFIAZQBhAGQAeQBIAGUAYQBkAGUAcgAiACAAdgBlAHIAcwBpAG8AbgA9ACIANAAuADMALgAwAC4AMAAiAD4APABEAEEAVABBAD4APABQAFIATwBUAEUAQwBUAEkATgBGAE8APgA8AEsASQBEAFMAPgA8AEsASQBEACAAQQBMAEcASQBEAD0AIgBBAEUAUwBDAEIAQwAiACAAVgBBAEwAVQBFAD0AIgBBAEEAQQBBAEEATwBZAEQAUgBuAGQAagBNAFMAQQBnAEkAQwBBAGcASQBBAD0APQAiAD4APAAvAEsASQBEAD4APAAvAEsASQBEAFMAPgA8AC8AUABSAE8AVABFAEMAVABJAE4ARgBPAD4APAAvAEQAQQBUAEEAPgA8AC8AVwBSAE0ASABFAEEARABFAFIAPgA=",KEYFORMAT="com.microsoft.playready",KEYFORMATVERSIONS="1",METHOD=SAMPLE-AES
#EXT-X-KEY:URI="data:text/plain;base64,AAAAOHBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABgSEAAAAAAD5ndGYzEgICAgICBI88aJmwY=",KEYFORMAT="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed",KEYFORMATVERSIONS="1",METHOD=SAMPLE-AES-CTR
```

yes, i will need to squash all the commits :)

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -89,6 +89,8 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
   if (map["METHOD"] == "NONE")
   {
     current_pssh_.clear();
+
+    Log(LOGLEVEL_INFO, "Supported encryption method found: NON-ENCRYPTED");
     return ENCRYPTIONTYPE_CLEAR;
   }
 
@@ -119,7 +121,8 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
       if (bufLen == 50)
         current_defaultKID_ = std::string(reinterpret_cast<const char*>(&buf[34]), 16);
     }
-    Log(LOGLEVEL_INFO, "Supported encryption method found: %s", map["METHOD"].c_str());
+  
+    Log(LOGLEVEL_INFO, "Supported encryption method found: WIDEVINE");
     return ENCRYPTIONTYPE_WIDEVINE;
   }
 
@@ -133,7 +136,8 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
       current_pssh_ = baseUrl + current_pssh_;
 
     current_iv_ = m_decrypter->convertIV(map["IV"]);
-    Log(LOGLEVEL_INFO, "Supported encryption method found: %s", map["METHOD"].c_str());
+
+    Log(LOGLEVEL_INFO, "Supported encryption method found: AES-128");
     return ENCRYPTIONTYPE_AES128;
   }
 
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -85,61 +85,61 @@ HLSTree::~HLSTree()
 
 int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::string>& map)
 {
-  if (map["METHOD"] != "NONE")
+  // NO ENCRYPTION
+  if (map["METHOD"] == "NONE")
   {
-    // WIDEVINE CHECK
-    if (!map["URI"].empty() && !map["KEYFORMAT"].empty() && map["KEYFORMAT"] == "urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed")
-    {
-      if (!map["KEYID"].empty())
-      {
-        std::string keyid = map["KEYID"].substr(2);
-        const char* defaultKID = keyid.c_str();
-        current_defaultKID_.resize(16);
-        for (unsigned int i(0); i < 16; ++i)
-        {
-          current_defaultKID_[i] = HexNibble(*defaultKID) << 4;
-          ++defaultKID;
-          current_defaultKID_[i] |= HexNibble(*defaultKID);
-          ++defaultKID;
-        }
-      }
-
-      current_pssh_ = map["URI"].substr(23);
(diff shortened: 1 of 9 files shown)


matthuisman:
`#EXT-X-KEY:URI="data:text/plain;base64,AAAAOHBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABgSEAAAAAAD5ndGYzEgICAgICBI88aJmwY=",KEYFORMAT="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed",KEYFORMATVERSIONS="1",METHOD=SAMPLE-AES` should be parsed as Widevine

matthuisman:
actually, this needs work

The parser fails as the first line has METHOD=SAMPLE-AES in it but no urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed
The parser needs to check all lines before failing.

Even if the above was correct and had METHOD=SAMPLE-AES-CTR - it would still fail as the first line has SAMPLE-AES

peak3d:
@matthuisman sorry, there has now much changed in hlskid, so you'll have to provide a new PR

src/parser/HLSTree.cpp
```diff
@@ -137,12 +137,12 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
 
     current_iv_ = m_decrypter->convertIV(map["IV"]);
 
-    Log(LOGLEVEL_INFO, "Supported encryption method found: AES-128");
+    Log(LOGLEVEL_DEBUG, "Supported encryption method found: AES-128");
     return ENCRYPTIONTYPE_AES128;
   }
 
   // NO VALID
-  Log(LOGLEVEL_WARNING, "Unsupported encryption method: %s with keyformat %s", map["METHOD"].c_str(), map["KEYFORMAT"].c_str());
```
peak3d: should remain WARNING

src/parser/HLSTree.h
```diff
@@ -32,11 +32,10 @@ namespace adaptive
   public:
     enum
     {
-      ENCRYPTIONTYPE_INVALID = 0,
```
peak3d: cosmetic? It's INVALID for use in inputstream.adaptive

matthuisman:
Should I base new PR on master?

matthuisman:
Should I base new PR on master?

peak3d (changes requested):
It's too hard for me too read, this PR should be splitted into 2, one for changes, and one for cosmetic.
Beside this 15 commits are not what we want to see in git history, pls. suqsh all your commits into one as long there are no reasons why to use separate commits

src/parser/HLSTree.h
matthuisman: i went for ENCRYPTIONTYPE_UNSUPPORTED (to cover UNKNOWN and invalid)
as with the new behavour, we are not actually checking for unspported.
Only checking for supported.
So any thing else would either be unknown or invalid - therefore UNSUPPORTED
It also matches the log message better

matthuisman:
ill close this pull request and do a nicer one based on lastest commit in hlksid branch

PR #395, opened 2020-03-14
HLS Widevine

glennguy:
Hi @peak3d 

This will address #353 

A test strm: 
```
#KODIPROP:inputstreamaddon=inputstream.adaptive
#KODIPROP:inputstream.adaptive.manifest_type=hls
#KODIPROP:inputstream.adaptive.license_type=com.widevine.alpha
#KODIPROP:inputstream.adaptive.license_key=https://cwip-shaka-proxy.appspot.com/no_auth||R{SSM}|
https://storage.googleapis.com/shaka-demo-assets/angel-one-widevine-hls/hls.m3u8
```
As part of these changes EXT-X-MAP and EXT-X-KEY values now persist into the next period/discontinuity as this is a requirement in HLS spec. The test stream above requires this as the first two segments are in clear, then the encryption begins. There is a discontinuity tag however the initialization segment remains the same - there is no new map tag.

The EXT-X-SESSION-KEY in the master manifest will allow the SSD to be set up in `Session::initialize` (at this time we have not yet read a variant stream and any EXT-X-KEY values in it)
As the EXT-X-SESSION-KEY is only optional in HLS spec, master playlists without it will not play. I have working code to allow for this but thought it should be left for a separate PR. I haven't found any examples of HLS/Widevine that don't contain the session key tag so far.

Apart from the test stream above, the only other example of HLS/Widevine I could find is the Disney+ streams (there is a link to master/variant stream example for this in #353 ).

One issue that I have observed: *edit - disregard, @michaelarnauts pointed out this was an exisiting issue not to do with this PR*
~~In the above example strm when transitioning from period 1 to period 2 the video halts while audio continues to play for approx 8 seconds and then the audio halts while the video plays for approx 8 seconds, then playback continues normally. The OSD timeline in Kodi also shows some weird behaviour with the now marker and buffer jumping forwards briefly then backwards. This also happens in the Disney stream where all periods are encrypted, however the correction is quicker (maybe related to segment length). I'm unable to test if this is the case in DASH/Widevine, the only multiperiod mpd I've found has very poor bandwidth/buffering issues.~~

Thanks for your time.

@matthuisman would you like to test this when you get a chance?

PR #424, opened 2020-04-17
[HLS] set pssh media type for widevine

glennguy:
Fixes crashing when playing hls streams with encrypted video AND audio with the same PSSH

Was not included in #395 as the workflow with EXT-X-SESSION-KEY was causing an extra psshset to be added -> crash when switching streams

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,90 @@
+---
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: DontAlign
+AlignOperands:   true
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: InlineOnly
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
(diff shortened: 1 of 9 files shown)


peak3d:
Sorry, this PR must be done again. Pls. make sure that you squash your commits into one, if you have cosmetic and changes, pls. provide 2 PR's, each with one commit as long there is no reason to split.

glennguy:
@Peak3d I think your comment was not intended for this PR? This is just one commit, one change.
Edit: did not notice force pushing sorry

matthuisman:
@glennguy 
Peak3d re-wrote the commit history for this branch (force-push)
So your branch is now conflicting.
You can see this under commits or files changed for this PR.
It has 3 he originally made that now exist in your PR as they no longer exist in this branch.

I had to checkout his branch and then make the changes again and push new PR.

Actually, that code your commit was for has now changed (so maybe already fixed)
https://github.com/peak3d/inputstream.adaptive/blob/hlskid/src/parser/HLSTree.cpp#L604
&
https://github.com/peak3d/inputstream.adaptive/blob/hlskid/src/parser/HLSTree.cpp#L627

But maybe this has something to do with my new issue of audio stopping after fw or seeking

glennguy:
Ah ok I see, will do

matthuisman:
@glennguy 
Just tried and that issue with crashing is still present with the new changes
And also another bug now present (see below)
https://github.com/peak3d/inputstream.adaptive/pull/417#issuecomment-615510224

Used the build from here:
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/PR-426/1/artifacts

Thats peaks latest branch with the fix for not erroring on SAMPLE-AES
Playing Apple TV with media = all - crashes

glennguy:
Perhaps since there's a bit of force pushing going on we should wait for hlskid to be merged into matrix then submit based off matrix?

@peak3d would this be your preference?

peak3d:
Yes. pls. wait until hlskid is finished

PR #426, opened 2020-04-17
Fix for HLS playlists with valid encryption after non-valid encryption and cosmetics

matthuisman:
@peak3d 
This should be a bit more tidier :)

I have found Apple TV+ playlists have a valid encryption after a non-valid encryption.
Currently IA will fail as soon as it hits a non-valid encryption.
This PR changes the behavior to keep checking for a valid encryption.

If someone is using IA, they should already know what encryptions are supported.
If playback fails, the log will show unsupported type.
Just "fails" a little less quickly than before.
Small trade-off to fix the issue of failing on playlists with valid encryptions.

I find it easier to view the processEncryption with the changes as a whole:
https://github.com/matthuisman/inputstream.adaptive/blob/f2f5d210f6e506054fe22e946d14e6e5d0e8f398/src/parser/HLSTree.cpp#L92

A lot less nesting and easier to follow.
Also easy to add a new encryption method just under the others

### DIFF
--- a/src/SSD_dll.h
+++ b/src/SSD_dll.h
@@ -35,7 +35,7 @@ namespace SSD
     virtual bool CURLOpen(void* file) = 0;
     virtual size_t ReadFile(void* file, void* lpBuf, size_t uiBufSize) = 0;
     virtual void CloseFile(void* file) = 0;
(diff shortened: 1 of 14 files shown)


matthuisman:
Closing until hlksid merged master

PR #430, opened 2020-04-20
fix: Role processing for non-subtitles

### LINKED ISSUE #422: AdaptationSet default flag regression
Previously used 2.3.x version and when I got:
```
<AdaptationSet audioTrackId="en-us_dialog_0" contentType="audio" group="1" default="true" lang="en" maxBandwidth="192000" mimeType="audio/mp4" minBandwidth="96000" segmentAlignment="true" startWithSAP="1" subsegmentAlignment="true" subsegmentStartsWithSAP="1">
```

then it was correctly marked as `default` in Kodi 18.
And when I skipped `default="true"` - it was no longer marked as `default` audio track.

With 2.4.4 regardless if I provide default="true" or not, all tracks are marked as `default` ones.

23doors:
fixes #422

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -600,11 +600,13 @@ start(void *data, const char *el, const char **attr)
           {
             if (strcmp(value, "subtitle") == 0)
               dash->current_adaptationset_->type_ = DASHTree::SUBTITLE;
-            //Legacy compatibility
-            if (strcmp(value, "forced") == 0)
-              dash->current_adaptationset_->forced_ = true;
-            if (strcmp(value, "main") == 0)
-              dash->current_adaptationset_->default_ = true;
+            else if (dash->current_adaptationset_->type_ == DASHTree::SUBTITLE) {
+              //Legacy compatibility
+              if (strcmp(value, "forced") == 0)
+                dash->current_adaptationset_->forced_ = true;
+              if (strcmp(value, "main") == 0)
+                dash->current_adaptationset_->default_ = true;
+            }
           }
         }
         else if (strcmp(el, "Representation") == 0)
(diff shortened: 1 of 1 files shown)


src/parser/DASHTree.cpp
```diff
@@ -600,11 +600,13 @@ start(void *data, const char *el, const char **attr)
           {
             if (strcmp(value, "subtitle") == 0)
               dash->current_adaptationset_->type_ = DASHTree::SUBTITLE;
-            //Legacy compatibility
-            if (strcmp(value, "forced") == 0)
-              dash->current_adaptationset_->forced_ = true;
-            if (strcmp(value, "main") == 0)
-              dash->current_adaptationset_->default_ = true;
+            else if (dash->current_adaptationset_->type_ == DASHTree::SUBTITLE) {
```
peak3d: We use to have brackets at new line (see clang-format file)
Beside this your change does only work, if `if (strcmp(value, "subtitle") == 0)` is at first place.

You'll have to store forced and main in local variables and set them later.

peak3d (changes requested):

src/parser/DASHTree.cpp
23doors: hopefully fixed it

23doors:
macos build is failing but that doesn't seem to be related

peak3d:
Sorry @23doors, but after reading the syntax of Role I see only one solution:
- Read the values as it has been before
- in OnExit of Representation tag set forced back to false if it is NOT subtitle.
- default should be kept, as we use it not onle for subs.

23doors:
@peak3d been a while since I used c++ so sorry for any potential mistakes. but hopefully right now it is as you requested?

glennguy:
Hi @23doors I'd like to close this off because:
* Nearly a year since last update
* PR against Kodi Leia - needs to be Matrix (and soon to be Kodi N)
* Is this even an issue with Amazon VOD still after this time?

23doors:
Sure, let's close it.

PR #434, opened 2020-05-04
Fix for HLS playlists with valid encryption after non-valid encryption and cosmetics

matthuisman:
@peak3d 
Here is my old PR based off latest Master

I have found Apple TV+ playlists have a valid encryption after a non-valid encryption.
Currently IA will fail as soon as it hits a non-valid encryption.
This PR changes the behavior to keep checking for a valid encryption.

If someone is using IA, they should already know what encryptions are supported.
If playback fails, the log will show unsupported type.
Just "fails" a little less quickly than before.
Small trade-off to fix the issue of failing on playlists with valid encryptions.

I find it easier to view the processEncryption with the changes as a whole:
https://github.com/matthuisman/inputstream.adaptive/blob/f2f5d210f6e506054fe22e946d14e6e5d0e8f398/src/parser/HLSTree.cpp#L92

A lot less nesting and easier to follow.
Also easy to add a new encryption method just under the others

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -91,65 +91,64 @@ HLSTree::~HLSTree()
 
 int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::string>& map)
 {
-  if (map["METHOD"] != "NONE")
-  {
-    if (map["METHOD"] != "AES-128" && map["METHOD"] != "SAMPLE-AES-CTR")
-    {
-      Log(LOGLEVEL_ERROR, "Unsupported encryption method: %s", map["METHOD"].c_str());
-      return ENCRYPTIONTYPE_INVALID;
-    }
-    if (map["URI"].empty())
-    {
-      Log(LOGLEVEL_ERROR, "Unsupported encryption method: %s", map["METHOD"].c_str());
-      return ENCRYPTIONTYPE_INVALID;
-    }
-    if (!map["KEYFORMAT"].empty())
-    {
-      if (map["KEYFORMAT"] == "urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed")
-      {
-        if (!map["KEYID"].empty())
-        {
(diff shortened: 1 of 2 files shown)


matthuisman:
also using just KEYFORMAT to check for Widveine.

As Apple TV+ uses SAMPLE-AES not SAMPLE-AES-CTR
This may be due to it being CBC (not CTR)
See issue here: https://github.com/peak3d/inputstream.adaptive/issues/425

CastagnaIT:
@matthuisman can you play widevine content on AppleTv tvOS?
where do you get the Widevine libraries files?

matthuisman:
Don't think you can...?

Appletv+ in browser uses widevine so an addon for it should be possible

CastagnaIT:
no currently i do not know a method to get the libraries for tvOS or iOS ..i misunderstood your post

if the apple browser supports widevine, i don't know if it's possible to make an extraction
i don't have any apple devices, so i am not able to try to do tests
also talking with guys of ISHelper some time ago they found no solution

matthuisman:
I also dont have any apple devices.
This PR helps towards a working add-on for Apple TV Plus streaming service - so less people need apple device for using it :)

It will also fix any other HLS playlists that may have a valid encryption come after a non-valid encryption in the file

peak3d:
@matthuisman there is a `dash->supportedKeySystem_` which can be used for checking for supported DRM methods. Sure AES-XXX should be done in place, as they are "build-in" and will never be in the supportedKeySystem. Other idea would be adding AES-XXX into the supportedKeySystem array to avoid special handling in HLS / AES.

Concept is: If content decryption is found in manifest, set `dash->current_period_->encryptionState_ |= DASHTree::ENCRYTIONSTATE_ENCRYPTED`. If the DRM is supported (part of `dash->supportedKeySystem_`), then set `dash->current_period_->encryptionState_ |= DASHTree::ENCRYTIONSTATE_SUPPORTED`

PR #440, opened 2020-05-09
Support Dolby Vision in HLS

matthuisman:
Referencing your commit for DV support: https://github.com/peak3d/inputstream.adaptive/commit/f61409ee6a3205cc1dc21b34e2dfd5dee146692f AP4 format for Dolby Vision is **dvhe** & **dvh1**

https://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-streams-within-the-http-live-streaming-format-v1.1.pdf

Example HLS codecs

**CODECS="dvhe.05.07,ec-3"
CODECS="dvh1.05.06,ec-3,mp4a.40.2"**

Side note: 

According to the dolby pdf, you can also have
CODECS="hvc1.1.0.L120.00,dvh1.04.09,ec-3"

"In this example, the highlighted string indicates that the video element is a dual-layer Dolby
Vision stream encoded as 10-bit HEVC video in each layer with a max resolution of 3840 ×
2160 at 60 fps, where the base layer is a backward-compatible, SDR Rec.709–compliant
signal."

Current HLS code will return hvc1 instead of dvh1
So not sure if we would want to re-arrange so dvh1 is check and returned first? 

**Always return the best codec? Or always return the most likely to be supported?**

I guess you would need to know what the device supports...
Same issue as having multiple audio codecs listed
But that's a problem for a different day.

For this PR, I just followed the same existing order ("worst" to "best")

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -70,6 +70,11 @@ static std::string getVideoCodec(const std::string& codecs)
     return "hvc1";
   else if (codecs.find("hev1.") != std::string::npos)
     return "hev1";
+  else if (codecs.find("dvh1.") != std::string::npos)
+    // Kodi only supports dvhe so return that
+    return "dvhe";
+  else if (codecs.find("dvhe.") != std::string::npos)
+    return "dvhe";
   else
     return "";
 }


M2k80:
just tested the jenkins build artefact on my shield, it seems thhis does not fix disney+ and dolby vision audio only

matthuisman:
Thanks for letting me know. I am unable to test. I'll look into it more soon and ping you so you can test for me :)

M2k80:
@matthuisman but it seems there is something confusing on the jenkins build. #440 says  support for DV in hls, the jenkins build is labeled with DV in hls with pr-380. so maybe there is something wrong?
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/activity

so if i would take pr-440 which is not marked as dolby vision support in hls, its only 2.5.6v so this one is for matrix only.

edit: tried 2.5.6 pr-440 on matrix, it did not work also

matthuisman:
95% sure found the issue and have updated PR :)

matthuisman:
@M2k80 
try this (kodi 18 with this pr + others)
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/PR-438/4/artifacts/

M2k80:
@matthuisman its a step forward. DV stream is now starting, but not correctly detected by the tv. picture gets strange colours but movie is running for the first time when DV is enabled. also the brightnes is set up like on hdr

matthuisman:
does the netflix add-on dolby vision work on your tv?

M2k80:
yes this is working perfect

matthuisman:
even with the version of IA I linked above?

Disney uses
dvh1.05.01 Dolby Vision Profile 5 (10-bit HEVC), Level 1 (720p24) vide
dvh1.05.06  Dolby Vision Profile 5 (10-bit HEVC), Level 6 (2160p24) video

Appears netflix is the same.
So i'll try forcing IA to dvhe (what netflix addon uses) instead of dvh1
However, everywhere online suggests using dvh1 over dvhe so not sure!

@CastagnaIT
Any ideas about this dolby vision stuff?

matthuisman:
ok found the issue.
IA added DV support here:
https://github.com/peak3d/inputstream.adaptive/commit/f61409ee6a3205cc1dc21b34e2dfd5dee146692f

Note: dvh1 and dvhe

However, Kodi code only had support for dvhe added:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp#L485

Therefore, dvh1 will have the wrong mimetype and format name set.

So, I'll PR a fix for that

Actually, I think dvhe is the same as dvh1 - so I'll just set the codec to dvhe in IA.
Then no update to Kodi needed.
I also wasn't sure if Android supports 'amc-dvh1' or only 'amc-dvhe'

M2k80:
ok, i see i am a little late, just tested again netflix, and its working fine. i think you dont need a screenshot? it look like it uses the wrong color profile? but i am not the expert xD

matthuisman:
new build in progress here:
should be ready in a few minutes - give it a try please 
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/PR-438/5/artifacts

M2k80:
just tested, its working. thank you. good job

CastagnaIT:
I have no experience with Dolby Vision with netflix because i do not have an android device compatible and do not have a tv 4k
so is implemented as is...

I don't know how to figure out what kind of codec profile netflix uses
the only string name that i have is hevc-dv5-main10-L**-dash-cenc-prk
the number 5 perhaps can suggest profile 5?..

I know that users who use Firestick 4k have problems with DV, so often lose the signal
unfortunately i cannot provide a detailed context or tests
to say it's a firestick firmware problem or it's badly implemented on ISA/Kodi

facakente:
Yes I have that problem , using Firestick 4K in some shows as lost in space it detects DV but couple of secs later it loses DV and plays in SDR.
Then playing Irishman it detects DV, it keeps playing DV but resolution changes from 4K to 1080p. 

Though in Netflix colours look correct.


On the other hand in Disney with latest build I can now play movies with DV profile enabled , tv detects DV but couple of secs later it loses DV signal , although Firestick still playing in DV mode , so there is a problem with signal between Firestick and tv , which I think makes the tv to play with wrong colours .

matthuisman:
Seems the shield is OK though.
Does HDR10 work ok for you?

I wonder if we try pass dvh1 instead of dvhe. Everywhere suggests dvh1 is preferred. Kodi code would need update

CastagnaIT:
you could add support, so there is the possibility of running tests

matthuisman:
Just not sure what the android format names are. This is what dvhe is using:

m_mime = "video/dolby-vision";
m_formatname = "amc-dvhe"

Is 'amc-dvh1' even a valid android format?
I can't find anywhere to check online etc

facakente:
Yes hdr10 is working fine in Disney , movie starts playing at 60p, changes later on to 24p , colours are correct.

matthuisman:
And official Disney and Netflix apps work with Dolby Vision fine in your setup?

facakente:
Yes sure , official apps work fine. 

But at 60p , my goal is to play those at 24p.

CastagnaIT:
m_formatname i think is used only for debug and the codec info window
m_mime is right as is https://source.android.com/devices/tech/display/hdr
or better https://developer.android.com/reference/android/media/MediaFormat#MIMETYPE_VIDEO_DOLBY_VISION

facakente:
Ok regarding the DV being played at 1080p in netflix  , same is happening with official app , so might be a problem on Netflix’s end.

M2k80:
@matthuisman any idea why the stream/player get paused after mode switch (60hz to 24hz)?
i cant see that on my netflix, but on netflix it switch the mode directly after the stream starts, but on disney+ it take 3 sec, then the switch happens and the stream is paused and i have to resume it manually.
could this commit be the reason why this happens? https://github.com/xbmc/xbmc/pull/17541

matthuisman:
@M2k80 
I've had confirmation that change is only in 18.7 (currently Leia Nightly) and also in Matrix nightly

If your using 18.6 - could you try Leia Nightly and see if you have the same result.
Also try Matrix Nightly if possible.
If your not using 18.6 - could you please try that (just from playstore)

If it doesn't do it in 18.6 - but does in both nightlies - then it could be that commit possibly.
as quick as possible please as 18.7 isn't far - so if that is the issue - we need to get it sorted :)

Does it pause on other content when switching framreate?
Or only DV? What about HDR10?

Can you also confirm on your Shield with DV content - that the colours stay correct after a frame-rate switch? 

It seems on Firestick DV content drops back to SDR after a framerate switch which causes funny colours etc. However, having the Firestick already set to 24fps - it works fine. It's the switch that causes issues. @facakente - is that correct - does yours pause on framerate switch?

M2k80:
i used the disney+ addon since dv is working, never saw any color mismatch since then.
i have actually installed latest leia nightly and matrix nightly.
i try to get a leia nightly before this audio noise commit and check it again

M2k80:
@matthuisman ok, just installed 18.6 final, there is no pause and stream is running
sadly i dont have an image to test directly before this commit or without this commit

matthuisman:
@M2k80 Did you have the pause for other content? Eg. HDR10? Or only Dolby digital?

facakente:
Yes indeed, DV works fine with Disney plus addon , I noticed what is causing the issue is when the frame rate changes from 60 to 24p.
My setup is Kodi is to set frame rate on start/stop 

As workaround I did set Kodi(no firestick) to 23.98p therefore it does not change frame rate when I play a video , though UI is not smooth.

facakente:
This same issue happens with Netflix with for example “ lost in space” when it changes frame rate tv loses DV signal and plays with wrong colours.

But on other hand it does not happen with Irishman.

facakente:
My video does not pause currently , but that’s true that it used to happen , maybe with hdr10 , I can test and let you know.

matthuisman:
Looks like pausing issue is not in 18.6. You'd need to use nightly Leia build to check

M2k80:
i am testing it now for hdr10,
but pause will for me only happen, if the modeswitch happens after 5-6sec instead of directly when starting the stream, my default delay is 0.3 to modeswitch, so netfilix is normaly not a problem.
but for ex. hevc stream needs 3-4 sec to detect the framerate so the switch is later then normal, what means this is maybe only happening on hevc streams....gonna test hdr10 now on 18.6 and 18.7 nightly, reporting back soon.

but you cant also go to settings and select resolution manually and change the Hz for testing, it will pause too,

facakente:
Ok I am on 18.6 and there is no pausing issue  with dV or hdr10, works fine.

But I do remember to see it , maybe it was on the matrix version which I have removed .

M2k80:
like i thought, it pause also on hdr10 hevc stream (x men) with latest nightly.

if someone could make latest nightly build without this xbmc/xbmc#17541 commit, i can test, if this is causing it

facakente:
“but for ex. hevc stream needs 3-4 sec to detect the framerate so the switch is later then normal”

Yes I confirm , that was the same issue I experienced .

matthuisman:
The commit we suspect is not in 18.6.
 But is  in Leia 18.7 nightly. Seems you need to have bigger refresh delay to cause it.

M2k80:
@matthuisman late mode switch happens also on 18.6 but without pausing the stream after, pause only happens on the latest nightlies. i can 100% say pausing is not happening 18.6 final.

matthuisman:
@M2k80 
Any other formats it does it on? I suspect maybe all? Just test by manually changing if you don't have content in that frame rate

M2k80:
i can test local some local movies

matthuisman:
Also. Try not having frame rate switch but try switching audio format after playback. Eg. On Disney change from 5.1 to 2.0. See if that change makes it pause

M2k80:
@matthuisman latest nightly, movie starts, framerate switches no pause, then manually change the frame rate, movie is paused.

ok tested changing the audio stream to different languages and 5.1 to 2.0 there is no pause. 
maybe those changes are so fast that no "noise" is detected??i had no delay when switching audio. idk

best would be, if could thest an image without this commit, then we were 100% sure if this is causing it

matthuisman:
I'll get you a 18.7 build without commit :)

Draft PR to revert commit here:
https://github.com/xbmc/xbmc/pull/17868

Jenkins should build that shortly and then have a APK to try.
This will be 18.7 without commit

Just to be 100%, when you say latest nightly - you mean 18.7 nightly (not 19 nightly)?

@facakente
Try Leia 18.7 nightly (not matrix 19 nightly) and see if you get pause.
Pretty sure you will
Also, easier to test 18.7 as same IA version keeps working, db is the same etc etc
It shouldn't "mess up" your install.
Plus 18.7 stable isn't far away anyway so it should be pretty stable - except for this pause issue

M2k80:
u didnt used the backport commit, does it matter?

i was always talking about 18.7 "latest nightly" :)

matthuisman:
Yes i did:)

Its just the commit msg doesn't have backport in it (only the PR title did)
https://github.com/xbmc/xbmc/pull/17541/commits/46d79a9543461bd4b71958d6381c886857f1fc14 (note: Leia branch)

https://github.com/xbmc/xbmc/pull/17868
"This reverts commit 46d79a9."

M2k80:
ok no problem, i am not a dev neither a github pro xD

matthuisman:
no worries.

I'm mainly Python but dabble in C++ if it's simple enough. haha
And also pretty 'git' novice. Lots of trial and error.

And thank you for all your testing!
Better to find the issue now before it gets shipped out to everyone in 18.7!

Building now:
https://jenkins.kodi.tv/job/Android-ARM64/14703/
Once done - should be a artifacts at the top which links to apk

facakente:
I suppose I need to download this one right ?
https://jenkins.kodi.tv/view/Android/job/Android-ARM/18067/artifact/

As I am using the firestick 4K

matthuisman:
This is Kodi 18.7 without the "noisy" commit.

@M2k80 
Please try this APK (ARM64 for Nvidia Shield):
https://jenkins.kodi.tv/job/Android-ARM64/14703/artifact/kodi-20200514-ee2959ce-PR17868-merge-arm64-v8a.apk

We want to compare this with latest 18.7 ARM64 nightly
http://mirrors.kodi.tv/nightlies/android/arm64-v8a/Leia/kodi-20200513-33760e28-Leia-arm64-v8a.apk

Please confirm the new build works correctly and the 18,7 nightly does not

@facakente 
Please try this APK (ARM for Firestick 4K (32bit OS)):
https://jenkins.kodi.tv/job/Android-ARM/18067/artifact/kodi-20200514-ee2959ce-PR17868-merge-armeabi-v7a.apk

We want to compare this with latest 18.7 ARM nightly
http://mirrors.kodi.tv/nightlies/android/arm/Leia/kodi-20200513-33760e28-Leia-armeabi-v7a.apk

Please confirm the new build works correctly and the 18.7 nightly does not

M2k80:
@matthuisman your build fixes stream getting paused after modeswitch!

facakente:
With this new build I am having stability problems , stream starts but sec later kicks me out of kodi.

matthuisman:
@facakente

We are only looking for the pause behaviour differences.
The new build only reverts the audio pause commit - nothing else that should cause instability

that could be due to other commits in 18.7
Did you compare with the 18.7 nightly here:
http://mirrors.kodi.tv/nightlies/android/arm/Leia/kodi-20200513-33760e28-Leia-armeabi-v7a.apk

facakente:
Same happens with leia nightly , every time I try to play something I just get kicked from kodi.

No idea why, if I install again stable leia , it works fine .

matthuisman:
ok, so not related to what we are doing here

18.7 is close though - so it may be best you start a bug report / kodi forum post about it
Could be Firestick only so no one has noticed yet

facakente:
happening specifically with mandalorian , playing Star Wars it works fine . Need to check further.

glennguy:
Hey @matthuisman 

Regarding the dvh1 vs dvhe - have a look here: https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/main.cpp#L2696

If we're indeed playing a dvh1 stream then let's allow for both in HLSTree.cpp and also add in dvh1 in the section linked above. Perhaps having the wrong fourcc set is causing ffmpeg issues?

matthuisman:
I had done that originally, how ever kodi source only allows for one

matthuisman:
Yip, I had done that originally.

I had this line
  else if (rep->codecs_.find("hvc") == 0)

Replaced with 
else if (rep->codecs_.find("hvc") == 0 || rep->codecs_.find("dvh") ==0)

And removed the dvhe block.
So less code duplicated

But kodi only supports dvhe.
You pass dvh1 to kodi and it doesn't recognise it. If you read comments above I was discussing adding support in kodi but wasn't sure if there is even a different android codec type for it.

https://github.com/peak3d/inputstream.adaptive/pull/440#issuecomment-626239027

Kodi just passes it through to android basically. 

That's why I added the comment above the line where I return dvhe instead of dvh1



Its working 100% on sheild except for the pause issue that was related to another PR.

Firetv has some other issues caused by framerate switch. But that also happens with Netflix addon which isn't using hls

CastagnaIT:
some time ago they were trying to update the kodi HDR management code,
that in the PR (now abandoned for other reasons)
was fixing latencies, switches and color issues

But you can find on Kodi HDR mod: https://forum.kodi.tv/showthread.php?tid=345566

it would be interesting to know if your problems are solved with this HDR mod version

CastagnaIT:
[DELETED]
ah hdr mod was only for windows...sorry

peak3d:
If the framerate is not in manifest, it needs computation in videoplayer and switch is done after detection. So most probably framerate is not in hls (m3u8) files

glennguy:
Ah ok, all good @matthuisman. Yeah looks like that line you linked to in the Kodi source should be updated for future perhaps? Not sure myself if it's needed since it looks like Kodi is taking a dvh1 file reported as dvhe anyway! My thinking of allowing for reporting both as they are was only extending from the fact that we currently differentiate between hevc/hvc1.

@peak3d Wrt to frame rate I've made a small PR - #448 - as we weren't getting the frame rate in HLS.

peak3d (approved):

matthuisman:
@facakente 
Could you try this IA build on your FireStick 4K?
Just using Leia stable (18.6)

https://jenkins.kodi.tv/job/peak3d/job/inputstream.adaptive/job/PR-438/8/artifact/cmake/addons/build/zips/inputstream.adaptive+android-armv7/inputstream.adaptive-2.4.4.zip

This contains @glennguy PR to parse framerate from HLS.
Want to check if it fixes your issue where DV is dropped on late framerate switch with Disney+.

@M2k80 
You may also want to try this build below
(may mean you don't need such delayed framerate switch)

https://jenkins.kodi.tv/job/peak3d/job/inputstream.adaptive/job/PR-438/8/artifact/cmake/addons/build/zips/inputstream.adaptive+android-aarch64/inputstream.adaptive-2.4.4.zip

It should fix the issue where framerate doesn't switch until a few seconds in.
I confirmed it fixed it for me.

However, I did notice without this fix on just HDR (I dont have DV) content on the late frame-rate switch, I got a HDCP error on the TV just after the change - it didn't like it.

So, the above fixes it for Disney+ - but there is still an issue with Kodi Android not liking changing framerate during playback.

facakente:
Ok tested and looks good, there was progress, so usually I was playing mandalorian to test it and I don’t know why I still experiencing problems with this show , it switches to 24 p then switches to 25p and at end it loses dv signal and messes with colour , however playing other shows as wreck it Ralph or Star Wars , issue was fixed. It switches directly to 24p and plays fine.

Also in Netflix fixed the frame rate issue I had with Irishman where it was switching to 1080p, but on the other hand issue with lost in space still present , similar to mandalorian where it detects DV and loses signal and messes with colours.

facakente:
https://paste.kodi.tv/zomasacudo 

Logs when playing mandalorian and behaviour described above

facakente:
Most of shows movies play fine in Disney or Netflix , just some of them like lost in space in Netflix and mandalorian in Disney issue happens , what I noticed in most of shows frame rate switches to 24p when stream starts , whereas in those where I have problems it starts with 60p and couple of sec later switches to 24p , losing DV and messing colours

CastagnaIT:
to netflix the value of the "frameRate" tag was wrong
i have fixed, now have the value as:
"rate/scale"

if you want to try again lost in space if perhaps now working
kodi 18: [plugin.video.netflix_1.2.2_20200516.zip](https://github.com/peak3d/inputstream.adaptive/files/4638027/plugin.video.netflix_1.2.2_20200516.zip)
kodi 19: [plugin.video.netflix_1.2.2+matrix.1_20200516.zip](https://github.com/peak3d/inputstream.adaptive/files/4638029/plugin.video.netflix_1.2.2%2Bmatrix.1_20200516.zip)

matthuisman:
Netflix returns mpd?

So the hls frame-rate update would have made no difference?

I'm guessing the pr just made you look into it?

Im guess the hls for mandalorian returns wrong frame-rate. That why it changes twice. However, kodi should be able handle that and not distort. So something else going on still

CastagnaIT:
i have see: https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/parser/DASHTree.cpp#L642
parse a string as "int/int" as some examples on the net

but to now netflix returned with pre-calculated float value, was wrong

i have found also this:
https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/main.cpp#L595
seem change the value stored, but i can't figure out how it works, it may not be related...this confuses me🤔

facakente:
Yes , lost in space fixed.
So currently Netflix all good.

In Disney only mandalorian has problems , also with hdr10 tv changes several times frame rate , but at the end plays in hdr10 , no problems with colours like in dv

CastagnaIT:
awesome at least one thing is solved!
in what kodi version?

facakente:
18.6 leia stable

matthuisman:
OK, so issue is changing frame-rate during Dolby Vision content only? Do you know what format it's dropping too? I'm guessing it's starting as Dolby vision but then after the frame-rate switching to HDR?

facakente:
Another show in Disney with same problem , lady and the tramp , as workaround I have de whitelisted 25p and 50p, this way it matches 24p directly and plays fine , otherwise it matches 25p later 50p and finally 24p which causes the problem

facakente:
Not with every show , most of them play fine.

So far only 2 with problem lady and tramp and mandalorian, problem is because it tries to match first 25 p , whereas most of shows matches directly 24p and it’s fine 


Problem with colours is due the tv is playing in 4K (loses Dv) but firestick still detecting DV codec.

CastagnaIT:
but in the manifest of lady and the tramp the framerate is right?
it seems that even in this case somewhere it's not being properly reported

matthuisman:
Lots of playlists may not include the frame-rate, so let's assume we don't have the information. 

Also a local dolby vision mp4 or mkv would not be able to advise fps until playback starts. 

The current getting fps right in file is just a plaster over the real issue. We shouldn't need to rely on the frame-rate in file being correct. Kodi will get correct frame rate once playback starts anyway. 

Might be time for a debug log with video component enabled.

If you open kodi info while playing, what codec does it say before and after the frame-rate switch? Does kodi still think its outputting DV?

facakente:
How do u see the codec info in kodi ? Because I have installed the Kodi remote and doesn’t mention anything about codec DV info, only bitrate info, also tried with key map editor and under codec info does not mention anything if playing DV.

However checking in advanced mode in the firestick codec info shows DV while playing even with wrong colours.

matthuisman:
@facakente  
Codec info is **o** on a keyboard.
I cant seem to see a way to get to it via the menus though..

There is now a Kodi issue created for DV after framerate switching over here
https://github.com/xbmc/xbmc/issues/17451
will continue discussion over there.
Please keep it on topic about that issue only.

I just posted a DV mp4 stream that would be interested to see if you have the issue with that.

PR #441, opened 2020-05-10
Don't retry subtitle segment downloads on http 415 error

### LINKED ISSUE #433: MPD manifest error causes playback delay
We discovered that an incorrect MPEG-DASH manifest, including a subtitle that gives HTTP 415 (Unsupported Media Type) cause IA to retry downloading 10 times causing playback delays from 10 to 20 seconds.

Maybe if a subtitle track causes HTTP 415, it should not be retried, but skipped.

See add-ons/plugin.video.vrt.nu#735 for the details.

mediaminister:
This is a first attempt to fix https://github.com/peak3d/inputstream.adaptive/issues/433

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -61,12 +61,24 @@ void AdaptiveStream::ResetSegment()
     absolute_position_ = current_rep_->current_segment_->range_begin_;
 }
 
-bool AdaptiveStream::download_segment()
+unsigned int AdaptiveStream::download_segment()
 {
   if (download_url_.empty())
-    return false;
+    return 400;
 
   return download(download_url_.c_str(), download_headers_);
+}
+
+bool AdaptiveStream::retry(unsigned int returnCode)
+{
+  if (returnCode == 200)
+    return false;
+
+  //Don't retry subtitle segment downloads on HTTP 415 Unsupported Media Type
+  if (type_ == AdaptiveTree::SUBTITLE && returnCode == 415)
+    return false;
+
+  return true;
 }
 
 void AdaptiveStream::worker()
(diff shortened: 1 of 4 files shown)


src/common/AdaptiveStream.cpp
```diff
@@ -61,36 +61,48 @@ void AdaptiveStream::ResetSegment()
...
 
   return download(download_url_.c_str(), download_headers_);
 }
 
+bool AdaptiveStream::retry(unsigned int returnCode)
+{
+  if (returnCode == 200)
+    return false;
+
+  //Don't retry subtitle segment downloads on HTTP 415 Unsupported Media Type
```
dagwieers, line 77: My only remark is to better document why we do this, and what we expect as behaviour.

So what is needed here is the following rationale/behaviour:
- A subtitle track is available from the manifest
- The files do not currently exist (HTTP 415) but may exist in the future
- So we do not need to retry the current segment
- But we do want to get the next segment

dagwieers (changes requested):

src/common/AdaptiveStream.cpp
```diff
@@ -61,36 +61,48 @@ void AdaptiveStream::ResetSegment()
...
+  if (type_ == AdaptiveTree::SUBTITLE && returnCode == 415)
```
dagwieers, line 78: You can now add 404 here as well. I got 404 errors on VRT Eén during normal playback and the retries cause repeated errors in logs here as well. So no retries are useful on 404 errors either.

dagwieers (changes requested):

src/common/AdaptiveStream.cpp
matthuisman: what about other 4xx? 5xx?
We already no it's not a an ok repsonse.
Why not just do if (type_ == AdaptiveTree::SUBTITLE) then.
That means any error response on subtitle doesn't retry.

src/common/AdaptiveStream.cpp
matthuisman: There are more codes that could be an ok response (206 for example)
Maybe check 200 <= status_code < 400

Actually, I dont think we even need that.
I'd revert back to current IA behaviour and then just return from the retry if subtitle.
That way is trys to parse the data and returns false if fails.
No need to check any codes

Simpler and means the PR is one line of code (easier to get merged)

src/common/AdaptiveStream.cpp
dagwieers: @matthuisman Fine for me as well, although if we want to do this right I would only make 404, 408, 410 and 415 retry the next segment. Everything else is fatal and should stop handling the subtitle altogether (and log this event).

matthuisman: looking at the code - the retry stuff is in it's own thread.
We juts need to tell IA to not wait for retry threads for subs?
Would that be a better fix? So it still retries - but doesn't wait for it.

matthuisman: maybe something like this:
https://github.com/peak3d/inputstream.adaptive/pull/445/files
(ill let it build and give it a test)

dagwieers: No, I think there is no use in retrying for the same subtitle segment ever if they failed.
Only for 404, 408, 410 and 415 trying the next segment makes sense.
But the easiest definitely is to always try the next segment on failure.

Only THEOplayer is doing this correctly IMO, Shaka player is doing one retry, then moves to the next. https://github.com/peak3d/inputstream.adaptive/issues/433#issuecomment-626994942

PR #444, opened 2020-05-11
Fix aes128 HLS playlists with > 1 key

glennguy:
Hi @peak3d 

HLS streams with more than 1 KEY tag are failing to play. This is due to segment.pssh_set_ being stuck at a value of 1 - it was only being reset to 0 after a discontinuity and so in this case any future segments within the period they will always inherit the same pssh set (1), which ends up being the key/iv of the last KEY in the playlist.

Since the values of current_iv_ and current_pssh_ are retained following a new discontinuity we don't need to reset segment.pssh_set_ to 0. As before the large hls rework commit, the segment psshset is assigned to for every segment in the case of aes128.

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -625,6 +625,7 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
               return PREPARE_RESULT_FAILURE;
             case ENCRYPTIONTYPE_AES128:
               currentEncryptionType = ENCRYPTIONTYPE_AES128;
+              segment.pssh_set_ = 0;
               break;
             case ENCRYPTIONTYPE_WIDEVINE:
               currentEncryptionType = ENCRYPTIONTYPE_WIDEVINE;


matthuisman:
Test stream here:
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_aes128_each_segment/test.strm

peak3d:
This block of code is quite new because of changes made in the rework PR.
It's due to ref-counting. The correct solution is reseting segment.pssh_set_ to 0 if a new key arrives

glennguy:
@peak thanks for the feedback! I've adjusted accordingly

glennguy:
~~Wait... this breaks for playlists with same key used over multiple segments, typically just one key for entire playlist~~

glennguy:
~~In any case (either single key for multiple segments or 1 key per segment) the HLS spec is to maintain the current key until changed, even across discontinuities/periods, therefore resetting after arrival of new key goes against this.~~ Nevermind, I am now seeing that I'm wrong (was testing by hand, not in kodi) - as I pointed out (to myself now 😄 ) current_iv_/current_pssh_ are still remembered. 

~~The reasoning behind removing the block as I did in 8780108 is with the current code as far as I can see we can never have segment.pssh_set_ be greater than 1, it will just increase the use count for psshset 1 - that block seems to exclude the use case of multiple keys. Am I incorrect in this thinking?~~

This change works fine, but I wonder will we ever reach the new block of code with segment.pssh_set_ > 1? i.e is the else clause now redundant?

peak3d:
IMO segment.pssh_set must be set to 0 here:

https://github.com/peak3d/inputstream.adaptive/blob/d9cecfadad147ccd3b5aaf33b3820e55e6e66866/src/parser/HLSTree.cpp#L628

glennguy:
Sure, will do.

One last question - do you agree it would be appropriate to remove this line in the discontinuity block: https://github.com/peak3d/inputstream.adaptive/blob/2a6dd62a5692a3b59ed6a546dab1d7a7829fdef3/src/parser/HLSTree.cpp#L603 (edit, sorry no line number before)

A discontinuity should not clear what key is in use, playlists should provide a new key or METHOD=NONE to remove current key

peak3d:
Not sure how this behaves, if advertisment is unencrypted for example. How do we reset encryption for this case?
For Disney they provide for every "Period" an own (but same) EXT_INF_KEY, so I suggest to not change this line.

glennguy:
It looks like the setting of keys in playlists may be explicit, at least for whatever software is packaging it as far as I can see.
This is the example that I have seen of unencrypted advertisement:
```
#EXT-X-KEY:METHOD=AES-128,URI="https://sbslive.akamaized.net/2876663893/303a61ed/1/12/665001_128K1000K512K1500K.mp4.csmil/crypt.key",IV=0x00000000000000000000000000000071
#EXTINF:2,
https://sbslive.akamaized.net/2876663893/303a61ed/1/12/665001_128K1000K512K1500K.mp4.csmil/segment113_2_av-A.ts
#EXT-X-DISCONTINUITY
#EXT-X-KEY:METHOD=NONE
#EXTINF:5,
https://redirector.googlevideo.com/videoplayback?id=98cae37de26287aa&itag=341&source=dclk_video_ads&ctier=L&sgovp=clen%3D1359054%3Bdur%3D30000%3Bitag%3D340%3Blmt%3D1581465635604642%3Bgir%3Dyes%3Bxtags%3Dvc%3Davc%3Avccf%3Dbaseline%3Avsz%3D640x360%3Avtbr%3D360&gomv=google_9177740322419316128%2Cgoogle_7147704119781262921%2Cgoogle_4628679744695446301%2Cgoogle_1801228858660746812%2Cgoogle_3751801312109021404&acao=yes&mime=video/MP2T&sgoap=clen%3D150442%3Bdur%3D30185%3Bitag%3D340%3Blmt%3D1581465632238004%3Bgir%3Dyes%3Bxtags%3Dac%3Daac%3Aach%3D2%3Aasr%3D22050%3Aatbr%3D64&faskm=216&fvskm=125&gomt=100%2C7500%2C14750%2C22500%2C29000&ip=0.0.0.0&ipbits=0&expire=1581738833&sparams=ip,ipbits,expire,id,itag,source,ctier,sgovp,gomv,acao,mime,sgoap,faskm,fvskm,gomt&signature=3B6B6AB0E7E48B77E06FFA0BBC72C806D768B87D.0691E8AFB884BAD6BF4709ED01234A1058125AF3&key=ck2&fvidshow=125&fvidskip=0&govp=slices%3D0-227830&faudshow=108&faudskip=0&goap=slices%3D0-50802
```

The Disney example is explicit too even though redundant

This is what I have found from the hls specifications:
```
   Any unencrypted Media Segment in a Playlist that is preceded by an
   encrypted Media Segment MUST have an EXT-X-KEY tag applied to it with
   a METHOD attribute of NONE.  Otherwise, the client will misinterpret
   those segments as encrypted.
```
There is no mention of resetting keys after applying a discontinuity tag

However with all this said, changing that line will probably not make any difference in the real world :)

So let's merge if you're happy!

matthuisman:
Awesome work guys.

@peak3d 
Please remember to backport this as the HLS rewrite that caused the HLS AES128 regression has been backported.

PR #445, opened 2020-05-12
Reduce retries to 3 and don't retry subtitle segments

matthuisman:
Test builds here (matrix)
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/PR-445/1/artifacts/

Test strm here
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_415_subtitles/test.strm
You need to have Kodi Settings > Player > Language > Preferred Subtitle Language > Dutch

Without PR: 20+ second delay - live-text_track_0_dut tried 1+10 times before playback
With PR: less delay - live-text_track_0_dut only tried once

Pretty easy to read PR as well.
If subtitle - then set retries to 0.

Not sure if this is the best fix.
If IA didn't wait on the download thread for Subtitles - then that should also work.
@glennguy  looking into that when has time :)

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -77,7 +77,7 @@ void AdaptiveStream::worker()
     thread_data_->signal_dl_.wait(lckdl);
 
     bool ret(download_segment());
-    unsigned int retryCount(10);
+    unsigned int retryCount(3);
 
     while (!ret && !stopped_ && retryCount-- && tree_.has_timeshift_buffer_)
     {
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -79,6 +79,11 @@ void AdaptiveStream::worker()
     bool ret(download_segment());
     unsigned int retryCount(3);
 
+    //Dont retry failed subtitle segments
+    if (type_ == AdaptiveTree::SUBTITLE) {
+      retryCount = 0;
+    }
+
     while (!ret && !stopped_ && retryCount-- && tree_.has_timeshift_buffer_)
     {
       std::this_thread::sleep_for(std::chrono::seconds(1));


dagwieers:
As I stated in #441 I don't think it makes sense to handle subtitles for anything but 404, 408, 410 and 415. Everything else to me are fatal errors and shouldn't be getting any next segments either.

For lowering the number of retries on video/audio segments I don't know. I would like to know why it was set to 10 in the first place.

This is based on our study on behaviour of other players: https://github.com/peak3d/inputstream.adaptive/issues/433#issuecomment-626994942

matthuisman:
this is doing that - it simply does not retry ever on subtitles.

You'll notice IA doesn't redownload the mpd after every segment like those other players do.
Pretty sure it only re-downloads the segment when the segments in it "run out" (need refresh)

Pretty sure the HLS spec said if there is an error in segments - refresh the mpd.
That may be why the web players download the mpd every segment when it's not actually required. Just to check if the subs have fixed themselves.

Thats a lot of extra requests to fix there work-around.

dagwieers:
So ideally it should be trying to get the next segment on 404, 408, 410 and 415 :-)
Because the intended reason for having this subtitle track with unavailable segments is that at some point in the future the subtitles are turned on and the player picks it up automatically.

Maybe that ought to be a feature request.

dagwieers:
Maybe the comment should reflect that ideally IA should try the next segment rather than giving up as a TODO/FIXME note.

matthuisman:
That would be done elsewhere.
This code (thread) is just for that single segment.
Some changes needed elsewhere to try next time.

You'll notice without this PR, after the initial 11 tries - it doesn't try get future segments either

yeh, not sure how IA works with trying future segments if the previous ones fail.

timescale="1000" S t="1589243476160" d="1920" r="468"
1920/1000 = 1.92 second duration
repeated 468 times.

So, I thought IA would try get the next one (1.92 seconds later)

Thats why I thought not retrying on subs is ok - as it could get it at the next one.

We should be seeing attempts every 2 seconds for the next text segment.
But only 1 attempt each time (not 11)

still not sure why all the other players keep downloading the mpd again.
they should have all the info the need until run out of segments.
Maybe it is related to the 415 error.
But even with subs disabled - most of them keep downloading the mpd..
hmmmm.

Just an idea:
What would be smart is making the number of retries based on the length of the segment.
If its a short segment - no worries - we will get it on the next try.
If it's longer (20 seconds for example) - maybe worth a retry.

@glennguy  actually has a IA dev env setup and can make / debug changes so maybe he can look into this for us :)

matthuisman:
doing a test if we set the ret value to true
Reason for this is this line:
https://github.com/peak3d/inputstream.adaptive/blob/86c3a32c833189a95a861176133c0440d3371f5c/src/common/AdaptiveStream.cpp#L99

I think _stopped may signal this segment set is finished - dont try further segments.
By forcing ret to true - it won't set stopped.
then I can test and see if we have the behaviour we want

Issue #413, opened 2020-04-02
 Unable to create SIDX from IndexRange bytes, unable to select stream


rjsg42:
I'm trying to develop a plugin for a VOD service in Spain. The DRM license retrieval works, but the stream does not start.

The service uses MPEG-DASH with widevine.

MPD: https://media-spanishflix.partner.maginepro.com/ca1a3f04-604c-42a4-ae91-e02d6fcabe7c/AccurateManifest.mpd

LOG:

2020-04-02 13:50:41.494 T:16729   DEBUG: AddOnLog: InputStream Adaptive: License update successful
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: Key Status (7):
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> LicenseDurationRemaining -> 21600
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> LicenseType -> Streaming
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> PersistAllowed -> False
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> PlayAllowed -> True
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> PlaybackDurationRemaining -> 9223372036854775807
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> RenewAllowed -> False
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> RenewalServerUrl ->
2020-04-02 13:50:41.495 T:16729   DEBUG: AddOnLog: InputStream Adaptive: License update successful
2020-04-02 13:50:41.496 T:16729   DEBUG: AddOnLog: InputStream Adaptive: Key Status (7):
2020-04-02 13:50:41.496 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> LicenseDurationRemaining -> 21600
2020-04-02 13:50:41.496 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> LicenseType -> Streaming
2020-04-02 13:50:41.496 T:16729   DEBUG: AddOnLog: InputStream Adaptive: -> PersistAllowed -> False
[...]

PR #450, opened 2020-05-16
[DASH] default timescale to seconds when not in segmenttemplate

glennguy:
Hi @peak3d 

This should fix the issue in #413. Test stream to try:
```
http://livesim.dashif.org/livesim/testpic_2s/Manifest.mpd
```

Please note there is also another issue affecting this test stream (unsigned int seg.range_end_ wrap below zero causing seg 404) however that is not the focus of this PR.

DASH spec says that timescale is optional in segmenttemplate, and should default to seconds if not given, having it at 0 was causing the wrong path of execution to be taken for this case.

I am worried though about the condition of tpl.timescale > 0 at https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/parser/DASHTree.cpp#L1118 . Since this commit will make timescale always greater than 0, would you be able to advise if this will affect a case of mpd that I'm not considering here?

Thanks!

Glenn

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -110,6 +110,9 @@ static unsigned int ParseSegmentTemplate(const char **attr, std::string baseURL,
     attr += 2;
   }
 
+  if (!tpl.timescale) // if not specified timescale defaults to seconds
+    tpl.timescale = 1;
+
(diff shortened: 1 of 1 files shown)


peak3d:
I'll merge it into Matrix, but do not backport for the 18.7 release (too risky)

PR #448, opened 2020-05-15
Parse frame rates from HLS master playlists

glennguy:

PR #451, opened 2020-05-18
[HLS] Copy fpsScale for new periods

glennguy:
fixes issue mentioned in #448 where we get for eg. 23976 fps instead of 23.976 in periods > 0


Ronny-nerd:
Thanks glennguy...it's stable for Android 9.

ksooo:
Confirming the fix working.

peak3d (approved):

PR #456, opened 2020-05-26
[WebVTT] Fix binary representation

mediaminister:
This fixes an issue where in some cases binary WebVTT subtitles are not shown or suddenly disappear.

The current implementation looks at the text content of the previous subtitle to decide if the next subtitle needs to be added to the vector. This is a bad practice as it sometimes throws away subtitles that should be displayed.
For instance, when someone is calling a name, it's totally possible that  the name is called several times in a row:
```
00:08:08,888 --> 00:08:10,368
Malu?

00:08:12,928 --> 00:08:14,488
Malu?
```
In this case **Malu?** should be displayed twice.

Checking if the previous subtitle end timestamp is set to decide if we need to add the new subtitle is a more failsafe approach.

### DIFF
--- a/src/parser/WebVTT.cpp
+++ b/src/parser/WebVTT.cpp
@@ -42,7 +42,7 @@ bool WebVTT::Parse(uint64_t pts, uint32_t duration, const void *buffer, size_t b
         cbuf += 4, buffer_size -= 4;
 
       std::string text(cbuf + 12, buffer_size - 12);
-      if (m_subTitles.empty() || text != m_subTitles.back().text[0])
+      if (m_subTitles.empty() || ~m_subTitles.back().end)
       {
         m_subTitles.push_back(SUBTITLE(pts));
         m_subTitles.back().text.push_back(text);


MastaG:
LGTM

src/parser/WebVTT.cpp
```diff
@@ -42,7 +42,7 @@ bool WebVTT::Parse(uint64_t pts, uint32_t duration, const void *buffer, size_t b
         cbuf += 4, buffer_size -= 4;
 
       std::string text(cbuf + 12, buffer_size - 12);
-      if (m_subTitles.empty() || text != m_subTitles.back().text[0])
+      if (m_subTitles.empty() || ~m_subTitles.back().end)
```
peak3d, line 45: The check was implemented to avoid double subs if same sub is provided in 2 segments (last of segment x and first in segment x+1) This is always true, where subs intersects the segment.

Instead this approach pls. simply add a comparision for the timestamps ( || back.start != start)

peak3d (changes requested):

src/parser/WebVTT.cpp
mediaminister: Okay, I changed it.

mediaminister: I can't get it to work with a timestamp comparison.
Tested with:
```
xbmc.executebuiltin('PlayMedia(plugin://plugin.video.vtm.go/play/catalog/episodes/0bc55667-3c4e-44a9-9d26-ba8f861e525b)')
```

peak3d: `if (m_subTitles.empty() || text != m_subTitles.back().text[0] || pts != m_subTitles.back().start)`
is not working?

mediaminister: No, it is not working.

mediaminister: Can you explain what is wrong with my approach, checking if the previous subtitle has an ending timestamp.

peak3d: For my example: identical sub in 2 segments, same start, and same end.
subtitles.back contains valid start / end, now the identical second one arrives, and `~m_subTitles.back().end)` is true, therefore the duplicate will be inserted, but shouldn't.

What is "not working" ? Does it not compile? because in your example back().start and pts should be different, (and back().start is always set), I don't see any eason, why it should not work.

Debug it pls.

mediaminister: >  identical sub in 2 segments, same start, and same end.

Is this even possible with binary webvtt, subtitle data is packaged in `vttc`,  `payl`, `vtte` iso boxes. I thought there is always an empty `vtte` box separating two subtitles and pts continuously increases over time and is used to get start and end timestamps.

It obviously does compile, but only the first subtitle after starting the stream is showed on screen. I'll debug it.

peak3d: @mediaminister: I'm short before releasing a new version, because next kodi version is soon.
Whats the title of the vtm.go stream? Then I can try to support here....

mediaminister: The stream where I found the problem is not online anymore. But you can try with Vlaamse series -> Coppers - Aflevering 3 to test.

peak3d: @mediaminister for the stream you provided first, external subtitles are used.
How do you force using internal subtitles? Is my vtm.go version too old?

mediaminister: You need the most recent version in official Kodi repo, internal subtitles are used when available. External subtitles are only used when there are no internal subtitles.

peak3d: @mediaminister I don't

peak3d: @mediaminister I debugged, and your fix is fine, thx!

peak3d (approved):

src/parser/WebVTT.cpp
mediaminister: Thx.

PR #461, opened 2020-06-06
Refactor processEncryption & Regression Fix

matthuisman:
This removes a lot of nesting etc and makes it a bit easier to read.
Also makes it easier to add new supported encryptions in the future.

Easier to see what it does by viewing the entire updated function here:
https://github.com/peak3d/inputstream.adaptive/blob/00850dbdff6d66e4142697b733cdc0056814d178/src/parser/HLSTree.cpp#L97

This will also fix regression issue here:
https://github.com/peak3d/inputstream.adaptive/pull/395#issuecomment-640109313
where IA is treating a HLS AES-128 stream with a keyformat as non-encrypted.

PR #462, opened 2020-06-06
Fix if aes-128 has keyformat

matthuisman:
If AES-128 has a keyformat, it would pass as ENCRYPTIONTYPE_UNKNOWN

https://github.com/peak3d/inputstream.adaptive/pull/461 also sorts this out.
But this has less changes so may be preferred.

@peak3d 
Take your pick :)

Once you've picked, I can backport PR if you like?

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -108,35 +108,32 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
       Log(LOGLEVEL_ERROR, "Unsupported encryption method: %s", map["METHOD"].c_str());
       return ENCRYPTIONTYPE_INVALID;
     }
-    if (!map["KEYFORMAT"].empty())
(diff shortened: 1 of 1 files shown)


peak3d:
I'm fine with #461 (if it still works)

matthuisman:
do you want it to return ENCRYPTIONTYPE_INVALID or ENCRYPTIONTYPE_UNKNOWN when it doesn't find a supported method?

ENCRYPTIONTYPE_INVALID means that IA will fail straight away
ENCRYPTIONTYPE_UNKNOWN lets IA still try (or move to the next ext-key)

Originally I had ENCRYPTIONTYPE_UNKNOWN to allow for supported keys to be in the file after non-supported as Apple TV+ has that.

It would mean IA doesn't fail straight away though. it would try to play Sample-aes for example.
But I think thats OK? there would still be the error logged about unspported encryption

peak3d:
I have to check tomorrow

matthuisman:
cool. I'll close this PR and we can discuses on the other :)

PR #474, opened 2020-06-23
[Android] only request and log (max)securitylevel and querykeystatus...

Maven85:
... if playready is not used

This PR fixes a Kodi crash when playing a stream on android and PlayReady is used.
kodinerds.net users have confirmed this fix for Shield TV and Philips ATV.


peak3d (approved):
thx @Maven85

PR #477, opened 2020-06-26
[HLS] EXT-X-DISCONTINUITY-SEQUENCE support

glennguy:
Hi @peak3d 

I only had to modify 2 lines in adaptivestream.cpp to have success in period transitions (same as before the rework branch)

There is a new member variable for Period:  `sequence_` for storing the discontinuity sequence number of a period.
I decided against using Period->id_ as I have seen several examples of mpd manifests that use characters or even whole words here, so I believe a future implementation for live dash multiperiod would be better off not directly using this.
The sequence number of the first period on initial playback is also stored in AdaptiveTree, this is because because with hls workflow the streamid will always be 1001, 1002 etc. but this period's sequence will not always remain 0. This way we can still know which period relates to streamid 1001, 1002...

In HLSTree.cpp there are a few things of note apart from the obvious extra parsing. Sorry if it's already obvious to you but just in case there's any concern...

RefreshLiveSegments no longer iterates through periods, only uses current_period_. Because of the logical layout of HLS where periods are described inside of representations we would otherwise end up checking the same thing multiple times. And as we can never delete current_period_ it avoids a potential crash.
We also prepare a list of representations to refresh and do after finishing iterating to again avoid erasing elements while iterating.

In the example strm below it's quite easy to pause for 5 minutes and end up with a condition where current_period_ is not longer a part of the playlist. current_period_ is removed from periods_ so it doesn't interfere with parsing the playlist then once finished parsing it is inserted at the beginning of periods_ to avoid a memory leak.

Manual stream switching was working fine before the rework branch and still is now however I've noticed that in its current form all streams point to the same representation: 
image
I'm sure you know about this, just writing for the benefit of other readers.

strm file for testing (period count fluctuates between 1 and 7, occasionally there will be a 10 minute featurette)
```
https://dai.google.com/linear/hls/event/YakHdnr_RpyszducVuHOpQ/master.m3u8
```

Thanks for your time and help with this!

Glenn

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -190,7 +190,7 @@ bool AdaptiveStream::start_stream(const uint32_t seg_offset,
                                   bool play_timeshift_buffer)
 {
   if (!play_timeshift_buffer && !~seg_offset && tree_.has_timeshift_buffer_ &&
-      current_rep_->segments_.data.size() > 1)
+      current_rep_->segments_.data.size() > 1 && tree_.periods_.size() == 1)
   {
     std::int32_t pos;
     if (tree_.has_timeshift_buffer_ || tree_.available_time_ >= tree_.stream_start_)
@@ -414,7 +414,7 @@ bool AdaptiveStream::ensureSegment()
       ResetSegment();
       thread_data_->signal_dl_.notify_one();
     }
-    else if (tree_.HasUpdateThread())
+    else if (tree_.HasUpdateThread() && current_period_ == tree_.periods_.back())
     {
       current_rep_->flags_ |= AdaptiveTree::Representation::WAITFORSEGMENT;
       Log(LOGLEVEL_DEBUG, "Begin WaitForSegment stream %s", current_rep_->id.c_str());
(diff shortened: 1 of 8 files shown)


glennguy:
Also I forgot to mention that still outstanding is being able to start a stream from the last period. I haven't had any further luck with this.

xluxeq:
Eta for PR? Crashes Kodi without this: https://forum.kodi.tv/showthread.php?tid=315513&page=27

peak3d (approved):
Looks like something which must be tested heavily before we release.......

peak3d:
Before release we (you??) should look at the 4-times display when using manual. Never noticed it before....

glennguy:
@peak3d thanks

The 4 x same stream was an issue during development when this was based against the rework branch but was solved in that branch at some point anyway.

I've only got one outstanding issue and I think its to do with VideoPlayer. If a new period starts but the segments haven't yet downloaded (thanks to wonderful Australian internet) a situation happens where Kodi thinks we're out of sync for a time equal to elapsed playback time i.e if stream has been playing for 10 minutes, Kodi will take 10 minutes to recover... As far as I can see everything in IA is still correct state when this happens. I'll try to reproduce again and post issue in xbmc.

Issue #443, opened 2020-05-11
Please clarify the stream selection: Manual video option string


john-tho:
In playing a video, I want the most suitable (to my other settings) stream selected, but with the ability to also select a different stream during playback. I would pick a setting name for this as something like:
- _Auto with video override_
- _Auto with video selection_
- _Manual with auto video selection_

_Manual video_ (c2c187bff4f1b25b84e09383401aa3409182531b) does this, after 87d4c4629e9ec408d7e5d231ec1a67fcb7cf6411 fixed a typo? (2.5.6-Matrix, 2.4.3-Leia)

I only found this through trial and error in Kodi. It seemed for `STREAMSELECTION`:
- _Auto_ did not allow any stream selection
- _Manual_ selected the first stream
  I gave up here until I looked at the source…
- _Manual video_ selected the most suitable stream as default, but also allowed stream selection.

Setting in question:
https://github.com/peak3d/inputstream.adaptive/blob/0d3bd30511c53003b96e01611c20c1e885ef6d65/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po#L112

Apologies for the trivial issue, and not indicating that my terminology is the best, but the existing could be improved to indicate the sensible default is selected?

Otherwise, the software works brilliantly for me, thank you very much.

Cheers

CastagnaIT:
i do not full understand how the code of _manual_ option works
but this is what Peak3d say to me long time ago:

> This is what "manual" is doing: provide every representation to kodi.

referred to dash representations

then if the numbers of streams representations exceede the Kodi limits, Kodi will cause a lack of streams

the kodi limits on Kodi <= 18.x is of 32 total streams
the kodi limits on Kodi >= 19.x is of 256 total streams

For example:
on Kodi 18.x if the user enable the "manual" option will accuse missing of audio/subtitle languages,
this because of too streams passed to Kodi example:
10 video streams
10 audio/subtitle languages
-for every audio streams 5 bandwidths and 2 type of channels
-for every subtitle normal and forced stream type

This will results in a:
10 video streams
100 audio streams
20 subtitle streams
Total of 130 streams

exceed Kodi <=18.x limits and this is a problem

but there are side effects:
- on audio selections window, all audio streams have same name, the final user do not understand why
- the auto audio language selection of kodi works bad, selects the first stream in language, but there is no distinction with bandwidth

for a novice user your suggestion is not very understandable
I would rather write something about debugging or warning for "manual" mode
[...]

PR #480, opened 2020-07-03
[Matrix] Language improvements

CastagnaIT:
As discussed on #443
i have changed the text descriptions
for ids: 30155 30156 30160
to allow users to better understand the meanings

In addition there are some other fixes as wrong folder name and positions
and used same scheme to all files to allow make fast comparisons between files translations
and other minor fixes

Please merge before any other language translations PR update


dagwieers:
@CastagnaIT What is the advantage of adding those comments?

I think it is best that if we plan to modify the English translations we also invalidate (clear) the translations (or if not invalidated, mark them as a FIXME: instead).

CastagnaIT:
> What is the advantage of adding those comments?

making comparative checks from en file to every translation is more immediate
otherwise the comparison software marks comments as difference and you waste your time for nothing

> I think it is best that if we plan to modify the English translations we also invalidate (clear) the translations (or if not invalidated, mark them as a FIXME: instead).

?!? translation are cleared yet
translators only need to update

peak3d (approved):

PR #483, opened 2020-07-15
Kodi API related code update

AlwinEsch:
Changes related to https://github.com/xbmc/xbmc/pull/17438

The commits here match mostly the flow on Kodi's request.

Background about is:
- Fix possible mem leaks
  - Before can in API  by "C" structure easily come a memleak as the delete was not always done
- Use the before with PVR brought way also by inputstream 
  - There are some "C" structures now in background handled by "C++" classes (also about memleaks)
  - Separate the "C" headers from "C++" to have base for other languages and to confirm clean ABI
- To have the documentation of API more usable (not addon code here related)
- Prepare the dev-kit folder more to have independent
  - About TM problems, code size and independent usage

Runtime test with Netflix was OK.


AlwinEsch:
Is now updated to have working on reworked inputstream system.

Netflix and some other test streams are working.
If nothing comes against, come request on Kodi in.

peak3d (approved):

PR #486, opened 2020-07-19
Don't log unknown encryption methods

matthuisman:
This log message was introduced in my PR to refactor the processEncryption method.

However, it has caused confusion when Disney+ playback didn't work and they thought the below log messages were the cause.

> WARNING: AddOnLog: InputStream Adaptive: Unknown encryption method: SAMPLE-AES-CTR with keyformat com.microsoft.playready
> WARNING: AddOnLog: InputStream Adaptive: Unknown encryption method: SAMPLE-AES-CTR with keyformat PRMNAGRA

So, let's just revert to old behaviour of not logging on unknown methods.

Also, on live streams, the above could spam the log every time the playlist is processed.

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -155,9 +155,6 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
     return ENCRYPTIONTYPE_INVALID;
   }
 
-  // UNKNOWN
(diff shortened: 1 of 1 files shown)


peak3d (approved):

peak3d:
Not mergeable, pls. fix

matthuisman:
@peak3d 
ready to merge now

peak3d (approved):

PR #488, opened 2020-07-23
[HLS][Revert] share parameters between master / sub playlists

smf007:
Addresses the 403 errors noted in Issue #184.

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -244,11 +244,6 @@ bool HLSTree::processManifest(std::stringstream& stream, const std::string& url)
         else
           rep->source_url_ = res->second;
 
-        if (!manifest_parameter_.empty() &&
-            rep->source_url_.compare(0, base_url_.size(), base_url_) == 0 &&
-            rep->source_url_.find('?') == std::string::npos)
-          rep->source_url_ += manifest_parameter_;
-
         // default to WebVTT
         if (type == SUBTITLE)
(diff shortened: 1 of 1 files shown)


matthuisman:
This is interesting and something that could have been easily overlooked on a lot of 404 stream errors.

I confirmed Bitmovin and kodi ffmpeg both don't send the query string to the subplaylists or EXT-MEDIA.
If query strings are needed for them, the playlist should already have them in it

Test Matrix Builds Here:
https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/PR-488/3/artifacts

matthuisman:
maybe do a Leia backport PR as well.
Not sure if there will be new Leia IA version, but good to have it ready in case

peak3d:
Even I'm quite sure that this will lead to regression for other streams, I'll let it in if the changes in AdaptiveTree are reverted.
I know that this is maybe currently dead code, but I'm sure that we need it later on.

peak3d (changes requested):
Even I'm quite sure that this will lead to regression for other streams, I'll let it in if the changes in AdaptiveTree are reverted.
I know that this is maybe currently dead code, but I'm sure that we need it later on.

smf007:
Thanks for the input Markus.  Updated as requested, and tested with my problematic CBC (Canadian Broadcast Corporation)  streams.

peak3d (approved):

PR #489, opened 2020-07-23
[Leia][Backport][HLS][Revert] share parameters between master / sub playlists

smf007:
Backport of pull #488 .

### DIFF
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -245,11 +245,6 @@ bool HLSTree::processManifest(std::stringstream& stream, const std::string& url)
         else
           rep->source_url_ = res->second;
 
-        if (!manifest_parameter_.empty() &&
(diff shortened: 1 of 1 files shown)


matthuisman:
I have gone ahead and done the cleanup that will make this PR obsolete:
https://github.com/peak3d/inputstream.adaptive/pull/492

peak3d (changes requested):
Even I'm quite sure that this will lead to regression for other streams, I'll let it in if the changes in AdaptiveTree are reverted.
I know that this is maybe currently dead code, but I'm sure that we need it later on.

peak3d:

smf007:
Updated as requested.  Is consistent with #488 and tested as noted in #488 with problematic CBC streams.

Scott

peak3d (approved):

Issue #473, opened 2020-06-18
No playback after redirects (MPD & HLS)


matthuisman:
This is not a new issue or regression. As far as I know, they have never worked.

However, dai.google.com urls are starting to redirect to urls for including ads etc, and that breaks playback if those mpds have baseurls

***HLS***

**Test without redirect (WORKS)**
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_redirected/hls_direct.strm

**Test with redirect (DOES NOT WORK)**
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_redirected/hls_redirected.strm

IA tries to get
`https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_redirected/matthuisman/ia_tests/master/hls_aes128_single_key/index_17_2744763.ts`

instead of correct
`https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_aes128_single_key/index_17_2744763.ts`

The test playlist used above
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_redirected/master.m3u8
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_redirected/sub.m3u8

Note:
Redirects with segments that don't start with / seem to work ok.

---

***MPD***

**Test without redirect (WORKS)**
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_redirected/mpd_direct.strm

**Test with redirect (NOT WORK)**
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_redirected/mpd_redirected.strm

IA tries to get:
`https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_redirected/audio/mp4a/2second/tears_of_steel_1080p_audio_32k_dash_track1_init.mp4`

Instead of correct:
`https://dash.akamaized.net/dash264/TestCasesIOP33/Content/audio/mp4a/2second/tears_of_steel_1080p_audio_32k_dash_track1_init.mp4`

The test mpd used above
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_redirected/mpd_redirected.mpd

Note:
Redirects to a MPD without a BaseURL seem to work fine.

smf007:
Regarding the HLS issue, seems to be related to the variable "base_domain".  That variable is populated in routine AdaptiveTree::PreparePaths.  When I get a chance this week, I'll add some debugging to see what's going on.

[...]

PR #490, opened 2020-07-23
[HLS] Update base_domain when an effective_url (i.e. redirect) is being used.

smf007:
Fixes HLS problems of Issue #473

When a redirect is used, the base_domain should be updated to reflect the domain of the effective URL.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -303,6 +303,14 @@ bool adaptive::AdaptiveTree::download(const char* url,
 
     if (effective_url_ == base_url_)
       effective_url_.clear();
+
+    if (!effective_url_.empty())
+    {
+      base_domain_ = effective_url_;
+      paramPos = base_domain_.find_first_of('/', 8);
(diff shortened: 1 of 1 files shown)


matthuisman:
I have gone ahead and done the cleanup that will make this PR obsolete:
https://github.com/peak3d/inputstream.adaptive/pull/492

peak3d (changes requested):
This change is against the concept, please add soething like this:

https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/parser/HLSTree.cpp#L430

at this place:

https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/parser/HLSTree.cpp#L167

and tell me if this solves the issue. If not, pls. provide a kodi log.

peak3d:

peak3d:
sould be fixed by this: 8bea82b3c323ef30b30806724304ce91aa86d7ae

PR #492, opened 2020-07-24
Fix / Clean Up Effective (redirected) URLS

matthuisman:
no need to track both original and redirected url.

Simply download the manifest then PreparePaths after using the effective url

This removes a lot of need to url.replace etc further on.
Remove effective_filename_ as no longer used and scanEffectiveURL not required either
Cleaned up the Renewed url.

Have tested this on as many streams / add-on as I could.

It even worked with a redirect > master playlist then has a redirect > sub playlist

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -368,21 +368,7 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment* seg)
   else
     download_headers_.erase("Range");
 
-  if (!tree_.effective_url_.empty() && download_url_.find(tree_.base_url_) == 0)
-    download_url_.replace(0, tree_.base_url_.size(), tree_.effective_url_);
-
   return true;
-}
-
-std::string AdaptiveStream::buildDownloadUrl(const std::string& url)
-{
-  if (!tree_.effective_url_.empty() && url.find(tree_.base_url_) == 0)
-  {
-    std::string newUrl(url);
-    newUrl.replace(0, tree_.base_url_.size(), tree_.effective_url_);
-    return newUrl;
-  }
-  return url;
 }
 
 bool AdaptiveStream::ensureSegment()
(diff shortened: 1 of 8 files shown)


michaelarnauts:
@matthuisman does this also fixes https://github.com/peak3d/inputstream.adaptive/issues/286 ?

I can't see much mention of the Location element in your pr, so I'm not sure if this is the same redirect.

It's marked as closed, and had an fix merged in the past, but that fix doesn't seem to work.

matthuisman:
This doesn't touch location at all.
But I can look into that :)
Any test strm?

Does location meant to just point to a different mpd? Nothing more special than that? Just needs to be treated like a redirect?

michaelarnauts:
Vtm go does a workaround for this. I just fetch the actual mpd in the add-on, extract the url in the location tag, and then pass this to Inputstream.

You could probably use the example from the issue and have it redirect to another mpd to test. Or you could use the vtm go addon to test.

matthuisman:
Thanks. So sounds like it needs to work like a redirect. If location, then grab that mpd instead. I've also seen location even in the redirected mpd, so a check to make sure it isn't the same as current URL before redirecting would be needed

michaelarnauts:
Thanks for looking into this. This will speed up the playing in my addon a bit :)

@mediaminister looked at this issue a while ago, so it might just be a bug in his implementation: https://github.com/peak3d/inputstream.adaptive/pull/297

smf007:
@matthuisman  looks good.  Would be nice to know if there was some intent to the original design that we are missing though.

smf007:
Are there any details why you closed this pull?

matthuisman:
It broke mpd with relative URLs.
A new fixed version I did yesterday is merged into our new Inputstream Adaptive Testing git (https://github.com/johnny5-is-alive/inputstream.adaptive.testing) Once fully tested in the wild, I'll PR again.

PR #499, opened 2020-08-18
Use CDllHelper for opening ssd_wv

glennguy:
Hi Peak3d

This should pave the way for being able to update inputstream.adaptive running on Android rather than relying on it being packaged.

Not included is changing the searchpaths for ssd_wv, I thought I'd ask for some input of yours first. What works is changing to `special://home/addons/inputstream.adaptive/` for Android, I'm mindful that this would need to be coordinated with the repo. Can add this to the PR if you want.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,7 +39,6 @@
 #include <stdio.h>
 #include <string.h>
 
-#include <dlfcn.h>
 #include <kodi/Filesystem.h>
 #include <kodi/General.h>
 #include <kodi/StreamCodec.h>
@@ -47,6 +46,10 @@
 
 #if defined(ANDROID)
 #include <kodi/platform/android/System.h>
+#endif
+
+#ifdef CreateDirectory
+#undef CreateDirectory
 #endif
 
 #define DVD_TIME_BASE 1000000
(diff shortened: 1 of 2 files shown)


peak3d:
@glennguy from my current knowledge .so files in Android must reside in application's lib folder to be able to execute them.
Issue is currently that the AddonManager only copies the main (ia) .so file into this durectory, copying ssd_wv is not implemented.

What is DLLHelper doing different? Is it copying the .so file into the correct folder (.lib) ?

glennguy:
That's correct, DLLHelper will do the same as AddonManager.
main .so file - https://github.com/xbmc/xbmc/blob/master/xbmc/addons/binary-addons/AddonDll.cpp#L53
DllHelper - https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-dev-kit/include/kodi/tools/DllHelper.h#L122

With the above mentioned change to the searchpaths it makes it possible to provide updates on Android, both ia .so and libssd_wv.so are 'cached' to the lib folder.

@AlwinEsch can confirm

AlwinEsch:
Yes the dllhelper was thought by to make on audiodecoder addons more easy to load a dll and later also to have usable on Android.

Tested there on Android with e.g. https://github.com/xbmc/audiodecoder.timidity/blob/Matrix/src/TimidityCodec.cpp#L55-L67

peak3d:
@glennguy is this PR a WIP? Or can it go in as it is?

peak3d:
@glennguy can we now skip this: https://github.com/peak3d/inputstream.adaptive/blob/Matrix/CMakeLists.txt#L48-L52 ?

glennguy:
I removed the dlfcn.h include in main as it's included in DllHelper.h - to me it made sense as we are no longer directly accessing functions from dlfcn.h in main now.

This is a PR, ready to go and will not affect the nightly Kodi builds.

At the time which you want to decouple IA from the Android packaging then the searchpaths will need to be modified, plus seeing android builds in https://mirrors.kodi.tv/addons/matrix/ along with the appropriate changes to remove it from Android packaging which I cannot change from my end.

I've tested on Android by changing the searchpath as mentioned above and it all works well, widevine content plays and the libssd_wv.so file is cached to the proper location and loaded correctly.

Let me know if you want any changes :)

peak3d:
thx @glennguy !

ksooo:
@glennguy could you open a PR with the missing pieces to get rid of the need to package inputstream.adptive on Android?

glennguy:
@ksooo sure thing, I will get it sorted later today

ksooo:
Cool, thanks.

PR #506, opened 2020-08-30
Network toggle + Display resolution update- 2

spyderboy92:


glennguy:
@spyderboy92 we will get to this some day soon!

glennguy:
@spyderboy92 @peak3d thanks for all your hard work, code is finally in with #732!

PR #507, opened 2020-08-31
[Matrix] API related change

AlwinEsch:
This change is related to request on Kodi here: https://github.com/xbmc/xbmc/pull/18352

And to make a reduced and splitted change about API cleanup (related to https://github.com/xbmc/xbmc/pull/17438).

As note from last time, is a Version increase not included, but addon need new release after them on Kodi is in!


AlwinEsch:
Can be come in.

peak3d (approved):

PR #508, opened 2020-09-02
Fix / Clean Up Effective (redirected) URLS

matthuisman:
This makes redirected playlists work correctly.
fixes issue: https://github.com/peak3d/inputstream.adaptive/issues/473

no need to track both original and redirected url.

Simply download the manifest then PreparePaths after using the effective url

This removes a lot of need to url.replace etc further on.
Remove effective_filename_ as no longer used and scanEffectiveURL not required either
Cleaned up the Renewed url.

Have tested this on as many streams / add-on as I could.

It even worked with a redirect > master playlist then has a redirect > sub playlist

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -368,21 +368,7 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment* seg)
   else
     download_headers_.erase("Range");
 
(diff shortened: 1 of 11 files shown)


peak3d (commented):
This is now the 4-th PR regarding effective-url, I suggest that we open an issue for discussing / evaluating what is not working and why (fom code POV).

peak3d:

peak3d:
The issue should be fixed: 8bea82b3c323ef30b30806724304ce91aa86d7ae

PR #535, opened 2020-10-20
Added audio/video format to SetFragmentInfo()

samnazarko:
Added audio/video format to SetFragmentInfo()

This adds support for informing a hardware based decoder running in the secure world of the stream properties.
It's necessary to prime the decoder before passing frames


peak3d:
Because of the API change between inputstream.adaptive and ssd_wv the version must be bumped:
https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/SSD_dll.h#L24

Otherwise ( if xchanging only either of them) will lead to segfaults.

samnazarko:
Thanks for the heads-up. We will update our ssd_WV accordingly.

CastagnaIT:
This PR is 2 year old...
looks incomplete and not more valid since bento is not anymore builtin in main repo

I close it, and if / when you will need similar changes please make a new one

PR #536, opened 2020-10-20
Return secure path and decoder flag

samnazarko:
The secure path is used to indicate that we support trusted video pathway and can handle Widevine L1.
The decoder flag indicates that we can decrypt Widevine L1 content.

At the moment, if the stream is L3, InputStream adaptive removes the decoder flag.

This causes the IsEncrypted() flag to report that the stream is encrypted, even after we have decrypted it because it only checks for the secure path flag. This patch changes this behaviour so that isEncrypted() returns the correct value.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1261,6 +1261,7 @@ public:
   bool IsEncrypted() const override
   {
     return (m_decrypterCaps.flags & SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_PATH) != 0 &&
+           (m_decrypterCaps.flags & SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_DECODER) != 0 &&
            m_decrypter != nullptr;
   };
   bool GetInformation(INPUTSTREAM_INFO& info) override


peak3d:
This change will brake Android L3 playback, because SECURE_DECODER is not set, but kodi / mediacodec still requires encrypted data (we don't encrypt in IA for android).
Unfortunately naming conventions are not nice, but SECURE_PATH indicates that IA passes encrypted frames to kodi, and SECURE_DECODER indicates that we need hardened security level (L1 for widevine).

If SECURE_DECODER is set, we select an secure decoder AND we initialize the render-surface with secure flag (==HDCP).
This is done in kodi::mediaCodecSurface Video decoder.

For your usecase you already send decrypted frames / handles, so we have to look for another way to signal, that your data is a handle from secure decryptor.

kszaq:
@samnazarko Is this fixed in any way by https://github.com/xbmc/inputstream.adaptive/pull/662? This PR disabled removing SECURE_DECODER flag.

samnazarko:
@kszaq -- unfortunately the test environment for Widevine is based on 18.x currently, so we can't verify it yet against the latest IA on Matrix.

CastagnaIT:
@samnazarko also this PR is 2y old,
if it is still required please observe the comments and make appropriate changes / tests for Nexus platforms/systems
otherwise we close the PR and if / when you have the need you can reopen it

samnazarko:
Let’s close this for now. We are re-working the Widevine implementation again and may take a different approach.

PR #545, opened 2020-10-30
Remove static declaration of 't' variable in SegmentTimeline

### LINKED ISSUE #543: IA is using a wrong initial segment the 2nd time a stream is played
I'm noticing an issue when playing a widevine protected stream. The first time the stream is played, it works as it should, but the 2nd time, I can see that IA is requesting wrong segments that don't exist.

Open PR: https://github.com/add-ons/plugin.video.viervijfzes/pull/47

**How to reproduce:**
Use above PR, enter credentials (you can register freely, or ask me on slack) and go to Channels > ZES > Catalog > NCIS: New Orleans > Season 3 > Episode 3x08 (for example, it happens with all episodes)

**Kodi debug log:**
[kodi.log](https://github.com/peak3d/inputstream.adaptive/files/5457174/kodi.log)

manifest: https://github.com/add-ons/plugin.video.viervijfzes/files/5443644/NCIS_NEW_ORLEANS_3_8_F0250070.mpd.txt

@mediaminister already debugged it a bit:
I already found out that the `range_begin_`  of the (next) segment has a completely wrong value the second time you start a stream. I guess there is something wrong in the initialization logic in AdaptiveStream.cpp but I didn't find a good way to debug this yet. 

Tagging some people who already looked at this.
@glennguy @mediaminister

glennguy:
Fixes #543 
Some SegmentTimelines will have an implicit `t="0"` for the first `S` tag e.g
```
<SegmentTimeline>
  <S d="314"/>
  <S t="314" d="256"/>
  <S t="570" d="430"/>
  <S t="1000" d="261"/>
  <S t="1262" d="253"/>
  etc...
```

This change will allow for this first segment to have `range_begin_` correctly set to 0 for representations other than the first one parsed.

Further explanation here if needed: https://github.com/peak3d/inputstream.adaptive/issues/543#issuecomment-719534565

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -338,7 +338,7 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
             {
               // <S t="3600" d="900000" r="2398"/>
               unsigned int d(0), r(1);
-              static uint64_t t(0);
+              uint64_t t(0);
 
               for (; *attr;)
               {


michaelarnauts:
This indeed fixed the issue. I haven't noticed any issues with other streams but maybe peak3d can tell more if there is a reason for that to be a static var.

PR #550, opened 2020-11-09
increase version to 2.6.5 (API related)

AlwinEsch:
Related to xbmc/xbmc#18705 and https://github.com/peak3d/inputstream.adaptive/issues/548

Hi @peak3d request only as note and can you close if version increase done by you direct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# inputstream.adaptive (2.6.4)
+# inputstream.adaptive (2.6.5)
 
 This is an adaptive file addon for kodi's new InputStream Interface.
 
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="inputstream.adaptive"
-  version="2.6.4"
+  version="2.6.5"
   name="InputStream Adaptive"
   provider-name="peak3d">
   <requires>@ADDON_DEPENDS@</requires>
@@ -18,7 +18,9 @@
     <description lang="en_GB">InputStream client for adaptive streams</description>
     <description lang="es_ES">Cliente InputStream para flujo de datos adaptativos</description>
     <platform>@PLATFORM@</platform>
-    <news>
+    <news>v2.6.5 (2020-11-09)
+- Matrix API change to v3.0.1 - Fix wrong flags bit shift
+
 v2.6.4 (2020-10-31)
 - Fix protocol search in manifest URL
 


CastagnaIT:
memo: before publish another version to Kodi please check what happen with dash drm streams
with the last changes on Matrix, netflix streams no longer work (refer to windows build, not tested on android)

FYI: if it can help i tried revert all commits until commit 522cbb900bf6617ea329c70c6aab86f3a6293041 and keep the new IS interface, and the problem persist, i think it is something inherent to the new Kodi IS interface, more than this i am not able to do..

Another note: the problem happen randomly seem like a mem leak, something the videos are played but after 1 or 2 videos played the videos do not start anymore so ISA is freezed, in other situations when you open Kodi ISA will be freezed at first try

ksooo:
> netflix streams no longer work (refer to windows build, not tested on android

Android works here in general, but skipping sometimes leads to frozen video while audio is playing fine.

AlwinEsch:
> netflix streams no longer work (refer to windows build, not tested on android

About this has tested with Netflix on Linux and was working.
Where was the Windows error coming? Between Leia and Matrix (now), or between e.g. 2.5.6-Matrix and 2.6.0-Matrix ...?

**EDIT:** You mean was coming with https://github.com/peak3d/inputstream.adaptive/commit/522cbb900bf6617ea329c70c6aab86f3a6293041?

CastagnaIT:
mmmh not exactly,
on Leia all works good no problems here

The problems on Kodi Matrix has started after your Kodi PR https://github.com/xbmc/xbmc/pull/17438
and the adaptions commit to ISA addon,
at this point seem affect only Windows build

On Kodi Matrix windows build, i am able to play video with netflix only with:
ISA version 2.6.1, that is compatible with e.g. Kodi build of 20201018-6384782d-master-x64.exe

Then i thought that some pr added to ISA from the version 2.6.1 onwards was the problem,
then i tried to revert all commits until https://github.com/peak3d/inputstream.adaptive/commit/522cbb900bf6617ea329c70c6aab86f3a6293041
but these reverted commits are not the cause of the problems

By following the problem in the code, i see that the code freeze when Kodi call this method and to this point
https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/main.cpp#L3922 it stuck.
The strange thing is that can happen that the first video played start and the second one the code will be freezed,
or it will be freezes immediately at first video start

This is an example when the code is freezed: https://paste.kodi.tv/tezigurapo.kodi at line 635 Kodi is stalled on loading screen

AlwinEsch:
Thanks for the line on addon, yes maybe possible that comes from this place. The way there is very hacky (also before :smirk:).

There becomes the C structure pointer from addon side given to target as `SSD_VIDEOINITDATA` possible that it can bring size and pointer position failures.

On my Linux/Android/OSX (gcc/clang) it can then equal between the two and works, but maybe under Windows by Visual Studio the `SSD_VIDEOINITDATA` now different and make problems.

the-dreamer:
not fully sure if somebody has RPI on the radar, but 2.6.4 is also not starting any drm video. i am on libreelec master from yesterday with 2.6.4 and no drm video from netflix/prime can be started.

ksooo:
@CastagnaIT fyi https://github.com/xbmc/xbmc/pull/18809

matthuisman:
jenkins build this please

CastagnaIT:
@ksooo thank you for this, let's start with Christmas gifts!
at the moment i can't start a Kodi build for Windows to do a test
i will do in the coming days

PR #563, opened 2020-12-01
Seperate ac3 (DD) and eac3 (DD+)

matthuisman:
Not sure why eac3 (Dolby Digital Plus) and ac3 (Dolby Digital) were both being set to eac3.

There may be a valid reason, but just in case it was an oversight, here is a PR.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1304,7 +1304,10 @@ public:
       case AP4_OTI_DTS_MASTER_AUDIO:
       case AP4_OTI_DTS_EXPRESS_AUDIO:
         info.SetCodecName("dca");
+        break;
       case AP4_OTI_AC3_AUDIO:
+        info.SetCodecName("ac3");
+        break;
       case AP4_OTI_EAC3_AUDIO:
         info.SetCodecName("eac3");
         break;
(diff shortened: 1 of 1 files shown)


matthuisman:
@phunkyfish @glennguy 
Who would know a bit about codecs in the team?

phunkyfish:
Not really sure. What’s the question you want to ask?

matthuisman:
want to check Kodi handles ac3 and eac3 as codecs?

It appears it does:
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StreamUtils.cpp

that is main reason for this PR so EAC3 streams in IA can have priority over AC3
Also, if Kodi knows what it is - then pass-through settings should also work

tmm1:
This PR seems pretty sane to me.

phunkyfish:
If @tmm1 thinks it’s good then I’m happy.

matthuisman:
Great :)

here is some screenshots playing back HBO Max which has both AC3 and EAC3 streams

You will see without this, Kodi shows the AC3 stream as DD+ and uses ff-ac3
With this PR - correctly shown as AC3 and uses ff-ac3

I've also opened up a new issue here https://github.com/xbmc/inputstream.adaptive/issues/650
to remind us about eventually supporting Atmos and mp4a.40.5 (HE-AAC)

**Before:**
screenshot00000
screenshot00003
**After:**
screenshot00005
screenshot00002

matthuisman:
@glennguy 
this is also ready for review and merge.

glennguy (approved):

PR #564, opened 2020-12-02
Fix MPD Timing (remove publishTime & presentationTimeOffset)

### LINKED ISSUE #22: Channels not playing "Live"
I'm on a Windows 10 machine running Kodi 18.9 and inputstream,adaptive 2.4.5 and every channel starts from the beginning of the program -  not live. Also, when the program ends and sometimes randomly, the channel disconnects and goes back to the Kodi home screen.

### LINKED ISSUE #560: [MPD] Wrong segment number calculated
Unix timestamp of test:  approx 1606270484

kodi calculated segment: **383716778** (/video6/segment-383716778.m4s)
correct segment: approx **401567613** (/video6/segment-401567613.m4s)

Segments are 4s in length

Difference = 17,850,835 segments = 71,403,340 seconds (826 days)

The UTCTiming in the mpd file is correct to unix timestamp

MPD attached
[manifest8.mpd.txt](https://github.com/peak3d/inputstream.adaptive/files/5594054/manifest8.mpd.txt)

### LINKED ISSUE #615: IA 2.6.8 has broken stream that work on IA 2.4.5
Our addon LA7-LA7D (present on Official Kodi Repo) cannot work on last IA for Kodi19.

The original Live channel is here: 
https://www.la7.it/dirette-tv
p.s. it's free, without registration, and not geoblocked

We tested it by old IA on Kody18 and it work, for example on IA 2.4.5 and also on IA 2.3.3

matthuisman:
Fixes: https://github.com/xbmc/inputstream.adaptive/issues/560
Fixes: https://github.com/xbmc/inputstream.adaptive/issues/615
Fixes: https://github.com/d21spike/plugin.video.sling/issues/22

https://github.com/xbmc/inputstream.adaptive/commit/dc7b42783d18c65ebcf205a4979a1c9ea6b825e7 seems to have broken quite a few streams. 
This PR basically reverts that

Looks like the problem is that it is using publishTime as a basis to calculate segment numbers.
"MPD@publishTime is merely a version label. The value is not used in timing calculations."
This PR reverts back to using dash->stream_start_ which is based off kodis clock.

Confirmed fix for BBC: https://github.com/xbmc/inputstream.adaptive/issues/560#issuecomment-809611820 (publishtime)
Confirmed fix for LA7: https://github.com/xbmc/inputstream.adaptive/issues/615#issuecomment-809268375 (publishtime)
Confirmed fix for Bein Mena (needed publishtime and presentationTimeOffset removed)
Confirmed fix for Sling: https://github.com/d21spike/plugin.video.sling/issues/22#issuecomment-812855081


matthuisman:
ok, new build with presentationTimeOffset removed as Bein Mena was still broken.

**https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-564/18/artifacts**

@CaptainTK
can you please test above build with BBC (and anything else using live MPD's)

@Testato @fawkes15 @light-wizard
can you please test above build with LA7 (and anything else using live MPD's)

@eracknaphobia @ckuyehar @d21spike
can you please test above build with Sling on Matrix (and anything else using live MPD's)

light-wizard:
Tested android aarch64 build with LA7: all functions working.
Android 8, Kodi 19.
Thanks for the fix @matthuisman

Testato:
On LA7 all work, the two Live MPD and also the On-Demand MPD programs
Tested on Firestick

glennguy:
Before bringing this in it would be good to confirm that presentationTimeOffset is definitely not needed in any segment calculations. We've got some tests now sure, but I'd like to review the DASH-IF guidelines one more time to see what the intended purpose is.

I can't recall exactly but I remember getting the impression previously that PTO might only be used for calculating where segments are in regard to program time for purposes of a visual timeline like Kodi's OSD - example: 2 hours into a 4 hour live event with a 30 minute buffer and being able to tell the player to show 1:30:00 as the start of the timeline, 2:00:00 as the end. In this case I believe there's nothing in the inputstream interface for passing on this information anyway. Someone please correct if that's not the case.

On the other hand it is possible that PTO is used in calculations, just that most often it is nullified by another value of similar or same size that is subtracted/added to bring back to a net or near zero.

This is just me thinking out aloud, I'll look into properly over the weekend.

matthuisman:
Sounds good. The test mpd is one I know fails without ignoring pto. 

this PR should be our next top priority for merging as its fixing regressions from 2.4.6. 

And maybe the live / vod fix too.

Once they are merged. I'll PR for Leia this and the other regression fixes. 

Still waiting to hear test results from Sling users / dev

Oh... And I remember seeing other PTO code somewhere too.

Here it is (segment list attribute)
https://github.com/xbmc/inputstream.adaptive/blob/7b0ade1e0bc61743bcd1c36aa2412230ee6e03cd/src/parser/DASHTree.cpp#L441

CaptainTK:
Apologies, I did not get to test this the last days. 

https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-564/12/artifacts

looks empty to me, so if there is a newer version to test, let me know, please.

matthuisman:
Try this please:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-564/18/artifacts

CaptainTK:
Thanks for that! Looks good to me, played around with a number of live channels, all worked well.

ckuyehar:
> @eracknaphobia @ckuyehar @d21spike
> can you please test above build with Sling on Matrix (and anything else using live MPD's)

@matthuisman - i'd like to test but im unsure how to test this fix - any quick guide/setup document to setup the test?

matthuisman:
Sling fix confirmed:
https://github.com/d21spike/plugin.video.sling/issues/22#issuecomment-812855081

Sling also requires the live fix:
https://github.com/xbmc/inputstream.adaptive/pull/626

d21spike:
@matthuisman I appreciate the hard work!

glennguy:
Based on my understanding, PTO is meant to be used to help align video/audio/text streams where the segments aren't aligned.
This is an example case where the start of a period is part way through a segment (in orange):
image

Consider now that this is for video segments, but in the case for audio segments in this playlist they are perfectly aligned to the period start. PTO will signal to the player that the start of the period is 8220 on the sample timeline, player will offset accordingly into the segment. Basically it allows sub-segment alignment. Please correct if I'm wrong in this.

Two things:
* IA only supports a simple list of segments that belong to a representation/period, there is currently no way to give a time or sample offset to apply a subsegment alignment. 
* There's no open issues that relate to out of sync audio/video streams. I can't be sure but I imagine VideoPlayer already takes care of aligning PTS between audio/video streams.

In other words, I don't believe presentationTimeOffset is required to be observed at the moment. In all cases so far we have enough information elsewhere to correctly calculate the segment numbers.

The changes fix broken streams while preserving existing ones that we're aware of, so LGTM!

PR #588, opened 2021-01-09
[Leia] License Renewal for Youtube movies

MisterD81:
At least purchased movies in Youtube addon stopped working. Always after 5 minutes we can see in the logs that there is no decryption key. Surprisingly I was only able to get this license behavior only on Windows and Linux ARM. On Linux x64 the license renewal was not required.
See the corresponding bug in YT addon: https://github.com/anxdpanic/plugin.video.youtube/issues/35 

I figured out with EME logger of Chrome that the CDM is sending periodically a session message to renew the license.
IAS is not doing that at the moment which results in the behavior that playback always stops after 5 minutes. I found in the code the commented-out parts for the license renewal and activated them. The additional adjustments were required to ensure that the thread is stopped correctly. Otherwise Kodi crashes when the playback is finished or is stopped.

We can argue about the sleep duration, but I could not observe a high CPU increase or high delay when playback stopped with the 100ms.

Corresponding Matrix PR: https://github.com/peak3d/inputstream.adaptive/pull/589

### DIFF
--- a/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
+++ b/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
@@ -64,18 +64,21 @@ void* GetCdmHost(int host_interface_version, void* user_data)
 
 }  // namespace
 
-std::atomic<bool> exit_thread_flag{false};
++std::atomic<bool> exit_thread_flag;
++std::atomic<bool> timer_thread_running;
 
 void timerfunc(std::shared_ptr<CdmAdapter> adp, uint64_t delay, void* context)
 {
+  timer_thread_running = true;
   uint64_t waited = 0;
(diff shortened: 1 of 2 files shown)


glennguy:
Hi @MisterD81 we're branching to Kodi N very shortly and now on 19.1 stable - Leia branch is unlikely to get any more releases at this stage unless there's a critical bug

MisterD81:
Hi @glennguy ,
understood. Too bad that the PR was already there before Matrix release. 
I close the request knowing that for example LibreElec RPi 2/3 users must build ISA on there own if they want to watch purchased YT movies with Kodi.

matthuisman:
LibreELEC / CoreELEC could always patch their own builds of IA if they wanted
OSMC for example adds their own patches specific for his OS.

PR #589, opened 2021-01-09
[Matrix] License Renewal for Youtube movies

MisterD81:
At least purchased movies in Youtube addon stopped working. Always after 5 minutes we can see in the logs that there is no decryption key. Surprisingly I was only able to get this license behavior only on Windows and Linux ARM. On Linux x64 the license renewal was not required.
See the corresponding bug in YT addon: https://github.com/anxdpanic/plugin.video.youtube/issues/35 

I figured out with EME logger of Chrome that the CDM is sending periodically a session message to renew the license.
IAS is not doing that at the moment which results in the behavior that playback always stops after 5 minutes. I found in the code the commented-out parts for the license renewal and activated them. The additional adjustments were required to ensure that the thread is stopped correctly. Otherwise Kodi crashes when the playback is finished or is stopped.

We can argue about the sleep duration, but I could not observe a high CPU increase or high delay when playback stopped with the 100ms.

Since I'm still on Leia this patch is only created based on the Leia PR (https://github.com/peak3d/inputstream.adaptive/pull/588)

### DIFF
--- a/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
+++ b/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
@@ -64,18 +64,21 @@ void* GetCdmHost(int host_interface_version, void* user_data)
 
 }  // namespace
 
-std::atomic<bool> exit_thread_flag{false};
+std::atomic<bool> exit_thread_flag;
+std::atomic<bool> timer_thread_running;
 
 void timerfunc(std::shared_ptr<CdmAdapter> adp, uint64_t delay, void* context)
 {
+  timer_thread_running  = true;
   uint64_t waited = 0;
   while (!exit_thread_flag && delay > waited) {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
     waited += 100;
   }
-  if (!exit_thread_flag){
+  if (!exit_thread_flag) {
     adp->TimerExpired(context);
   }
+  timer_thread_running = false;
 }
 
 cdm::AudioDecoderConfig_1 ToAudioDecoderConfig_1(
@@ -138,7 +141,9 @@ CdmAdapter::CdmAdapter(
 CdmAdapter::~CdmAdapter()
 {
   exit_thread_flag = true;
-  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  while (timer_thread_running) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
   if (cdm9_)
     cdm9_->Destroy(), cdm9_ = nullptr;
   else if (cdm10_)
@@ -155,6 +160,8 @@ CdmAdapter::~CdmAdapter()
 
 void CdmAdapter::Initialize()
 {
+  exit_thread_flag = false;
+  timer_thread_running = false;
   if (cdm9_ || cdm10_ || cdm11_)
   {
     if (cdm9_)
@@ -319,7 +326,9 @@ void CdmAdapter::CloseSession(uint32_t promise_id,
   uint32_t session_id_size)
 {
   exit_thread_flag = true;
-  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  while (timer_thread_running) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
   if (cdm9_)
     cdm9_->CloseSession(promise_id, session_id, session_id_size);
   else if (cdm10_)
--- a/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
+++ b/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
@@ -5,6 +5,7 @@
 #include "cdm_adapter.h"
 #include <chrono>
 #include <thread>
+#include <atomic>
 
 #define DCHECK(condition) assert(condition)
 
@@ -63,10 +64,18 @@ void* GetCdmHost(int host_interface_version, void* user_data)
 
 }  // namespace
 
+std::atomic<bool> exit_thread_flag{false};
+
 void timerfunc(std::shared_ptr<CdmAdapter> adp, uint64_t delay, void* context)
 {
-  std::this_thread::sleep_for(std::chrono::milliseconds(delay));
-  adp->TimerExpired(context);
+  uint64_t waited = 0;
+  while (!exit_thread_flag && delay > waited) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    waited += 100;
+  }
+  if (!exit_thread_flag){
+    adp->TimerExpired(context);
+  }
 }
 
 cdm::AudioDecoderConfig_1 ToAudioDecoderConfig_1(
@@ -128,6 +137,8 @@ CdmAdapter::CdmAdapter(
 
 CdmAdapter::~CdmAdapter()
 {
+  exit_thread_flag = true;
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
   if (cdm9_)
     cdm9_->Destroy(), cdm9_ = nullptr;
   else if (cdm10_)
@@ -307,6 +318,8 @@ void CdmAdapter::CloseSession(uint32_t promise_id,
   const char* session_id,
   uint32_t session_id_size)
 {
+  exit_thread_flag = true;
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
   if (cdm9_)
     cdm9_->CloseSession(promise_id, session_id, session_id_size);
   else if (cdm10_)
@@ -475,7 +488,9 @@ cdm::Buffer* CdmAdapter::Allocate(uint32_t capacity)
 
 void CdmAdapter::SetTimer(int64_t delay_ms, void* context)
 {
-  //LICENSERENEWAL std::thread(timerfunc, shared_from_this(), delay_ms, context).detach();
+  //LICENSERENEWAL
+  exit_thread_flag = false;
+  std::thread(timerfunc, shared_from_this(), delay_ms, context).detach();
 }
 
 cdm::Time CdmAdapter::GetCurrentWallTime()
--- a/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
+++ b/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
@@ -71,11 +71,13 @@ void timerfunc(std::shared_ptr<CdmAdapter> adp, uint64_t delay, void* context)
 {
   timer_thread_running  = true;
   uint64_t waited = 0;
-  while (!exit_thread_flag && delay > waited) {
+  while (!exit_thread_flag && delay > waited) 
+  {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
     waited += 100;
   }
-  if (!exit_thread_flag) {
+  if (!exit_thread_flag) 
+  {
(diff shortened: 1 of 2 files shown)


glennguy:
@MisterD81 Thanks for the PR, sorry just getting to it now. Looks fine to me and doesn't crash but @phunkyfish could you have a once over whenever suits? I'm no expert in c++, just want to check the thread sleeping has no issue with you.

MisterD81 do you know does the CDM call `timerfunc` at the ~5 minute mark or is it at the start of the stream (called with 300000ms)?

MisterD81:
It is actually even a little bit more complicated. Initially you get keys that are valid for 5 minutes.
CDM requests a timer via "SetTimer" at startup, which is set in my case to 1-2 minutes (can't remember in detail the time). Once the timer is over in timerfunc we call CDM and inform that the timer has expired. CDM then triggers the license renewal via a specific CDM message (for which we have CheckLicenseRenewal). This will then set the timer again.
So you'll always have a timer running and should inform CDM when the CDM is expired. I have no clue if this timer is also used by CDM for other purposes then extending the license.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -63,10 +64,18 @@ void* GetCdmHost(int host_interface_version, void* user_data)
 
 }  // namespace
 
+std::atomic<bool> exit_thread_flag{false};
```
phunkyfish: For simpler types regular assignment is easier to understand.

`std::atomic<bool> exit_thread_flag = false;`

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -63,10 +64,18 @@ void* GetCdmHost(int host_interface_version, void* user_data)
...
 void timerfunc(std::shared_ptr<CdmAdapter> adp, uint64_t delay, void* context)
 {
-  std::this_thread::sleep_for(std::chrono::milliseconds(delay));
-  adp->TimerExpired(context);
+  uint64_t waited = 0;
+  while (!exit_thread_flag && delay > waited) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    waited += 100;
+  }
+  if (!exit_thread_flag){
```
phunkyfish: Missing space `flag) {`

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -128,6 +137,8 @@ CdmAdapter::CdmAdapter(
 
 CdmAdapter::~CdmAdapter()
 {
+  exit_thread_flag = true;
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
phunkyfish: This value appears arbitrary. What exactly is this waiting for?

This value could very well work on one system but not on another. It would be better to continue once a condition is satisfied.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -307,6 +318,8 @@ void CdmAdapter::CloseSession(uint32_t promise_id,
   const char* session_id,
   uint32_t session_id_size)
 {
+  exit_thread_flag = true;
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
phunkyfish: This value appears arbitrary. What exactly is this waiting for?

This value could very well work on one system but not on another. It would be better to continue once a condition is satisfied.

phunkyfish (changes requested):

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
MisterD81: Moved initialization to "Initialize()" as suggestion does not compile.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
MisterD81: I added another flag to check if timer thread is running or not. The sleep is/was for ensuring that the timer is not running anymore as otherwise cdm would get destroyed to early.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
MisterD81: Same as above.

matthuisman:
I rented a movie last night so can test. Only for next 48 hours tho

I'll pull in your PR over latest Matrix branch and test

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
phunkyfish: Ah, of course. It's atomic bool. My bad.

matthuisman:
@MisterD81 are you able to quickly run through what this pr is doing? Where does the sleep time come from? How does it know when its needed? Is this going to make all other content keep refreshing its license? Do we know why the code was commented out?

I see its mostly uncommenting code so hard to tell what the missing pieces are

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -307,6 +325,10 @@ void CdmAdapter::CloseSession(uint32_t promise_id,
   const char* session_id,
   uint32_t session_id_size)
 {
+  exit_thread_flag = true;
+  while (timer_thread_running) {
```
phunkyfish: Bracket on next line.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -128,6 +140,10 @@ CdmAdapter::CdmAdapter(
 
 CdmAdapter::~CdmAdapter()
 {
+  exit_thread_flag = true;
+  while (timer_thread_running) {
```
phunkyfish: Bracket on next line.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -63,10 +64,21 @@ void* GetCdmHost(int host_interface_version, void* user_data)
...
 {
-  std::this_thread::sleep_for(std::chrono::milliseconds(delay));
-  adp->TimerExpired(context);
+  timer_thread_running  = true;
+  uint64_t waited = 0;
+  while (!exit_thread_flag && delay > waited) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    waited += 100;
+  }
+  if (!exit_thread_flag) {
```
phunkyfish: Bracket on next line.

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -63,10 +64,21 @@ void* GetCdmHost(int host_interface_version, void* user_data)
...
+std::atomic<bool> exit_thread_flag;
+std::atomic<bool> timer_thread_running;
+
 void timerfunc(std::shared_ptr<CdmAdapter> adp, uint64_t delay, void* context)
 {
-  std::this_thread::sleep_for(std::chrono::milliseconds(delay));
-  adp->TimerExpired(context);
+  timer_thread_running  = true;
+  uint64_t waited = 0;
+  while (!exit_thread_flag && delay > waited) {
```
phunkyfish: Bracket on next line.

phunkyfish (changes requested):

matthuisman:
ok tested. without this PR - YT movie stopped around 4:50 for me
and a bunch of DecodeVideo: kNoKey for key 43C36BB3877B5304913F9ED079E553EA  messages

with this PR, I see a license request every minute and it played pass 5minutes without issue
Checked and the browser does same behavior (license request every minute)

@MisterD81
as far as I can tell, all the refresh stuff is done in the cdm itself?
Is it the response from the license server that maybe contains the refresh info?
Seems all the delay between refresh etc comes internally from the cdm.
then it must just call the same function that calls back into IA to get the license?

@anxdpanic
Not sure if this version of Youtube add-on is still being maintained, but noted 1x issue with a purchased movie.
its higher quality streams would pause after the non-encrypted first few segments.
Im pretty sure they need L1 device for playback. in the browser, the quality is limited to 480p.
If Youtube gives you that info (that it needs a certain widevine level), maybe you could exclude them if not Android device. 
And if Android, then you'd have a new "L1 Device" setting. so that you can remove them on android if not L1.

matthuisman (approved):
looks good and works good!

just need those formatting changes fixed that phunky pointed out.
And maybe squash down into a single commit :)

phunkyfish:
> @MisterD81 are you able to quickly run through what this pr is doing? Where does the sleep time come from? How does it know when its needed? Is this going to make all other content keep refreshing its license? Do we know why the code was commented out?
> 
> I see its mostly uncommenting code so hard to tell what the missing pieces are

I'm also curious on these questions.

MisterD81:
I adjusted the formatting.
I found that out also only by trial and error. This what I know from debugging also with the EME logger of Chrome:
- YT gives you on some platforms only keys that are valid for around 5 minutes.
- The CDM requests a timer and wants to be notified (CdmAdapter.SetTimer()). The wake up time itself comes from the CDM. So Chrome or ISA should notify the CDM that the timer is over. The time is for YT around 1 minute and is controlled by the CDM itstelf. Nothing actively. This code was already in ISA, but commented out.
- Once the timer expires the CDM should be notified via TimerExpired(). Then SendClientMessage() will be triggered by the CDM which contains a renewal request in the end to the license server. Unfortunately the check for a new Key Challenge was commented out in wvdecrypter.cpp and the method CheckLicenseRenewal() which would check for a new challenge from the CDM would never get triggered. Once the new challenge is there it follows more or less the same procedure as initially. Of course CDM sets again a timer at the end and the process begins from the beginning.
- The problematic part is that once you set the timer and would wait until the timer expires Kodi crashes when you stop playback. Perhaps someone could do it also with locks/guards and so on, but I decided to sleep only for a small period in the timerfunc to recognize that playback should stop. This way the timer gets stopped properly and Kodi stops crashing. Probably that was the reason why it was commented out. I found the commit when the code was introduced but it was introduced commented out.
- Additionally keep in mind that most licenses that ISA is receiving do not expire that fast. They are not setting the timer. I checked it with Amazon for example and there the timer is not set. Also YT is doing that only on some platforms/CDM combinations. Unfortunately it seems that all RPi2/3 and Windows (x64) users are affected for the moment. So this is completely CDM driven based on what the license server is returning.
Hope that helps.

phunkyfish:
Thanks for explaining. Did you know there is actually a Timer class available as part or the add-on API?

https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-dev-kit/include/kodi/tools/Timer.h

So you can just implement ITimerCallback in your class and have any number of timers.

phunkyfish (approved):
Thanks for contributing!

matthuisman:
@MisterD81

thanks for the explanation.
this is very nice - implementing a feature of the CDM itself.
This could also help any other add-ons that require refreshing the license (its just not Youtube specific)

I assume it was originally commented out due to the crashing issue that you fixed.
Good work!

@glennguy 
this should be good to go in once you review :)

matthuisman:
I am finding this fixes Binge (and possible other addons) that now requires license renewal even on non-Android platforms.
@glennguy 
maybe a candidate for Kodi 18 backport?

Issue #583, opened 2021-01-03
More download attempt are not managed? with CURLOpen error case "FillBuffer - Failed"/"Open failed with code 0"


CastagnaIT:
This is a borderline case which happens a few times when servers are unstable due to too many connections from users
can happen that Curl fail the HTTP request

Line 5188
<pre>
2021-01-02 21:26:39.950 T:2616    DEBUG <general>: AddOnLog: inputstream.adaptive: Download https://ipv4-c040-mil001-ix.1.oca.nflxvideo.net/?o=AQNI2phcDnT2ANJfO7Th_NEcq69y5_xxBt_ygw1itRgy2xx83hbbCqrcyYzUU3-fBwlCu7UTQo13hfiST3SPxpMEg5ey5yjPba4VfYH6IoHeSCdM6DG8rOJMwmIKCY-5xbvKha_7-Q2OsDk44hIzRgwpw2yqBOUuP_oVU38&v=5&e=1609661985&t=2iZXRUua7pmCGRX9ijz2S8AITQ0 finished, avg speed: 3637019.45byte/s, current speed: 3663700.00byte/s
2021-01-02 21:26:40.078 T:9952    ERROR <general>: CCurlFile::FillBuffer - Failed: Couldn't resolve host name(6)
2021-01-02 21:26:40.078 T:9952    ERROR <general>: CCurlFile::Open failed with code 0 for https://ipv4-c040-mil001-ix.1.oca.nflxvideo.net/?o=AQNI2phcDnT2ANJfObLl-9Mdrqtx5_xxBsjygw1itRgy2xx83hbbCqrcyYzUU3-fBwlCu7UTQo13hfiST3SPxpMEg5ey5yjPba4VfYH6IoHeSCdM6DG8rOJMwmIKCY-5xbvKhav09Q2OsDk44hIzRgwpw2yqBOUuP_oVU38&v=5&e=1609661985&t=6e0vpK0idQCu7bUw0I9AQteZfAY:
</pre>

LOG: [kodi.log](https://github.com/peak3d/inputstream.adaptive/files/5761366/kodi.log)

What is happened:
- The Curl has failed, in my case the audio stream<br/>
CCurlFile::FillBuffer - Failed: Couldn't resolve host name(6)
CCurlFile::Open failed with code 0 for...
- ISA does not make any attempt to retry downloading therefore the download of the audio stream is blocked
[...]

PR #594, opened 2021-01-24
Try retry the download when CURLOpen fail on stream download

CastagnaIT:
I have noticed that on days with heavy server traffic (on december/jennuary for christmas days/new year's days) the stream download sometime failed causing unexpected behaviour to Kodi like my issue #583

the problem is that when CURLOpen fails the download will be interrupted
so let's make at least 2 attempts before breaking the playback

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -89,7 +89,7 @@ void AdaptiveStream::worker()
     if (type_ == AdaptiveTree::SUBTITLE)
       retryCount = 1;
 
-    while (!ret && !stopped_ && retryCount-- && tree_.has_timeshift_buffer_)
+    while (!ret && !stopped_ && retryCount--)
     {
       std::this_thread::sleep_for(std::chrono::seconds(1));
       Log(LOGLEVEL_DEBUG, "AdaptiveStream: trying to reload segment ...");


src/main.cpp
```diff
@@ -411,6 +412,15 @@ bool KodiAdaptiveStream::download(const char* url,
...
   }
+  else
+  {
+    if (retry_count > 0)
+    {
+      kodi::Log(ADDON_LOG_ERROR, "Download %s failed, new attempt...", url);
+      retry_count -= 1;
+      goto RETRY;
+    }
+  }
```
MPParsley:
```suggestion
  else if (retry_count > 0)
  {
    kodi::Log(ADDON_LOG_ERROR, "Download %s failed, new attempt...", url);
    retry_count--;
    goto RETRY;
  }
```

CastagnaIT: thanks

MPParsley (approved):
Could this be the cause for:

```
2021-02-02 20:56:20.607 T:1762808032 WARNING: AddOnLog: InputStream Adaptive: Initializing stream with unknown KID!
2021-02-02 20:56:21.843 T:1762808032   ERROR: CCurlFile::FillBuffer - Failed: HTTP returned error 403
2021-02-02 20:56:21.843 T:1762808032   ERROR: CCurlFile::Open failed with code 403 for https://wv-keyos.licensekeyserver.com/:

```

CastagnaIT:
@MPParsley for error 403 sometimes means problem with DNS used or you are under VPN and the server refuse your connection, perhaps you can also try check possible http headers to add

MPParsley:
> @MPParsley for error 403 sometimes means problem with DNS used or you are under VPN and the server refuse your connection, perhaps you can also try check possible http headers to add

Thanks for the tip @CastagnaIT!

In my case it's probably a real 403 since there's no KID, I opened another issue for it on https://github.com/peak3d/inputstream.adaptive/issues/599

glennguy:
@CastagnaIT I don't think this is the right solution to the problem. AdaptiveStream already has a segment retrying capability, just that it isn't enabled for VOD streams:
https://github.com/xbmc/inputstream.adaptive/blob/99d72cdc57a783c1c88bc23ba231a2dd2bc16b60/src/common/AdaptiveStream.cpp#L92-L97
My guess would be that 404s are quite common on the very edge of a live stream but it's assumed (obviously not the case though) that VOD won't ever 404 or fail otherwise.
If you remove the live stream requirement then all should be good :)

CastagnaIT:
oh thanks i will do

arnova:
This looks like a ugly hack. This implicates that if we need to retry because a server is already overloaded, we hammer it even more. And still: If we want something like this, I think it should go into CCurlfile where one has greater control over when this should kick in.

CastagnaIT:
@glennguy done, but i not understand why one build test is failed...
@arnova this PR is only to improve existing behaviour
if you wish develop/review all the code that manage the requests "retry" in a better way is better do it into another PR

arnova:
> @glennguy done, but i not understand why one build test is failed...
> @arnova this PR is only to improve existing behaviour
> if you wish develop/review all the code that manage the requests "retry" in a better way is better do it into another PR

Agreed.

glennguy (approved):
LGTM

matthuisman (commented):
unsigned int retryCount(10);

Is 10 retries too many?
Do we need delay between?
Could we do 10 for live, 3 for vod for example?

glennguy:
I think the 1 second delay is appropriate, no delay would be just hammering the server.

Remember the issues with Kayo - 1 second delay and retry was usually enough to get it resolved. The only time I see multiple retries in a log is when there's another issue causing wrong segment number or filename calculation, then we get the full 10 retries due to 404 and stream eventually stops.

10 was probably a number plucked out of the air and could be reduced.

Is this outside the scope of the PR though (let VOD streams have the functionality to retry)?

matthuisman:
Is there already a delay here that missed?

Or maybe its better to retry based on total time taken. So a slow server may only retry 3 times. But fast server could retry 10

Anyway. Something to keep in mind.
Looks good to me

matthuisman (approved):

PR #601, opened 2021-02-09
Change decrypter search path for Android

glennguy:
@ksooo 

As requested this one-liner will direct inputstream.adaptive to search in the user addons folder instead of packaged folder for libssd_wv.so

Now installations and updates from the official repo will function correctly, myself and @matthuisman have verified using a parallel installation with a different add-on id.

One potential issue that I haven't investigated and don't have immediate time for is how things work when upgrading from an Android installation of Kodi packaged with IA to a package without - will the packaged IA libraries and other add-on files remain and cause conflict? Or are they removed? If this is an unfamiliar scenario I'll find time in the next few days to setup and test it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,11 +86,7 @@ else()
   set(ADP_ADDITIONAL_BINARY $<TARGET_FILE:ssd_wv>)
 endif()
 
-if(CORE_SYSTEM_NAME STREQUAL android)
-  set(DECRYPTERPATH "special://xbmcbinaddons")
-else()
-  set(DECRYPTERPATH "special://home/cdm")
-endif()
+set(DECRYPTERPATH "special://home/cdm")
 
 list(APPEND DEPLIBS bento4)
 list(APPEND DEPLIBS mpegts)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(version: "Matrix", deploy: ['osx-x86_64', 'windows-i686', 'windows-x86_64', 'ubuntu-ppa'])
+buildPlugin(version: "Matrix")
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# inputstream.adaptive (2.6.6)
+# inputstream.adaptive (2.6.12)
 
 This is an adaptive file addon for kodi's new InputStream Interface.
 
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="inputstream.adaptive"
-  version="2.6.11"
+  version="2.6.12"
   name="InputStream Adaptive"
   provider-name="peak3d">
   <requires>@ADDON_DEPENDS@</requires>
@@ -19,6 +19,9 @@
     <description lang="es_ES">Cliente InputStream para flujo de datos adaptativos</description>
     <platform>@PLATFORM@</platform>
     <news>
+v2.6.12 (2021-04-09)
+- Remove Android specific decrypter search paths
+
 v2.6.11 (2021-04-08)
 - Fix ampersand in changelog causing issues from v2.6.9 and v2.6.10
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2137,14 +2137,8 @@ void Session::GetSupportedDecrypterURN(std::string& key_system)
   kodihost->SetLibraryPath(kodi::vfs::TranslateSpecialProtocol(specialpath).c_str());
 
   std::vector<std::string> searchPaths(2);
-#ifdef ANDROID
-  searchPaths[0] = getenv("KODI_ANDROID_LIBS")
-                       ? getenv("KODI_ANDROID_LIBS")
-                       : kodi::vfs::TranslateSpecialProtocol("special://xbmcbinaddons/");
-#else
   searchPaths[0] =
       kodi::vfs::TranslateSpecialProtocol("special://xbmcbinaddons/inputstream.adaptive/");
-#endif
   searchPaths[1] = kodi::GetAddonInfo("path");
 
   std::vector<kodi::vfs::CDirEntry> items;


matthuisman:
@glennguy 
looking at the code, isn't searchPaths[1] = kodi::GetAddonInfo("path");  after the #ifdef  actually adding the add-ons path to to be searched anyway? (no hard-coded addon id as well)

https://github.com/xbmc/inputstream.adaptive/blob/ef78a5715dd436f955ed9635f773343bc515ec79/src/main.cpp#L2147

If so, this change shouldn't actually be needed?
It'll just search xbmcbinaddons and then the add-on path.
So as long as the .so is removed from xbmcbinaddons, it'll pick up the version in IA's folder.

If anything, maybe the order needs to be reversed.
Addon path first, then xbmcbinaddons 2nd as fallback?

Or, if there are no versions that use xbmcbinaddons anymore, then can get rid of the whole ifdef  else
And just  std::vector<std::string> searchPaths(1);  searchPaths[0] = kodi::GetAddonInfo("path");

(I suspect we needed this change for IA Testings to stop it picking up IA's files)

glennguy:
@matthuisman you're correct, I'll push the changes soon if I get a chance, otherwise tonight.
Best to take out searching any place where it doesn't need to, whether we expect to see a file there or not.
Not sure myself if xbmcbinaddons is pointing to somewhere different on non-android platforms, I assume it's there for a reason...

matthuisman:
i suspect a path still needs to be added?
Otherwise will the searchPaths[1] = line fail if there is no searchPaths[0] ?
Maybe the other xbmcbinaddons is for UWP?

So, maybe
```
searchPaths[0] = kodi::vfs::TranslateSpecialProtocol("special://xbmcbinaddons/inputstream.adaptive/");
searchPaths[1] = kodi::GetAddonInfo("path");
```
for all platforms?

or
```
searchPaths[0] = kodi::GetAddonInfo("path");
#ifndef ANDROID
  searchPaths[1] = kodi::vfs::TranslateSpecialProtocol("special://xbmcbinaddons/inputstream.adaptive/");
#endif
```

glennguy:
was just thinking 
```
  std::vector<std::string> searchPaths(2);
  searchPaths[0] =
      kodi::vfs::TranslateSpecialProtocol("special://xbmcbinaddons/inputstream.adaptive/");
  searchPaths[1] = kodi::GetAddonInfo("path");
```

glennguy:
Thanks for the suggestion, much cleaner and no redundancy now.
The potential issue I mention above is more about the inputstream.adaptive shared library itself as the fear is the addon db might still be looking at the baked-in version... but I'm just speculating without any proper knowledge of how it actually works.

matthuisman:
just trying to think through the rollout of moving IA to repo

If this PR is merged, a new version of IA with this PR included in kodi android will actually fail as it wont find the .so.
If IA is put into the repo without this PR, then IA will keep picking up the old built-in .so.

So, once this is merged - the next version of IA must be in repo.

I think this could happen after 19 is released... as long as a built-in binary add-on can be replaced / updated by a repo version?
@howie-f may know if that should be fine?

The built-in .so will remain, but IA will no longer search for it there.

howie-f:
without having tested it (and not using android at all) i think the packaged version will strand in `xbmcbinaddons`; the update installed from the repo in the `user addon folder` which is mapped to home.
the built-in is not replaced until a new kodi version is deployed.
so the two .so-versions might co-exist. does kodi need to checkout both versions and pick the higher in this case? honestly i don't know

matthuisman:
jenkins build this please

matthuisman:
@glennguy
i just clicked about this PR for android path... I've been thinking its fixing the path to find the inpustreamadaptive.so...
I'm so dumb - obviously the code running is in inpustreamadaptive.so.  HAHA. It's just for the ssd_wv
Now I see your concern about what inpustreamadaptive.so kodi will choose...
The in-built one or the the one in the add-on itself 

Doing some tests now to find out

My idea to test is:
1) Install Kodi 18.9 on Android
2) Install build from this PR: https://github.com/xbmc/inputstream.adaptive/pull/634 which includes your PR
3) Get a stream that doesn't work on IA 2.4.6
4) Confirm it doesn't work
5) Install the new zip and then check it works

glennguy:
Yeah that's right and with search path I think we might need to change to special://home
You can see which ssdwv.so is being loaded in debug log, it will 'cache' it first time around.

matthuisman:
im more worried about what actual inputstreamadaptive.so Kodi loads after I install the zip?
Surely trying a stream that the latest version has a fix for will tell me which is loaded (and the log)
and yes, also test correct ssdwv is loaded to

matthuisman:
**Test Results**

**Kodi 18.9**
Success!
you can install a new version and it'll correctly use the new libinputstream.adaptive.so from the add-on directory and still use the old libssd_wv.so which hasn't had any changes - so is fine.
Note: If IA has been used before update - you need to restart Kodi otherwise it keeps using the old .so

**Kodi 19**
Success!
No reboot needed.

Here is the log
[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6282370/kodi.log.txt)

1) Boot Kodi 19 
2) Try to play add-on that needs latest fix - error (https://codeload.github.com/luivit/plugin.video.rivedila7/zip/refs/heads/master)
3) Install latest zip with above PR (https://jenkins.kodi.tv/job/xbmc/job/inputstream.adaptive/job/PR-636/1/artifact/cmake/addons/build/zips/inputstream.adaptive+android-aarch64/inputstream.adaptive-2.6.11.zip)
4) Try to play add-on - works 

From the log:
```
Loading: /data/app/org.xbmc.kodi-kHmUamwIh-bXF2AKco5lSQ==/lib/arm64/libinputstream.adaptive.so
Found decrypter: /data/app/org.xbmc.kodi-kHmUamwIh-bXF2AKco5lSQ==/lib/arm64/libssd_wv.so
```
changes to
```
caching /storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/inputstream.adaptive/libinputstream.adaptive.so to /data/user/0/org.xbmc.kodi/cache/lib/libinputstream.adaptive.so
Loading: /data/user/0/org.xbmc.kodi/cache/lib/libinputstream.adaptive.so

Caching '/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/inputstream.adaptive/libssd_wv.so' to '/data/user/0/org.xbmc.kodi/cache/lib/libssd_wv.so'
Found decrypter: /storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/inputstream.adaptive/libssd_wv.so
```
So using the correct libinputstream.adaptive.so and libssd_wv.so without any reboot needed

@glennguy  interesting that libssd_wv.so automatically gets cached but we don't use the cached version...
Unless we are just not logging the cached path and it actually uses /data/user/0/org.xbmc.kodi/cache/lib/libssd_wv.so


**Plan of attack**
Kodi 18 doesn't need this PR
As it's not going into the Repo for Android - users can manually install it on Android if required

Kodi 19
1) Merge this PR with a version bump and android builds added to https://github.com/xbmc/inputstream.adaptive/blob/Matrix/Jenkinsfile
(pretty sure it just needs changing to **buildPlugin(version: "Matrix")** - same as PVR simple)
2) That should get it into the Matrix Repo and Android Kodi 19 should pull the update and work fine
3) Make sure Inputstream Adaptive is removed from being included in Android Kodi builds

matthuisman:
jenkins build this

glennguy:
Thanks heaps for testing

> Unless we are just not logging the cached path and it actually uses /data/user/0/org.xbmc.kodi/cache/lib/libssd_wv.so

I think that's the case, DllHelper logging where it's moved to and IA is just logging the path the decrypter was found in.

matthuisman:
did up a test repo:

1) Start with stock Kodi 19
2) Enable debug logging
3) Install https://f.mjh.nz/test/plugin.video.rivedila7-Matrix.zip (test add-on)
4) Try to play "LA7 Live" - won't work
5) Install repo: https://f.mjh.nz/test/repository.inputstream.adaptive.0.0.1.zip
6) Update Inputstream Adaptive to 2.6.12
 (@howie-f  - I noticed it wouldn't auto-update or show in updates even with Update official add-ons from Any repositories.
  It does show via Select version though - maybe because binary add-on. I'm just using stock 19 from playstore)
7) Try to play "LA7 Live" - will work
8) Confirm in log that IA correctly loads 
` /data/user/0/org.xbmc.kodi/cache/lib/libinputstream.adaptive.so`&
`/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/inputstream.adaptive/libssd_wv.so`

Here is my log:
[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6289131/kodi.log.txt)

**Conclusion - this PR looks good once @howie-f can confirm what may be happening with the update**
hopefully if it's in the official repo it will detect update

howie-f:
> what may be happening with the update ?

update should be detected when the add-on hits the official repo. as system add-ons and manually installed are bound to official repo update checks.

glennguy:
As @matthuisman pointed out above:
> Kodi 19
>
> * Merge this PR with a version bump and android builds added to https://github.com/xbmc/inputstream.adaptive/blob/Matrix/Jenkinsfile
    (pretty sure it just needs changing to buildPlugin(version: "Matrix") - same as PVR simple)
> * That should get it into the Matrix Repo and Android Kodi 19 should pull the update and work fine
> * Make sure Inputstream Adaptive is removed from being included in Android Kodi builds

Point 1 is solved in this PR
Point 2 has been tested as far as "work fine" goes, and @howie-f has pointed out that the new add-on system for 19 will allow auto updating the built-in/system add-on if from the official repo
Point 3 should be then taken care of after merge and confirmation that we don't have any unforeseen issues (🤞 ). I've looked but haven't found anything on github/xbmc that controls which binary add-ons are included in Android build/packaging... Apart from this we are ready to go.

phunkyfish:
So this is ready to merge and release?

glennguy:
yes ready to go

phunkyfish:
Ok I’ll merge and release 2.6.12 shortly.

phunkyfish:
Do you want to add to the changelog that this make adaptive update from repo for android?

EDIT, I can update that after this is merged.

phunkyfish (approved):
LGTM

PR #607, opened 2021-02-24
Fix Base Domain (fixes uri=/path/)

### LINKED ISSUE #591: [HLS] URI="/path/" fails (regression)
Works fine in 2.4.6.
Fails in 2.6.6

`#EXT-X-KEY:METHOD=AES-128,URI="/nlds_vod/csmrugby/vod/2021/01/16/8563/whole/v1/stream/2_8563_edinburgh_glasgow_2016_b_whole_1_4500.mp4/000000.m3u8.key?hdntl=exp=1611043963`

Correct URL (2.4.6. works):
`https://neuclnlds1620oc.akamaized.net/nlds_vod/csmrugby/vod/2021/01/16/8563/whole/v1/stream/2_8563_edinburgh_glasgow_2016_b_whole_1_4500.mp4/000000.m3u8.key?hdntl=exp=161104396`

2.6.6 errors with
` AddOnLog: inputstream.adaptive: Cannot download /nlds_vod/csmrugby/vod/2021/01/16/8563/whole/v1/stream/2_8563_edinburgh_glasgow_2016_b_whole_1_4500.mp4/000000.m3u8.key?hdntl=exp=1611043963`

So appears it isn't appending the base url for urls starting with /
Relative paths (../../../) work fine when I tested

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/591

without this commit, base_domain_ is always empty 
causing any URI=/path/ to not have the base domain prepended.

Current code: **paramPos = base_url_.find("://", 0, 8);**
http://www.cplusplus.com/reference/string/string/find/
size_t find (const char* s, size_t pos, size_t n) const;

str = sequence to find
pos = start position to start looking
n = length of sequence of characters to match

the length of our sequence is only 3 - "://"
So, it's failing when trying to get 8 characters of that sequence.
The if statement below never gets executed and we end up with no base_domain_ 

without pos, start of string is used (what we want)
without n, all the sequence is used (what we want)

ie:
paramPos = base_url_.find("://");
is the same as
paramPos = base_url_.find("://", 0, 3);

This could also fix other issues where urls are relying on base domain

UPDATE:
It also fixes DASH manifests using media or init urls starting with "/" as well: https://github.com/xbmc/inputstream.adaptive/issues/608

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -331,7 +331,7 @@ namespace adaptive
     }
     base_url_.resize(paramPos + 1);
 
-    paramPos = base_url_.find("://", 0, 8);
+    paramPos = base_url_.find("://");
     if (paramPos != std::string::npos)
     {
       base_domain_ = base_url_;


smf007:
That makes sense now.  When I was looking at this previously, I was interpreting the n to be the length to check in str, not the length of s.  So I would agree with this fix.

smf007:
@phunkyfish is there anyone in the team taking the lead with IA.  Now that Matrix is out the door, is there someone that can help get these pulls reviewed and merged?  Thanks.

phunkyfish:
I'm happy to do that but I would require a number of users to be able to test the changes. I don't use adaptive myself so apart from a simple play test I would require more testing to be sure the change does not introduce bugs.

I can arrange for whomever these users are to join the kodi slack to make things move a little quicker. What do you think?

BTW, I don't specifically mean this PR as it's an obvious fix. But many others are much more complex.

Maybe as a start someone could group them. 1) almost no risk, 2) needs some testing 3) needs a whole lotta testing.

That way we get some of them in more quickly.

matthuisman:
@glennguy is working on a testing framework for IA which should make it a lot faster to unit test against a lot of different files.
This should remove quite a bit of the risk. 

@phunkyfish 
do you think this PR could be at least merged?
It's pretty obvious and fixes HLS and DASH files using urls starting with "/".

glennguy:
Yes it's coming, I hope this weekend sometime :)

kszaq:
FWIW I've been running IA with this patch (and admittedly all @matthuisman's PRs) and have had no issues, this fix is really required.

phunkyfish:
> FWIW I've been running IA with this patch (and admittedly all @matthuisman's PRs) and have had no issues, this fix is really required.

The plan was to take this and the other issues that improve the test coverage together once that is in. Since we will have some failing tests initially we can mark each PR that causes a failing test to pass. Once all tests pass we can make this part of CI so tests must all pass to get a green from Jenkins.

dagwieers:
How much I like having automated tests, nothing beats getting stuff out and tested by our stakeholders. Most of which are hit by issues that have fixes sitting in the pipeline for many many months.

So if possible, please can we get automated builds going (including popular distributions like LibreELEC, CoreELEC, Debian, Ubuntu and Fedora) so we can ask stakeholders (affected users and add-on devs) to test such pre-releases and give the ability to bisect (i.e. a build per commit could be ideal).

We need to merge fixes and find regressions at a faster pace IMO.

matthuisman:
Jenkins does builds for every PR:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-607/1/artifacts

Not sure Team Kodi will want to start doing builds for systems they don't technically support (Linux ARM etc)?
In saying that, they do have a PPA for Linux that gets build - so maybe it's possible to get artifacts for Linux x86_64 at least?

glennguy:
I think part of the issues we see are because we've been relying on only run-time tests which don't give an immediate answer or any sort of consistent coverage.

If you did want to set up automated builds/repo, the hard work has already been done here: https://github.com/johnny5-is-alive/inputstream.adaptive.testing , just needs a bit of love moving from TravisCI to GitHub Actions.

phunkyfish:
Co-ordinating automated builds across those platforms/distributions would be tricky. Any idea on how to achieve that @dagwieers ?

dagwieers:
It is not different from the Windows or macOS builds. In fact, I would expect this to be easier on Linux in general.

Both @peno64 and @CastagnaIT have documented it here: https://github.com/xbmc/inputstream.adaptive/wiki/How-to-build

We did this exercise a few times but this was pre-release, not continuous: https://github.com/xbmc/inputstream.adaptive/wiki/Test-results

dagwieers:
> If you did want to set up automated builds/repo, the hard work has already been done here: https://github.com/johnny5-is-alive/inputstream.adaptive.testing , just needs a bit of love moving from TravisCI to GitHub Actions.

I think it would be worthwhile to invest in, and a common way to offer those artefacts. GitHub Kodi package support? 😉

CastagnaIT:
>  just needs a bit of love moving from TravisCI to GitHub Actions

@glennguy i have already done the transition to GitHub Actions
you can found here: https://github.com/CastagnaIT/inputstream.adaptive
or take my commit as ispiration

the only not tested is the MacOs artifact build
NOTE: for each build the zip artifacts are automatically stored in action run page (below) you can see the example on Action page

Issue #617, opened 2021-03-15
HLS master playlist redirect edge case


glennguy:
Problem scenario example:
A master m3u8 is requested from a url which 3xx redirects to another location. The resulting master m3u8 has an absolute uri within that is of the same domain as the master m3u8 before redirection. e.g:
* IA is given url `https://bit.ly/1234`
* This redirects to `https://baz.qux/master.m3u8`
* master.m3u8 contains a variant stream which is located at `https://bit.ly/abcd`
* IA incorrectly builds the download url for this stream as `https://baz.qux/abcd` when it should still be `https://bit.ly/abcd`
This is not an issue if the variant stream has a uri with a different domain name eg `https://link.to/abcd`

Thanks to @matthuisman for discovering this

matthuisman:
the issue is here:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/common/AdaptiveTree.cpp#L413

url = https://bit.ly/abcd
effective_url_ = https://baz.qux/
base_url_ =  https://bit.ly/

so, that check is returning true.
we do have effective url and bit.ly is in https://bit.ly/abcd
so it then replaces the length of base_url_ with that many characters from the start of effective_url_ 
which i think could also cause issues. if base_url is https://bit.ly/ and effective is https://baz.qux/ - then I think we would get https://bit.lyx/

A possible fix would be to make sure effective_url_ is cleared somewhere after doing all the master stuff
but then it may be needed later on?

smf007:
As it is an absolute URI, wouldn't that be a third condition, so a third conditional check should be used?

`url.compare(0, 7, "http://") != 0 && url.compare(0, 8, "https://") != 0`

matthuisman:
That was my first attempt. Ignore replacing with full ulrs. But that breaks some streams as IA earlier on generates full URLs that actually need replacing. There is no way to tell between an actual full URL from file or one that has been appended earlier on.

smf007:
Ah... that's the problem then.  Why was a full URL generated previously?  Perhaps that is either an error or structurally/process wise it should be occurring later, or alternatively this code is occurring too late.

matthuisman:
I think URLs without leading / are appended before this. 

That was my 2nd approach. Making sure all URLs are basically unaltered until this function. That way it could ignore full URLs. It worked and removed quite a bit of duplication, but broke something I think. Possibly just an oversight as I didn't spend too much time on it.

[...]

PR #619, opened 2021-03-25
[HLS] Clear the effective URL if an absolute URI is specified.

smf007:
This fixes the HLS master playlist redirect edge case identified as issue #617.

Basically, if an absolute URI is specified, then drop the effective URL.

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -410,12 +410,6 @@ namespace adaptive
     {
       if (url.front() == '/')
         return effective_domain_.empty() ? base_domain_ + url : effective_domain_ + url;
-      else if (!effective_url_.empty() && url.compare(0, base_url_.size(), base_url_) == 0)
-      {
-        std::string newUrl(url);
-        newUrl.replace(0, base_url_.size(), effective_url_);
-        return newUrl;
-      }
     }
     return url;
   }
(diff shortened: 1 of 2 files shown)


matthuisman:
what about if the next entry is a relative url?

Your also clearing it in other cases too (not just absolute) eg "/my_relative_path/sub.m3u8"

If anything you'd want to check "://" is present in line, and if yes - then clear effective.
That still wont' solve the issue above if the next url was relative

glennguy:
Yes I agree there's an issue with the case where we have something like:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-INDEPENDENT-SEGMENTS
#EXT-X-STREAM-INF:BANDWIDTH=582476,AVERAGE-BANDWIDTH=581432,RESOLUTION=480x270,FRAME-RATE=25.000,CODECS="avc1.42C015,mp4a.40.2"
https://bit.ly/abcd
#EXT-X-STREAM-INF:BANDWIDTH=1000000,AVERAGE-BANDWIDTH=1000000,RESOLUTION=720x576,FRAME-RATE=25.000,CODECS="avc1.42C015,mp4a.40.2"
/variant2.m3u8
```

Personally I've never seen a master m3u8 like this but never say never, I'm just a sample size of 1.

smf007:
Thanks for the input.  I did have some concerns after pushing that, which triggered some follow up ideas which hopefully should work.

smf007:
Just updated with my best idea, why use the base_URL for relative URIs when we know we are going to have to use the effective_URL in the end.  All the gtest cases pass with this and should resolve the case with mixed absolute and relative URIs.

matthuisman:
Still not right.
If there is no redirect, effective_url will be empty.
Also, your removing a chunk of code that is used by other calls all over the place. Eg. A redirected key URI.

Our tests are still limited. We can't assume passing all tests means the code is correct. 

You are starting to see my point in regards to not needing the original URL anymore

I've started a draft PR here to start refactoring it: https://github.com/xbmc/inputstream.adaptive/pull/620

Issue #226, opened 2019-02-04
media_renewal_url for HTTP status code 401


flubshi:
Hi peak3d,

I am currently working on creating a pvr plugin for the german tv stream provider waipu.tv
I noticed that a running stream is stopped three hours after the passed token was generated and a 401 HTTP status code is returned, because the token is expired. This raises two questions:

**First problem:**
Currently, the _inputstream.adaptive.media\_renewal\_url_ option is limited to 403 HTTP status code:
https://github.com/peak3d/inputstream.adaptive/blob/71be2d9b0a0c3a3c1c411cebff08caa368330fdd/src/main.cpp#L354
Is it possible either to add the 401 HTTP status code to this condition or to add an option to set the renew-status-code dynamically while setting the stream options? e.g. _inputstream.adaptive.media\_renewal\_http\_code_ 

**Second problem:**
How to implement this for a pvr plugin?

peak3d:
@flubshiI I only see the option that you implement an tcp server either in binary pvr addon or a separate python service addon (bundled with your pvr addon) which proxies the license request. The license renewal impl. in inputstream.adaptive was a workaround, IMO the proper solution would be the proxy.

For example netflix addon uses such an license request proxy.

glennguy:
@flubshi the workaround code mentioned has been removed, are we able to close this issue now?

flubshi:
I will close this and check later. Thanks!

PR #620, opened 2021-03-25
Refactor / Simplify effective (redirect) logic

### LINKED ISSUE #617: HLS master playlist redirect edge case

matthuisman:
This fixes: https://github.com/xbmc/inputstream.adaptive/issues/617

Changes

- removed manifest_parameter_ as not used
- dont track original url and effective (redircted) url. Simply update all the url variables to the effective url
- move base_url appending into BuildDownloadUrl. This removes a lot of duplicate code from HLSTree
- no more newUrl.replace(0, base_url_.size(), effective_url_);  replacing parts of urls with other urls "yuckiness"
- HLS: when playing a substream playlist directly, use manifest_url_ as this will be the effective url 
This stops doing the redirect again when IA requests the suburl
- HLS: have substream update effective urls so no longer needs to track its own base_url
- HLS: store the full urls for substreams when parsing the master playlist to allow switching substreams. 
Without this, due to the above - we no longer have access to the master url to calculate the substream url
- **Remove media_renewal_url & media_renewal_time** as no longer work with above due to relying on replacing domains (yuck)
Peak himself said this was a workaround https://github.com/xbmc/inputstream.adaptive/issues/226#issuecomment-494290363. Correct approach would be manifest_update_param="full" or a Python proxy if that won't work.

TEST BUILDS HERE:
**https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-620/13/artifacts**

Looking for anything that plays in current 19 that then doesn't play with this build
(Dont expect it to fix any currently broken streams)

Apart from media_renewal_url  - No actual behaviour should have changed

### DIFF
--- a/src/test/TestDASHTree.cpp
+++ b/src/test/TestDASHTree.cpp
@@ -91,13 +91,12 @@ TEST_F(DASHTreeTest, CalculateBaseDomain)
   EXPECT_EQ(tree->base_domain_, "https://foo.bar");
 }
 
-TEST_F(DASHTreeTest, CalculateEffectiveUrlFromRedirect)
+TEST_F(DASHTreeTest, CalculateBaseUrlFromRedirect)
 {
-  // like base_url_, effective_url_ should be path, not including filename
   testHelper::effectiveUrl = "https://foo.bar/mpd/stream.mpd";
   OpenTestFile("mpd/segtpl.mpd", "https://bit.ly/abcd", "");
-
-  EXPECT_EQ(tree->effective_url_, "https://foo.bar/mpd/");
+  EXPECT_EQ(tree->base_url_, "https://foo.bar/mpd/");
+  EXPECT_EQ(tree->manifest_url_, "https://foo.bar/mpd/stream.mpd");
 }
 
 TEST_F(DASHTreeTest, CalculateBaseURLFromBaseURLTag)
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -10,7 +10,7 @@
     name="adaptive"
     extension=""
     tags="true"
-    listitemprops="license_type|license_key|license_data|license_flags|manifest_type|server_certificate|stream_headers|manifest_update_parameter|original_audio_language|media_renewal_url|media_renewal_time|max_bandwidth|play_timeshift_buffer"
+    listitemprops="license_type|license_key|license_data|license_flags|manifest_type|server_certificate|stream_headers|manifest_update_parameter|original_audio_language|max_bandwidth|play_timeshift_buffer"
     library_@PLATFORM@="@LIBRARY_FILENAME@"/>
   <extension point="xbmc.addon.metadata">
     <summary lang="en_GB">InputStream client for adaptive streams</summary>
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -39,7 +39,6 @@ AdaptiveStream::AdaptiveStream(AdaptiveTree& tree, AdaptiveTree::StreamType type
     currentPTSOffset_(0),
     absolutePTSOffset_(0),
     lastUpdated_(std::chrono::system_clock::now()),
-    lastMediaRenewal_(std::chrono::system_clock::now()),
     m_fixateInitialization(false),
     m_segmentFileOffset(0),
     play_timeshift_buffer_(false)
@@ -115,21 +114,6 @@ int AdaptiveStream::SecondsSinceUpdate() const
   return static_cast<int>(
       std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - tPoint)
           .count());
-}
-
-uint32_t AdaptiveStream::SecondsSinceMediaRenewal() const
-{
-  const std::chrono::time_point<std::chrono::system_clock>& tPoint(
-      lastMediaRenewal_ > tree_.GetLastMediaRenewal() ? lastMediaRenewal_
-                                                      : tree_.GetLastMediaRenewal());
-  return static_cast<int>(
-      std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - tPoint)
-          .count());
-}
-
-void AdaptiveStream::UpdateSecondsSinceMediaRenewal()
-{
-  lastMediaRenewal_ = std::chrono::system_clock::now();
 }
 
 bool AdaptiveStream::write_data(const void* buffer, size_t buffer_size)
--- a/src/common/AdaptiveStream.h
+++ b/src/common/AdaptiveStream.h
@@ -89,10 +89,6 @@ namespace adaptive
     virtual bool parseIndexRange() { return false; };
     bool write_data(const void *buffer, size_t buffer_size);
     bool prepareDownload(const AdaptiveTree::Segment *seg);
-    const std::string& getMediaRenewalUrl() const { return tree_.media_renewal_url_; };
-    const uint32_t& getMediaRenewalTime() const { return tree_.media_renewal_time_; };
-    uint32_t SecondsSinceMediaRenewal() const;
-    void UpdateSecondsSinceMediaRenewal();
     adaptive::AdaptiveTree& GetTree() { return tree_; };
 
   private:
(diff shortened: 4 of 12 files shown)


matthuisman:
pretty happy with this now

Things still need testing
1) MPD with update manifest url

matthuisman:
so this all works great except for media_renewal_url.

media_renewal_url  relies on a "hard replace" of a domain with another in the preparedownload function.
this is "yucky" and unreliable. the urls should be correctly calculated in the 1st place - not relying on replacing one base_url with another.  

the code for live dash manifests (which is where the media_renewal_url is used) calculates the full segment urls based off the `<baseurl>` in the manifest. Not the base_url_ in preparedownload.  


media_renewal_url essentially says. "hey - replace all http://mydomain/mytoken1/  with http://mydomain/mytoken2/"
But it will keep using the existing mpd.

Surely if the segment paths changed, the mpd would be updated with the new paths on each fetch.
In that case, just using "full" manifest update parameter should work

In the odd case of the MPD path also changing eg. http://mydomain/mytoken2/manifest.mpd  - then I think it would be better that media_renewal_url returns a new url to the manifest and IA goes and parses that.

I personally think media_renewal_url  should be removed. Its definitely a hack / work around.

Also, I think media_renewal_url is broken in Matrix IA anyway.
I created a test add-on which worked with it for 18, but broke on 19.

@dut-iptv is one author I see using media_renewal_url in plugin.video.ziggo add-on
Maybe you can explain why manifest parameter = "full" doesn't work in your case?

dut-iptv:
Media renewal url / time is used in my Ziggo IPTV addon because this provider:
-Does not require MPD updates after the initial request to https://ziggoapi.nl/token1/manifest.mpd
-Uses segment URLs in the format https://ziggoapi.nl/token1/segments.mp4
-A new token needs to be requested every 60 seconds after which the url changes to https://ziggoapi.nl/token2/segments.mp4 and so on

My addons try to mimic the HTML5 DASH implementations of the providers websites as close as possible and this is the way the system functions on the Ziggo website (without MPD updates)

I already use a proxy which could probably be adapted to provide the existing functionality without relying on media_renewal_url or time.

matthuisman:
thank you.

After 60seconds - is there a https://ziggoapi.nl/token2/manifest.mpd ?
If you requested https://ziggoapi.nl/token1/manifest.mpd - would it point to the https://ziggoapi.nl/token2/segments.mp4 ?
If so, just updating the mpd after every segment would work right?

How does the website know where to find the new segments?

dut-iptv:
After 60seconds - is there a https://ziggoapi.nl/token2/manifest.mpd ?  

Yes

If you requested https://ziggoapi.nl/token1/manifest.mpd - would it point to the https://ziggoapi.nl/token2/segments.mp4 ? 

No it would not, it would point to https://ziggoapi.nl/token1/segments.mp4 (by way of relative paths, the token is never in the MPD itself). The MPD would also not be available after 60s at the token1 url.

Easiest solution in my addon would probably be to let the proxy take care of changing token1 to token2 to token3 etc. Media renewal will not be required in that case.

You have my vote to remove it as I understand and support your reasoning for removing it and besided me no one seems to use it. If possible please allow me a week or so to write the required changes to my addon before merging.

matthuisman:
ok, so it's really you just need to tell IA that there is a new MPD url to use?

The add-on would return the new mpd path and IA would switch to using that.
It would download that new mpd which would give it the new base-url with the token in it and away it goes.

I think it would be possible to re-write it in IA to request the add-on to get the new manifest url.
Then download and process that.
Could you the same block as the current update manifest code

Last question: how does the website know where to switch the mpd url too?
They don't have have a static url that redirects to the latest token or something?

dut-iptv:
You request a token from https://ziggotoken.nl/token which returns token1 in JSON
Next you request the manifest from https://ziggoapi.nl/token1/manifest.mpd

Segment Templates/Timelines look like this:

`<SegmentTemplate initialization="$RepresentationID$/Header.m4s" media="$RepresentationID$/$Time$.m4s" presentationTimeOffset="3504693338944"><SegmentTimeline><S t="3555849519424" d="180480" r="1"/><S t="3555849880384" d="178560"/></SegmentTimeline></SegmentTemplate>`

$RepresentationID$ equals video[1-6], audio[1-2] for the different representations

which means IA requests https://ziggoapi.nl/token1/Header.m4s and https://ziggoapi.nl/token1/video1/firstsegment.m4s + https://ziggoapi.nl/token1/audio1/firstsegment.m4s and following segments from there.

After media renewal time has passed IA signals my kodi addon to request a new token from https://ziggotoken.nl/token which returns token2 in JSON and passes the media renewal url https://ziggoapi.nl/token2 back to IA after which IA starts requesting following segments from https://ziggoapi.nl/token2/video1/nextsegment.m4s.

So there is only one MPD request and following segments requests need to switch from one baseurl (including the first token) to the next baseurl (including second, third etc. token)

matthuisman:
yes, I was meaning what does the browser do...

How does the website know where to switch the mpd url too?
They don't have have a static url that redirects to the latest token url?

dut-iptv:
The player in the browser runs some JS/HTML5 code which knows to request a new token every 60 seconds and then start requesting segments using the url with the new token. 

There is no static url that redirects to the token url

matthuisman:
so browser never requests the latest MPD either?
Just switches up the urls?

dut-iptv:
Yes, there is only one MPD request when you initially start the stream/playback

dut-iptv:
Although an interesting discussion in how some providers decide to implement DASH it is now moot as far as I am concerned.

I have modified my proxy to take care of the requests of new tokens and redirecting to the correct URLs.

My addons no longer require media renewal and as you said, no one else seems to use it and it is pretty hacky so I fully support removing it.

matthuisman:
oh ok! Good work :)

src/test/TestDASHTree.cpp
```diff
@@ -49,10 +49,11 @@ TEST_F(DASHTreeTest, CalculateBaseDomain)
 
 TEST_F(DASHTreeTest, CalculateEffectiveUrlFromRedirect)
   {
-    // like base_url_, effective_url_ should be path, not including filename
+    // like base_url_, base_url_ should be path, not including filename
```
glennguy: Can probably remove this comment now and rename test to CalculateBaseUrlFromRedirect?

matthuisman: done :)

matthuisman:
closing this for now as just found a new issue

we can't update manifest_url_ in SetEffectiveURL.
as we then lose track of the original manifest url which is needed for updateTree

ChomArStudio:
Hi,

please help me.
horizon.tv in your js used replace url
`
if (e.indexOf("vxttoken") >= 0)
            return e.replace(/(vxttoken=)(.+?)(?=\/)/, "$1" + t); 
`
Base_url is not used by horizon.tv in your mpd file.

You are removed the renewal options and scripst now stop working.

And now, how am I supposed to change the effective url in script
please helps

matthuisman:
You'll need to use a http proxy in the addon to sit between IA and their server

You can look how dut-iptv did it with his add-on

ChomArStudio:
wasn't it simpler to hack the renewal option than to build a "proxy server" for one script?

matthuisman:
Simpler for who?

ChomArStudio:
for me.... and other script authors ;-)
if it worked, what was it for removing

matthuisman:
because it no longer worked with latest changes.

Feel free to submit a PR to bring it back in, in a less hacky way :)

ChomArStudio:
OK, thanks

PR #625, opened 2021-03-30
Don't overwrite manifest headers with stream_headers & default stream_headers to manifest headers

### LINKED ISSUE #624: Use custom headers for all downloads
Our iptv provider starts deliberately blocking user agent Kodi.

Our addon uses inputstream.adaptive with widevine DRM, manifest type MPD.
Noticed on Kodi 19, inputstream.adaptive version 2.6.8, but same behavior is reported on Leia.

Path used in list item is with custom headers, like:
`https://..../Manifest.mpd|Origin=https%3A//....&Referer=https%3A//....&User-Agent=Mozilla..etc`

Manifest is succesfully parsed, license update successful, but then:
```
CurlFile::ParseAndCorrectUrl() adding custom header option 'connection: keep-alive'
CurlFile::Open(0x2d383da20d0) https://..../init.m4v
AddOnLog: inputstream.adaptive: Download https://..../init.m4v failed with error: 403
AddOnLog: inputstream.adaptive: Unable to select stream!
```
Reason is - CurlFile is not using custom headers from original path.
I have used Chrome UA Spoofer to test, that our provider explicitly search for "kodi" in User Agent and then block Manifest.mpd, init.m4v and chunks with error 403..

Please, use passed custom headers for all downloads.

matthuisman:
Fixes https://github.com/xbmc/inputstream.adaptive/issues/624

**1**
Currently if you provide stream_headers and also | headers on manifest url, the stream_headers will overwrite those manifest headers.

This is because it removes them from the manifest url before they get processed further down.
Lets stop that behaviour.

If you provide | headers on the manifest url, you expect them to be used regardless of stream_headers

**2**
It makes sense to set the stream_headers to the manifest headers if they aren't provided.
They are part of the manifest (children), so many assume this would be the behaviour (inherit parents headers)
(It would be nice to also set the default headers for license url, but they are processed much later on in the ssd decrypter class.)

**3**
This PR keeps the current behaviour of using stream_headers for manifest headers if they aren't provided
So is backwards compatible with anyone using just stream_headers.
(Not sure if this is wanted correct behaviour though)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3359,7 +3359,6 @@ bool CInputStreamAdaptive::Open(const kodi::addon::InputstreamProperty& props)
                 prop.second.c_str());
       parseheader(manh, prop.second);
       medh = manh;
-      mpd_url = mpd_url.substr(0, mpd_url.find("|"));
     }
     else if (prop.first == "inputstream.adaptive.original_audio_language")
     {
@@ -3400,6 +3399,9 @@ bool CInputStreamAdaptive::Open(const kodi::addon::InputstreamProperty& props)
     parseheader(manh, mpd_url.substr(posHeader + 1));
     mpd_url = mpd_url.substr(0, posHeader);
   }
+
+  if (medh.empty())
+    medh = manh;
 
   kodihost->SetProfilePath(props.GetProfileFolder());
 


matthuisman:
@glennguy 
Is is possible to write tests for this?
The code is in main.cpp
when I tried to include that in tests cmake - a heap of "unresolved external symbol"

I thought we would have a new file called "TestCommon.cpp"
In setup it would create a new inputstream addon instance.
then tests could be written calling functions on that

glennguy:
Not easily.
A lot of the unresolved external symbols would be sorted by adding the rest of the source and header files to the executable (we currently only have what's needed for DASH/HLSTree). But this is not the main issue.

Creating an instance of the addon / main.cpp would require an instance of Kodi to host it I believe. I'm not sure that it would be trivial to make a mock Kodi instance to do this. @AlwinEsch might be able to give a quick answer to whether this would be feasable and if I'm right in my thinking.

Another option is doing similar to the changes made for aes_decrypter where an abstract class used as an interface is created so we can derive new classes from the ones in main.cpp and replace the calls to the host. Still not quick to do but maybe this is a better way to go and more flexible than above.

Lastly, perhaps all of this is an indication that a major refactor is in order. It's grown and grown over time and at 4000 loc perhaps a lot of it should be in separate modules.

glennguy (approved):

glennguy:
Looks good, brings functionality in line with what should have been the original intention

PR #626, opened 2021-03-31
[Dash] Correctly set timeshift_buffer (live)

### LINKED ISSUE #22: Channels not playing "Live"

matthuisman:
Fixes: https://github.com/d21spike/plugin.video.sling/issues/22 (also needs PR: https://github.com/xbmc/inputstream.adaptive/pull/564)

With the below MPD, currently IA will set the stream to VOD when it is actually LIVE

```
<MPD xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:cenc="urn:mpeg:cenc:2013" xmlns:dvb="urn:dvb:dash:dash-extensions:2014-1" xmlns:move="http://www.movenetworks.com/dash/v1" xmlns:ms="urn:microsoft" xmlns:XMLSchema-instance="http://www.w3.org/2001/XMLSchema-instance" XMLSchema-instance:schemaLocation="urn:mpeg:DASH:schema:MPD:2011 DASH-MPD.xsd" profiles="urn:mpeg:dash:profile:isoff-live:2011" maxSegmentDuration="PT2.048000S" minBufferTime="PT2.048000S" publishTime="2020-12-07T20:30:00Z" type="dynamic" minimumUpdatePeriod="PT2.048000S" availabilityStartTime="2020-12-07T20:30:15Z" mediaPresentationDuration="PT9000.000000S" suggestedPresentationDelay="PT4.096000S">
```

This is because "mediaPresentationDuration" comes after "type".
mediaPresentationDuration sets bStatic to true

So, currently it goes
1) "type"="dynamic" which makes bStatic = false
2) "mediaPresentationDuration" which sets bStatic = true
3) dash->has_timeshift_buffer_ is set to NOT bStatic = false
4) IA think it's VOD

The order or attributes should be irrelevant.

With this PR, VOD is default and it's ever only setting LIVE to true when either dynamic or timeShiftBufferDepth found
It doesn't unset LIVE.

### DIFF
--- a/src/test/TestDASHTree.cpp
+++ b/src/test/TestDASHTree.cpp
@@ -13,14 +13,14 @@ protected:
   }
 
   void TearDown() override
-  { 
+  {
     effectiveUrl.clear();
     delete tree;
     tree = nullptr;
   }
 
   void OpenTestFile(std::string testfilename, std::string url, std::string manifestHeaders)
-  { 
+  {
     SetFileName(testFile, testfilename);
     if (!tree->open(url, manifestHeaders))
     {
@@ -68,7 +68,7 @@ TEST_F(DASHTreeTest, CalculateSegTplWithNoSlashs)
 
   adaptive::AdaptiveTree::SegmentTemplate segtpl =
       tree->periods_[0]->adaptationSets_[0]->representations_[0]->segtpl_;
-  
+
   EXPECT_EQ(segtpl.initialization, "https://foo.bar/guid.ism/dash/media-video=66000.dash");
   EXPECT_EQ(segtpl.media, "https://foo.bar/guid.ism/dash/media-video=66000-$Number$.m4s");
 }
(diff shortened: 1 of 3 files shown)


Fireblade89:
I've been waiting for this for a long time. I can't wait to test it !

glennguy:
Looks good. `mediaPresentationDuration` doesn't imply VOD at all, it can be in either VOD or live. The guidelines are slightly ambiguous on this (mentions MPD 'may' be present in the context of static playlists, however doesn't explicitly mention them under dynamic playlist heading) which may have lead to the current interpretation

matthuisman:
That's what I read as well. timeShiftBufferDepth is only present with dynamic (live) so is a good identifier as well. 

We could also check profiles for is-off-live. But profiles can list multiple profiles. And I couldn't see a c++ function to search within a const chr. We would need to convert to std:string then find on that. Maybe something we can do in the future if we find live manifests not using type and not using timeshiftbuffer

matthuisman:
Fixes sling:
https://github.com/d21spike/plugin.video.sling/issues/22#issuecomment-812855081

src/test/TestDASHTree.cpp
```diff
@@ -128,4 +134,4 @@ TEST_F(DASHTreeTest, CalculateCorrectSegmentNumbersFromSegmentTimeline)
   EXPECT_EQ(segments.size(), 13);
   EXPECT_EQ(segments[0]->range_end_, 487050);
   EXPECT_EQ(segments[12]->range_end_, 487062);
-}
+}
```
phunkyfish: New line required

phunkyfish (changes requested):

matthuisman:
@phunkyfish  conflict fixed. New line added.added a 2nd test. squashed into a single commit

phunkyfish (approved):
LGTM

PR #627, opened 2021-03-31
[Dash] Remove old testTime (used for old internal testing)

matthuisman:
this was originally used in the old testing framework to set the clock time in the manifest itself.
The new framework overrides GetNowTime function so this is no longer required


phunkyfish:
If we push this again it should not pass the test suite run on mac

matthuisman:
you want me to do a force push?
And did you mean it should pass?

UPDATE: Did that and it did pass all tests

I think ill write some tests for my other PRs
this one doesn't need a test as it's just removing some of the old testing framework

PR #630, opened 2021-04-04
[Dash] Allow both $Time and $Number in URL

### LINKED ISSUE #629: $Time$ parameter in mpd manifest is not processed
Hi again,

our operator, which hates Kodi, hits with new round..

Mpd manifest now contains segment template with $Time$ parameter:

`<SegmentTemplate initialization="$RepresentationID$/init.m4v" media="$RepresentationID$/seg-0000000000-$Number$.m4v?t=$Time$" timescale="1000"  startNumber="00000000">`

Which ends up in several attempts like this in log, before it gives up:

```
DEBUG <general>: AddOnLog: inputstream.adaptive: AdaptiveStream: trying to reload segment ...
DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'Origin: https://....'
DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'connection: keep-alive'
DEBUG <general>: CurlFile::Open(0x1f2d3e5a8c0) https://..../seg-00000000-00000000.m4v?t=$Time$
ERROR <general>: CCurlFile::FillBuffer - Failed: Stream error in the HTTP/2 framing layer(92)
ERROR <general>: CCurlFile::Open failed with code 0 for https://..../seg-00000000-00000000.m4v?t=$Time$:
- empty line -
```

Note: i have replaced real numbers with "00000000" for safety :)

### LINKED ISSUE #644: Kodi crashes on DRM protected content
I originally posted this on the discoveryplus addon issue tracker, but it appears to be related to DRM and widevine, so I'll try reporting it here.

See this issue for original discussion: https://github.com/Dis90/plugin.video.discoveryplus/issues/42

As can be seen there, it crashes on both kodi 18.9 and 19, on Linux, Windows and Android. And only on DRM protected content.

I am trying to watch The Big Bang Theory on Discovery+ using the above addon, and kodi crashes when decoding DRM content. Episode 4 is not protected, episode 5 is.

The files attached are these:

Kodi 18.9 on Linux:

Working log (episode 4): https://paste.kodi.tv/wapuzedaza.kodi
Crashing log (episode 5): https://paste.kodi.tv/urebasedag.kodi
Kodi's own crashlog/debuglog: https://paste.kodi.tv/gidezuqepa.kodi

Kodi 19 on Win 10 64 bit: https://paste.kodi.tv/hukopuyogo.kodi

manifest.mpd from episode that crashes Kodi but works in Chrome: https://paste.kodi.tv/ebejadigad

and Dis90's manifest from finnish discovery+, not TBBT though: https://paste.kodi.tv/udufabiqic 

As Dis90 points out, there is an error in the logs:
```
InputStream Adaptive: DecryptSampleData: Decrypt failed with error: 2 and key: C53A9877F32C40A3B663C35CDA0AC26A
```

Any ideas on how to solve this?

Like I also say in that issue, the viaplay addon (https://github.com/emilsvennesson/kodi-viaplay) does work, and those are also using inputstream.adaptive and widevine.

Please, point me in the right direction, and if you need me to provide any more logs or files of any kind, let me know!

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/629 (confirmed)
fixes: https://github.com/xbmc/inputstream.adaptive/issues/644 (confirmed)

Currently, we can only have $Time$ OR $Number$ templates in a segment url.
However, both are possible to have in the same url.

Also fix spelling (old function was ReplacePlacehoder)

**Also, the current function segfaults if a url has neither $Number or $Time$.**
This is due to never checking for std::string::npos after the $Time$ block (which runs if no $Number$)

[Backport Required]

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -262,19 +262,15 @@ bool AdaptiveStream::restart_stream()
   return true;
 }
 
-void AdaptiveStream::ReplacePlacehoder(std::string& url, uint64_t index, uint64_t timeStamp)
+void AdaptiveStream::ReplacePlaceholder(std::string& url, const std::string placeholder, uint64_t value)
 {
-  std::string::size_type lenReplace(7);
-  std::string::size_type np(url.find("$Number"));
-  uint64_t value(index); //StartNumber
+  std::string::size_type lenReplace(placeholder.length());
+  std::string::size_type np(url.find(placeholder));
   char rangebuf[128];
 
   if (np == std::string::npos)
-  {
-    lenReplace = 5;
-    np = url.find("$Time");
-    value = timeStamp; //Timestamp
-  }
+    return;
+
   np += lenReplace;
 
   std::string::size_type npe(url.find('$', np));
@@ -333,7 +329,8 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment* seg)
     else if (seg != &current_rep_->initialization_) //templated segment
     {
       download_url_ = current_rep_->segtpl_.media;
-      ReplacePlacehoder(download_url_, seg->range_end_, seg->range_begin_);
+      ReplacePlaceholder(download_url_, "$Number", seg->range_end_);
+      ReplacePlaceholder(download_url_, "$Time", seg->range_begin_);
     }
     else //templated initialization segment
       download_url_ = current_rep_->url_;
@@ -344,7 +341,8 @@ bool AdaptiveStream::prepareDownload(const AdaptiveTree::Segment* seg)
         seg != &current_rep_->initialization_)
     {
       download_url_ = current_rep_->segtpl_.media;
-      ReplacePlacehoder(download_url_, current_rep_->startNumber_, 0);
+      ReplacePlaceholder(download_url_, "$Number", current_rep_->startNumber_);
+      ReplacePlaceholder(download_url_, "$Time", 0);
     }
     else
       download_url_ = current_rep_->url_;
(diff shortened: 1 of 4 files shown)


matthuisman:
@glennguy 
I tried to write a test for this to check but wasn't sure how to call AdaptiveStream functions...

glennguy:
We need to bring AdaptiveStream into the test framework, AdaptiveTree is a member of it. I'll look at it tomorrow, looks to be fairly straight forward.

matthuisman:
i did do that, but still couldn't call the replaceplaceholders function as they are private.
Then tried created a new AdaptiveStream with the test tree.. which worked...
but then still couldn't call the private functions and all the public functions that would call those private functions do a bunch of other stuff.

I thought - probably just easier to move the replaceplaceholders into helpers.cpp

matthuisman:
@glennguy 
Tests added

did notice something a bit interesting.
needed **llu** when formatting the $Time$ due to it's length..
I tested with a smaller "t" and the format worked correctly.
But at a certain length - the formatted number becomes wrong.
It'll be a 64 bit thing I assume.
I've never seen a format on the $Time$ anyway - but something to keep in mind.

You'll see here:
https://github.com/xbmc/inputstream.adaptive/blob/dd862285f8889ae5e813cd6a94523ce49ef91876/src/common/AdaptiveStream.cpp#L280

if no % format it uses "%" PRIu64  which would become **%llu**
But for example if the mpd used **$Time%014d$** - you'd get the wrong result.
In my test, I used **$Time%014llu$** which works fine

glennguy:
Tests look great and code changes are great, much easier to follow!

The `d` translates to `int` which is why it wasn't working, so we should never see a `%14d` anyway.

glennguy (approved):

PR #635, opened 2021-04-08
fix debian/ubuntu packaging

wsnipex:
PPA builds fail without this

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,7 +1,7 @@
 Source: kodi-inputstream-adaptive
 Priority: extra
 Maintainer: peak3d <info@peak3d.de>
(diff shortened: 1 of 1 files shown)


phunkyfish (approved):
LGTM

phunkyfish:
Does this need to be released for the PPA builds to work or just merged?

wsnipex:
I can trigger a manual build if needed, but if the next release is in the next couple of days, we can just wait for that.

phunkyfish:
I'm not sure when the next release will be. I would just trigger a manual build so that there is a version there. I'll merge this now.

PR #637, opened 2021-04-09
Only enable testing if BUILD_TESTING is set

candrews:
BUILD_TESTING is the convention for enabling test building.

This condition allows building without gtest.

See https://cliutils.gitlab.io/modern-cmake/chapters/testing.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 project(inputstream.adaptive)
+option(BUILD_TESTING "Build the testing tree." ON)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 project(inputstream.adaptive)
+option(BUILD_TESTING "Build the testing tree." ON)
```
wsnipex, line 3: we use ENABLE_TESTING in kodi

candrews: Is that a problem?

Perhaps adding the condition `CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME` would be the right behavior?

wsnipex: it should be consistent. I assume that other addons will sooner or later get tests as well.
Once more addons have tests, this should be added to the addon helper scripts in kodi.

notspiff:
Kodi should be changed imo as BUILD_TESTING is widely used.

lør. 10. apr. 2021, 06:56 skrev Wolfgang Schupp ***@***.***>:

phunkyfish (approved):

PR #645, opened 2021-04-12
[Dash] Default update_parameter to full if using SegmentTimeline

matthuisman:
Live dash manifests using SegmentTimeline always need to refresh the dash manifest as there is no way a SegmentTimeline can list segments forever.. A refresh is always needed eventually.

This change sets the update_parameter_ to "full" if the manifest is adp_timelined_ (using a SegmentTimeline) and is live (has_timeshift_buffer_)

Lots of issues of no audio are caused when this property is missing - so lets set it for them if we know it's going to be needed.
This means users will no longer have to manually set "manifest_update_parameter" property.


matthuisman:
@glennguy 

updated to make it explicit only set it if has_timeshift_buffer_ is true as well.
Makes the tests easier as update_thread_ is private.
And explicit is always good :)

Also added tests

and renamed some of my old tests as I had them named segtimeline when they are actually using seg templates

Ready for review and merge :)

**due to renames above - it would be best to merge this before the other PR as it will be easier to rebase the other**

glennguy (approved):

glennguy:
Sounds and looks good!

PR #649, opened 2021-04-14
Add capability to use AdaptiveStream in DASH tests

glennguy:
Currently we can test most areas of parsing DASH/HLS manifests. There are however some functions that run in `AdaptiveStream` at 'stream time' as opposed to 'parse time', such as ReplacePlaceHolders (replaces eg. `$Number$` and `$Time$` variables that need to be calculated using current time) and we would like to be able to put these under test.

This PR brings the capability to create an `AdaptiveStream` under test in order to use its functions. The download function used for retrieving the segment files has been replaced allowing us to keep track of the eventual url that is requested to download, and a helper function in the test fixture is included to easily call x amount of read/segment requests and store all of the 'downloaded' urls in an array for later retrieval.

No tests included (Don't worry, @matthuisman will be itching to add some to his PRs 😄) but a sample test could be as such:
```
TEST_F(DASHTreeAdaptiveStreamTest, replacePlaceHolders)
{
  OpenTestFile("mpd/segtimeline_live_ast.mpd", "https://foo.bar/test.mpd", "");
  videoStream->prepare_stream(tree->current_period_->adaptationSets_[0], 0, 0, 0, 0, 0, 0, 0,
                              mediaHeaders);
  videoStream->start_stream(~0, 0, 0, true);

  ReadSegments(videoStream, 16, 16);

  EXPECT_EQ(downloadedUrls[0], "https://foo.bar/videosd-400x224/segment_487050.m4s");
  EXPECT_EQ(downloadedUrls.back(), "https://foo.bar/videosd-400x224/segment_487062.m4s");
  EXPECT_EQ(downloadedUrls.size(), 13);
}
```

This will call `read` for 16 bytes on the stream 16 times. Note that the stream only contains 13 segments so `ReadSegments` will exit early.

### DIFF
--- a/src/test/TestHelper.cpp
+++ b/src/test/TestHelper.cpp
@@ -2,6 +2,7 @@
 
 std::string testFile;
 std::string effectiveUrl;
+std::string lastDownloadUrl;
 
 
 void Log(const LogLevel loglevel, const char* format, ...){}
@@ -43,6 +44,25 @@ bool adaptive::AdaptiveTree::download(const char* url,
   fclose(f);
 
   SortTree();
+  return nbRead == 0;
+}
+
+bool TestAdaptiveStream::download(const char* url,
+                                  const std::map<std::string, std::string>& mediaHeaders)
+{
+  size_t nbRead = ~0UL;
+  std::stringstream ss("Sixteen bytes!!!");
+
+  char buf[16];
+  size_t nbReadOverall = 0;
+  while ((nbRead = ss.readsome(buf, 16)) > 0 && ~nbRead && write_data(buf, nbRead))
+    nbReadOverall += nbRead;
+
+  if (!nbReadOverall)
+  {
+    return false;
+  }      
+  lastDownloadUrl = url;
   return nbRead == 0;
 }
 
--- a/src/test/TestHelper.cpp
+++ b/src/test/TestHelper.cpp
@@ -1,9 +1,8 @@
 #include "TestHelper.h"
 
-std::string testFile;
-std::string effectiveUrl;
-std::string lastDownloadUrl;
-
+std::string testHelper::testFile;
+std::string testHelper::effectiveUrl;
+std::string testHelper::lastDownloadUrl;
 
 void Log(const LogLevel loglevel, const char* format, ...){}
 
@@ -26,13 +25,13 @@ bool adaptive::AdaptiveTree::download(const char* url,
                                       void* opaque,
                                       bool scanEffectiveURL)
 {
-  FILE* f = fopen(testFile.c_str(), "rb");
+  FILE* f = fopen(testHelper::testFile.c_str(), "rb");
   if (!f)
     return false;
 
-  if (scanEffectiveURL && !effectiveUrl.empty())
-    SetEffectiveURL(effectiveUrl);
-
+  if (scanEffectiveURL && !testHelper::effectiveUrl.empty())
+    SetEffectiveURL(testHelper::effectiveUrl);
+ 
   // read the file
   static const unsigned int CHUNKSIZE = 16384;
   char buf[CHUNKSIZE];
@@ -62,7 +61,7 @@ bool TestAdaptiveStream::download(const char* url,
   {
     return false;
   }      
-  lastDownloadUrl = url;
+  testHelper::lastDownloadUrl = url;
   return nbRead == 0;
 }
 
--- a/src/test/TestHelper.h
+++ b/src/test/TestHelper.h
@@ -4,13 +4,17 @@
 #include "../parser/DASHTree.h"
 #include "../parser/HLSTree.h"
 
-
-extern std::string testFile;
-extern std::string effectiveUrl;
-extern std::string lastDownloadUrl;
 std::string GetEnv(const std::string& var);
 void SetFileName(std::string& file, const std::string name);
 void Log(const LogLevel loglevel, const char* format, ...);
+
+class testHelper
+{
+public:
+  static std::string testFile;
+  static std::string effectiveUrl;
+  static std::string lastDownloadUrl;
+};
 
 class TestAdaptiveStream : public adaptive::AdaptiveStream
 {
--- a/src/test/TestHelper.h
+++ b/src/test/TestHelper.h
@@ -1,14 +1,27 @@
 #include "../Iaes_decrypter.h"
 #include "../log.h"
+#include "../common/AdaptiveStream.h"
 #include "../parser/DASHTree.h"
 #include "../parser/HLSTree.h"
 
 
 extern std::string testFile;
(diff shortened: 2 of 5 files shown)


src/test/TestHelper.cpp
```diff
@@ -46,6 +47,26 @@ bool adaptive::AdaptiveTree::download(const char* url,
   return nbRead == 0;
 }
 
+bool TestAdaptiveStream::download(const char* url,
+                                  const std::map<std::string, std::string>& mediaHeaders)
+{
+  size_t nbRead = ~0UL;
+  std::stringstream ss("Sixteen bytes!!!");
+
+  char* buf = (char*)malloc(16);
```
phunkyfish: Why use c memory management here? And if we have casts they should be c++ casts.

src/test/TestHelper.h
```diff
@@ -1,15 +1,28 @@
 #include "../Iaes_decrypter.h"
 #include "../log.h"
+#include "../common/AdaptiveStream.h"
 #include "../parser/DASHTree.h"
 #include "../parser/HLSTree.h"
 
 
 extern std::string testFile;
 extern std::string effectiveUrl;
+extern std::string lastDownloadUrl;
```
phunkyfish: Is this common practice to define extern variable? Are they used across classes?

phunkyfish (changes requested):

src/test/TestHelper.h
glennguy: I don't know so much about if it's common practice but for this use case these variables are used across TestHelper and TestDASHTree/TestHLSTree. Making them global allows us to avoid having to define what would effectively be the same functions in both TestDASHTree.cpp and TestHLSTree.cpp (`adaptiveTree::download` and `adaptiveStream::download`) and just keeping them in TestHelper

src/test/TestHelper.cpp
glennguy: I copy/pasted from the implementation in main.cpp, my bad. I'll replace with new/delete

src/test/TestHelper.h
phunkyfish: Hmmm, it still feels like there should be a C++ way of doing this.

glennguy: No problem, I will investigate :)

glennguy: I've done a lot of research and there's really only 2 ways to go about this from what I gather.

A lot are just recommending to use `extern` to share the variable between source files.

The other method which I've tried and tested is to define a small helper class with static members like so:
```
//TestHelper.h

class staticVars
{
public:
  static std::string testFile;
  static std::string effectiveUrl;
  static std::string lastDownloadUrl;
};
```
```
//TestHelper.cpp

std::string staticVars::testFile;
std::string staticVars::effectiveUrl;
std::string staticVars::lastDownloadUrl;
```

And of course in TestHLSTree.cpp and TestDASHTree.cpp we access by eg. `staticVars::testFile = "foo"`

Which way do you think?

phunkyfish: Definitely the C++ way using the class!

glennguy: Cool, changes made :)

src/test/TestHelper.cpp
```diff
@@ -46,6 +46,26 @@ bool adaptive::AdaptiveTree::download(const char* url,
   return nbRead == 0;
 }
 
+bool TestAdaptiveStream::download(const char* url,
+                                  const std::map<std::string, std::string>& mediaHeaders)
+{
+  size_t nbRead = ~0UL;
+  std::stringstream ss("Sixteen bytes!!!");
+
+  char* buf = new char[16];
```
phunkyfish: char buf[16];

src/test/TestHelper.cpp
```diff
@@ -46,6 +46,26 @@ bool adaptive::AdaptiveTree::download(const char* url,
   return nbRead == 0;
 }
 
+bool TestAdaptiveStream::download(const char* url,
```
phunkyfish, line 49: Why char* for URL instead of std::string? Although this is probably more of a refactoring task for another time.

src/test/TestHelper.cpp
```diff
@@ -46,6 +46,26 @@ bool adaptive::AdaptiveTree::download(const char* url,
...
+                                  const std::map<std::string, std::string>& mediaHeaders)
+{
+  size_t nbRead = ~0UL;
+  std::stringstream ss("Sixteen bytes!!!");
+
+  char* buf = new char[16];
+  size_t nbReadOverall = 0;
+  while ((nbRead = ss.readsome(buf, 16)) > 0 && ~nbRead && write_data(buf, nbRead))
+    nbReadOverall += nbRead;
+  delete buf;
```
phunkyfish: Not required if not using new.

src/test/TestHelper.cpp
```diff
@@ -25,13 +25,13 @@ bool adaptive::AdaptiveTree::download(const char* url,
                                       void* opaque,
                                       bool scanEffectiveURL)
 {
-  FILE* f = fopen(testFile.c_str(), "rb");
+  FILE* f = fopen(testHelper::testFile.c_str(), "rb");
```
phunkyfish, line 28: Maybe not for now either but Kodi provides a file class for addons that can be used to read files.

E.g. from iptvsimple using this: https://github.com/kodi-pvr/pvr.iptvsimple/blob/292ce5f4bf3a59917677de0473cb175ac3e0b84a/src/iptvsimple/utilities/FileUtils.cpp#L49-L61

phunkyfish (changes requested):
It's really only the allocation of `buf` that needs to be changed the rest are comments.

Also, you have a conflict.

src/test/TestHelper.cpp
glennguy: The big problem with wanting to do this is that we need an instance of Kodi to use it. It'd be wonderful to have a mock instance to use but I think building something like that is out of my depth.

src/test/TestHelper.cpp
glennguy: Just following the declaration here: https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/common/AdaptiveStream.h#L88
Yes it would be nice to get rid of all the C string stuff where possible at some point.

glennguy:
@phunkyfish conflicts sorted, hoping to meet your approval!

src/test/TestHelper.cpp
phunkyfish: Of course. That makes perfect sense!

phunkyfish (approved):
👍your button @glennguy

peak3d:
Glorious what you guys are rocking here!!!!!

PR #652, opened 2021-04-19
Stop intermittent crashing on AdaptiveStream tests

glennguy:
adaptiveTree was being destroyed before adaptiveStream, when adaptiveStream's destructor still needs access to the tree.


phunkyfish (approved):
👍

PR #654, opened 2021-04-21
[Dash] Support ttml subs without codecs

matthuisman:
Allow the same behaviour as wvtt by setting the codec to ttml for mimeType=application/ttml+xml
This also adds support for mimeType="text/vtt" on representations (not adaptionset) to work correctly
Also fixes a bug where using codecs="wvtt" without a mimeType, SUBTITLESTREAM would not be set.

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -710,7 +710,8 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
                 else if (strncmp(dash->current_adaptationset_->mimeType_.c_str(), "audio", 5) == 0)
                   dash->current_adaptationset_->type_ = DASHTree::AUDIO;
                 else if (strncmp(dash->current_adaptationset_->mimeType_.c_str(), "application",
-                                 11) == 0)
+                                 11) == 0 ||
+                         strncmp(dash->current_adaptationset_->mimeType_.c_str(), "text", 4) == 0)
                   dash->current_adaptationset_->type_ = DASHTree::SUBTITLE;
               }
               if (strstr(dash->current_adaptationset_->mimeType_.c_str(), "/webm"))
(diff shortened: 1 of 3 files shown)


matthuisman:
getting odd error
"undefined reference to `adaptive::AdaptiveTree::Representation::SUBTITLESTREAM'"

But i'm using
adaptive::DASHTree::Representation::SUBTITLESTREAM

works fine on windows..

src/parser/DASHTree.cpp
```diff
@@ -710,7 +710,8 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
                 else if (strncmp(dash->current_adaptationset_->mimeType_.c_str(), "audio", 5) == 0)
                   dash->current_adaptationset_->type_ = DASHTree::AUDIO;
                 else if (strncmp(dash->current_adaptationset_->mimeType_.c_str(), "application",
-                                 11) == 0)
+                                 11) == 0 ||
+                         strncmp(dash->current_adaptationset_->mimeType_.c_str(), "text", 4) == 0)
```
matthuisman, line 714: this is same as existing adaption set code below but for representation attributes
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/parser/DASHTree.cpp#L931

without this mimeType="text/vtt" on the **representation** won't work but mimeType="application/ttml+xml" would work

glennguy:
Try `#include "../common/AdaptiveTree.h"` in TestHelper.h

edit: nvm, should already be included in DASH/HLSTree.h

glennguy:
Due to `ATTRIBUTE_HIDDEN` we can't access this way.

Use `DASHTestTree::SUBTITLE` and `DASHTestTree::Representation::SUBTITLESTREAM`

edit: that doesn't fix the issue

glennguy:
I think the correct solution is to assign `adaptive::AdaptiveTree::Representation::SUBTITLESTREAM` to a variable outside of the test assertion (`EXPECT_EQ`), then we will avoid the visibility issue.

matthuisman:
@glennguy 
Success! thank you.
Added a comment in the code as well to remind us.

also added more tests because - why not!

ready for review and merge :)

src/parser/DASHTree.cpp
```diff
@@ -722,6 +723,28 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
             attr += 2;
           }
 
+          if (dash->current_adaptationset_->mimeType_ == "text/vtt" &&
+              dash->current_representation_->codecs_.empty())
+            dash->current_representation_->codecs_ = "wvtt";
+
+          if (dash->current_adaptationset_->mimeType_ == "application/ttml+xml" &&
```
glennguy: `else if` might be better here

src/parser/DASHTree.cpp
```diff
@@ -722,6 +723,28 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
...
+            dash->current_adaptationset_->mimeType_ = "text/vtt";
+          }
+
+          if (dash->current_adaptationset_->type_ != DASHTree::SUBTITLE &&
+              dash->current_representation_->codecs_ == "ttml")
+          {
+            dash->current_adaptationset_->type_ = DASHTree::SUBTITLE;
+            dash->current_adaptationset_->mimeType_ = "application/ttml+xml";
+          }
+
```
glennguy, line 747: These could both be nested under 1 condition for if (dash->current_adaptationset_->type_ != DASHTree::SUBTITLE)

matthuisman: how so?
they both need to check for text/vtt or ttml
so if I did just a check for ! SUBTITLE, I'd need two more checks inside that to check the codec.

glennguy (approved):

glennguy:
Looks great, going in :)

PR #658, opened 2021-04-23
[Dash] Support suggestedPresentationDelay & minBufferTime

### LINKED ISSUE #647: Add support for suggestedPresentationDelay
_Originally posted by @matthuisman in https://github.com/xbmc/inputstream.adaptive/issues/615#issuecomment-818234029_

> For a start - I see it's using SegmentTemplate now... so you'll need to set manifest_update_param to "full" so it refreshes the MPD
> 
> ```
> if live:
>     #listitem.setLabel(titolo_diretta)
>     listitem.setInfo('video', {'plot': titolo_diretta, 'title': titolo_diretta})
>     listitem.setProperty('inputstream.adaptive.manifest_update_parameter', 'full')
> ```
> 
> i actually have a PR to automatically do this I made last night: https://github.com/xbmc/inputstream.adaptive/pull/645
> 
> Even with the above - it seems to pause with a lot of "Begin WaitForSegment stream 1" so maybe another issue needs looked into
> 
> **Probably best to open a new issue after you have set the above update_parameter.**

matthuisman:
Fixes: https://github.com/xbmc/inputstream.adaptive/issues/647

**Fix Update interval**
updateInterval_ is time to wait between manifest updates in ms https://github.com/xbmc/inputstream.adaptive/blob/dc047c3ba7e1e43adfe8110b5623bbed22fcbb67/src/common/AdaptiveTree.cpp#L476  In a dash file, it's set with minimumUpdatePeriod which is in "PT32S" format
  AddDuration returns the time in (seconds * timescale).
  We want ms - so correct timescale is 1000 not 1500
  In HLS, EXT-X-TARGETDURATION is provided as seconds. Again, we want in ms so need to times by 1000 not 1500

_when looking into this, I found updateInterval_ is usually never used as currently IA will refresh after every segment here:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/common/AdaptiveStream.cpp#L387

RefreshSegments will call RefreshUpdateThread which calls updateVar_.notify_one(); which breaks the updateVar_.wait_for.
Something we should look into. RefreshSegments also calls RefreshLiveSegments which the updatethread will also call basically at the same time. So - downloading and processing the manifest twice?_
  
**Add suggestedPresentationDelay** 
Then moved the current 12sec delay from head into an adaptivetree variable called presentation_delay_ .
This gets set to 12 by default which results in current behaviour.
Now in the dash file, we can set it to the value provided by suggestedPresentationDelay which is provided in "PT32S" format

[Needs Tests]

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1051,7 +1051,7 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       else if (strcmp((const char*)*attr, "minimumUpdatePeriod") == 0)
       {
         uint64_t dur(0);
-        AddDuration((const char*)*(attr + 1), dur, 1500);
+        AddDuration((const char*)*(attr + 1), dur, 1000);
         dash->SetUpdateInterval(static_cast<uint32_t>(dur));
       }
       attr += 2;
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1048,10 +1048,15 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       }
       else if (strcmp((const char*)*attr, "availabilityStartTime") == 0)
         dash->available_time_ = getTime((const char*)*(attr + 1));
+      else if (strcmp((const char*)*attr, "suggestedPresentationDelay") == 0 ||
+               strcmp((const char*)*attr, "minBufferTime") == 0)
+      {
+        AddDuration((const char*)*(attr + 1), dash->buffer_time_, 1);
+      }
       else if (strcmp((const char*)*attr, "minimumUpdatePeriod") == 0)
       {
         uint64_t dur(0);
-        AddDuration((const char*)*(attr + 1), dur, 1000);
+        AddDuration((const char*)*(attr + 1), dur, 1500);
         dash->SetUpdateInterval(static_cast<uint32_t>(dur));
       }
       attr += 2;
@@ -1063,9 +1068,7 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
     AddDuration(mpt, dash->overallSeconds_, 1);
     dash->has_overall_seconds_ = dash->overallSeconds_ > 0;
 
-    uint64_t overallsecs(dash->overallSeconds_ ? dash->overallSeconds_ + 60 : 86400);
     dash->minPresentationOffset = ~0ULL;
-
     dash->currentNode_ |= MPDNODE_MPD;
   }
 }
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -203,10 +203,15 @@ bool AdaptiveStream::start_stream(const uint32_t seg_offset,
       if (!pos)
         pos = 1;
     }
-    //go at least 12 secs back
     uint64_t duration(current_rep_->get_segment(pos)->startPTS_ -
                       current_rep_->get_segment(pos - 1)->startPTS_);
-    pos -= static_cast<uint32_t>((12 * current_rep_->timescale_) / duration) + 1;
+
+    if (tree_.buffer_time_ > 0)
+      pos -= static_cast<uint32_t>((tree_.buffer_time_ * current_rep_->timescale_) / duration);
+    else
+      //go at least 12 secs back
+      pos -= static_cast<uint32_t>((12 * current_rep_->timescale_) / duration) + 1;
+
     current_rep_->current_segment_ = current_rep_->get_segment(pos < 0 ? 0 : pos);
   }
   else
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -48,6 +48,7 @@ namespace adaptive
     , stream_start_(0)
     , available_time_(0)
     , base_time_(0)
+    , buffer_time_(0)
     , minPresentationOffset(0)
     , has_timeshift_buffer_(false)
     , has_overall_seconds_(false)
@@ -171,7 +172,7 @@ namespace adaptive
   }
 
   void AdaptiveTree::OnDataArrived(unsigned int segNum, uint16_t psshSet, uint8_t iv[16], const uint8_t *src, uint8_t *dst, size_t dstOffset, size_t dataSize)
-  { 
+  {
     memcpy(dst + dstOffset, src, dataSize);
   }
 
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -425,7 +425,7 @@ public:
   uint32_t currentNode_;
   uint32_t segcount_;
   uint32_t initial_sequence_ = ~0UL;
-  uint64_t overallSeconds_, stream_start_, available_time_, base_time_;
+  uint64_t overallSeconds_, stream_start_, available_time_, base_time_, buffer_time_;
   uint64_t minPresentationOffset;
   bool has_timeshift_buffer_, has_overall_seconds_;
 
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -539,7 +539,7 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
         }
         else if (line.compare(0, 22, "#EXT-X-TARGETDURATION:") == 0)
         {
-          uint32_t newInterval = atoi(line.c_str() + 22) * 1000;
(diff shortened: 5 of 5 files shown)


glennguy:
> updateInterval_ is time to wait between manifest updates in ms 
> https://github.com/xbmc/inputstream.adaptive/blob/dc047c3ba7e1e43adfe8110b5623bbed22fcbb67/src/common/AdaptiveTree.cpp#L476

`updateInterval_` is used when the stream has been either paused or can't download more segments. The update thread will wait 1.5 * the length of the `minimumUpdatePeriod` without seeing a segment refresh before forcing a refresh. If the duration was set to 1000 instead of 1500 we have the situation you describe above where there'd be a race to update. Having this mechanism is useful as when you pause a live stream the tree will still stay current. Also the timeline will be updated and you'll see your 'now' marker slowly make its way towards the back of the buffer, it won't go 'out of date'.

I've done a little research and it seems that when mpd is using segmentTimeline, MUP should always be the segment length so we update after each segment which as above we do anyway.
Having MUP in the mpd basically tells the player it *must* update every x. It might though for example disappear after an mpd update when a live event has terminated and there's no more need for mpd updates (similar to EXT-X-ENDLIST in hls). Not that it's a big issue but we don't handle this currently from what I see, although it's possible another change in the mpd will trigger updates to stop anyway (mpd should also change from dynamic to static).


> Add suggestedPresentationDelay

The way this is done doesn't get at the root of the issue in #647. I did mention in the issue thread that it could be an option but on second thought I believe it will be completely wrong if we end up going down the path of removing segments from the live edge, we'll then be 64 seconds from live edge in this case.
SPD is supposed to come out of the timeshift buffer window, so instead of seeing 4 minutes on the Kodi timeline we should have 3:28. Then start 12 seconds back from there. Although the last ~8 segments at some time will be available for download, they should not be 'seen' by Kodi until they've moved out of the SPD window.

matthuisman:
This isn't removing any segments. Its just starting the stream further back. making the 12s adjustable. Same as rewinding basically. It would prepare us for downloading the segments ahead as well. 

Its quite easy to adjust the total time in kodi. You just decrease the overalseconds. So that can easily be adjusted down by the presentationdelay. But then you have the stream actually going past it as its current time also needs adjusting.

I thought fixes to timeline could go in another PR. As they actually already exist with our current 12s. Its just more noticeable when you increase that 12s

I'll remove the 1500 > 1000. Makes perfect sense. Its more of a.backup ifdoesnt get a segment before then. Does notify break the wait for and make it run its function? Or does it just reset the wait for time?

glennguy:
> This isn't removing any segments.

Yes that's right but I'm trying to say that the correct solution should be doing that. Then in turn reduce the overallseconds as part of that (if by that stage it's already been calculated). It would be a very bad idea however to reduce overallseconds without also removing the appropriate segments.

Where we need to be careful is the amount to remove, it shouldn't be the fixed 32s. It should be 32s back from 'now'. In the example in #647 the last segment doesn't bring us to the live edge, it's anywhere from 0-32s back from 'now', i.e we remove somewhere between 0 and 8 segments to make last segment finish >= 32s behind now.

kodi spd

And before implementing this we will need to get an accurate wall clock time (definitely not system clock, there's a lot out there that don't synchronise online). Maybe from mpd tag, 3rd party source, server headers, or falling back to system clock if those are unavailable.

Doing a blanket 'start stream at (now - SPD)' in IA is a bit 'hacky' imo. Users will see that they're not near the live edge, ffwd to live and it will stall. Then they'll come here and open a new issue. In the absence of a fix right now add-on authors can use your solution of `listitem.setProperty('ResumeTime', '206') \n listitem.setProperty('TotalTime', '240')` to get the same effect.

matthuisman:
We would still want to track the segments ahead though so if a cache is implemented, we can start caching them ready? 

Would this not be a start though? If a stream needs 30sec behind, should we not do that for now? Saves the addon dev opening issues here. And us telling them to use resumetime. Which will become wrong if streams buffer length changes. Sort of same situation as manifest update param. We know when that's needed so implement it for them.

We already doing 12s as it is, so why not at least use a more correct / compatible value instead if available?


If we didn't have the timeline in kodi and it just played like ffmpeg, would we want this in? 

I still dont see it as a hack. We are just choosing the start segment here..

Even if we remove segments to fix kodi timeline, we will still be using this start segment.

I think its working with what we currently have. The 12s has the same ffwd issue. We actually want 12s of segments removed not just start 12s back.

What we are doing here is setting an initial segment offset. What we eventually need to do is make it a persistent offset. And then remove segments after it. Or at least tell kodi there isnt anymore after.

I just think this is safer than trying to adjust timings for now. 

Just think of it as setting a more sensible start position :)

src/parser/DASHTree.cpp
```diff
@@ -1048,6 +1048,11 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       }
       else if (strcmp((const char*)*attr, "availabilityStartTime") == 0)
         dash->available_time_ = getTime((const char*)*(attr + 1));
+      else if (strcmp((const char*)*attr, "suggestedPresentationDelay") == 0 ||
+               strcmp((const char*)*attr, "minBufferTime") == 0)
```
glennguy, line 1052: We probably should ignore the `minBufferTime` attribute. It's to signal "buffer at least this amount of data before beginning playback" and is not to do with selecting *when* to start. Kodi already has its 8 second internal buffer which is linked to when playback begins.

glennguy:
One thing that we don't know for sure is that this 1 provider is using SPD as is intended. Everything that I've read suggests that SPD is to keep the player from an area of the timeline where segments may 404/not yet exist. There's been no mention anywhere that I've found that talks about mpds that are only updated server-side after multiple new segments are created, and using SPD to signal it. 
Are we looking at this problem from the correct perspective? If this is a case of poor provider implementation we should cater to the DASH-IF side of the fence, not LA7.

> We would still want to track the segments ahead though so if a cache is implemented, we can start caching them ready?

That is a good point, but SPD is trying to tell the player that this area of segments probably don't yet exist. Trying to cache them will still 404 on the server which goes against the intent of SPD.

> If we didn't have the timeline in kodi and it just played like ffmpeg, would we want this in?

Yes definitely as user would not be able to seek into the live edge which may not update with new segments for up to 32 seconds. Also the LA7 website player has no timeline bar to seek, just play/pause function. 

>We already doing 12s as it is, so why not at least use a more correct / compatible value instead if available?
> Even if we remove segments to fix kodi timeline, we will still be using this start segment.
>I think its working with what we currently have. The 12s has the same ffwd issue. We actually want 12s of segments removed not just start 12s back.

I believe 12 seconds has been chosen because of Kodi's 8 second buffer plus a little extra so we don't end up with `WAITFORSEGMENT` conditions. It's not trying to prevent 404s like SPD, just making sure we're not trying to run on the very last segment, giving Kodi 50% buffer to work with and getting stuck waiting for an mpd refresh to have new segments. If we were in this situation of course after enough stalling to buffer we'd end up at a more sensible distance from the edge. Starting 12s back avoids the issue. Similar but different to the issue here.

`start_stream` happens after all parsing is complete. If segments are removed and duration corrected then we'll be further behind again. Your solution wont select the segment 32s back from `now`, it will select the segment which is 32 back from the adjusted window, when 12s back would be perfectly fine.

> Would this not be a start though?
> Saves the addon dev opening issues here.

Because of the last point it wouldn't form part of the eventual solution, which makes it more of a temporary workaround. @Testato may see it as fine but how is it communicated to users, who will undoubtedly try to ffwd at some point and then raise an issue either there or here.


With regards to adjusting overallSeconds:
We really need to investigate if there are any unwanted effects of having the sum of segment lengths being greater than the overallSeconds. 
From my memory when working on HLS multi period stuff Kodi doesn't like it when the now time and overall duration don't make sense. So even if we restrict the overallSeconds and through that the duration that Kodi sees, if DemuxRead() is called and Kodi's buffer is filled from segments that reside outside of our window - what happens then? What happens when we seek within the timeshift buffer, does the intended segment get selected? In other words - Is Kodi set up for and going to allow this, and will IA accept the dissonance or freak out?

matthuisman:
Yeh. Good points!

I had a change of heart overnight, if we going to do it -may as well do it correctly first time and not be lazy :)

Testato:
But why via Browser the .mpd is refreshed every 4sec and not on Kodi ?
How tecnically it is possibile ?
Chrome and Kodi do not do the same thinks about the .mpd ?

glennguy:
It is refreshed every 4 seconds on Kodi too.
The actual content inside the mpd is only changed/updated twice per minute, on browser and Kodi.

Issue #638, opened 2021-04-10
Black Screen on Widevivene L1 Device (Samsung - Android10)


Testato:
Audio is ok

tested on Kodi19 by Samsung S8 and Samsung s9+ smartphone on Android10 64bit
using last version of LA7 plugin on IA 2.6.13
https://github.com/luivit/plugin.video.rivedila7/archive/refs/heads/master.zip
(the two Live channel si widevine encrypted, it's not Geoblocked, and is free, not need login)

The Samsung is Widevine L1, I attach the info from DRF Info

On cheeper smartphone L3 device the video and audio stream work regularly

I do not know how check the logfile on android, my phone is not rooted

![image](https://user-images.githubusercontent.com/6893841/114268338-db02bf80-9a00-11eb-9e35-6799e7d8d36f.png)

glennguy:
Hi @Testato 

Looking at the mpd there isn't any L1 content in it (720p max resolution)
If you set inputstream.adaptive settings to manual stream selection, can you get a picture on any of the 4 video streams?

@matthuisman has a lot of experience with L1/blank screens with his add-ons, he may know more.

Testato:
Thanks for your tips, manually choosing the video stream only the best one, 1280x720, is blank.
The other three stream, at lower resolution, work regularly.
Do you think is not L1 certification related ?

Testato:
On my LA7 plugin there is also Ondemand programs, some episodes are widevine protected and some not.
I tested that ondemand stream at 1280x720 not Widevine encrypted work regularly. Only the protected 1280x720 is blank and only on L1 certified device.

On an Android L3 phone, or on Ubuntu PC, or on Windows Pc also the 1280x720 protected stream work regularly.

matthuisman:
I'll try on my Shield (L1) when I get time.

you could try force IA to use the L1 secure decoder...
By adding this code
```
listitem.setProperty("inputstream.adaptive.license_key", lic_url)
listitem.setProperty("inputstream.adaptive.license_flags", "force_secure_decoder")
```
Note: This could break non-L1 devices
If that works, you need to have an add-on setting that allows people to Enable L1 Widevine Device.
Then add that above property if that's true.

I need to do this for Disney+

Testato:
ok, it work.
i tested the "force_secure_decoder" property on ubuntu, that it's not L1 device, and it have no problem, so for now i do not create special setting.

[...]

PR #661, opened 2021-04-26
Add a setting to always force secure decoder

kszaq:
Related to https://github.com/xbmc/inputstream.adaptive/issues/638

Some Android devices don't allow to pass decrypted data to non-secure decoder. Provide a setting to allow user always force using a secure decoder.


kszaq:
Closing in favour of #662

Issue #582, opened 2021-01-02
Force hardware decryption on android tv


lpuglia:
Hello, happy new year! sorry if this is not the right place but I need some clarification about how the decryption process is performed when input stream.adaptive is used.

I just bought a phillips ambilight android tv (it should have an arm cortex 53). Kodi seems to work great in all use cases I have tested so far, the only problem I came across is with encrypted streams (widevine DRM).

In particular I'm experiencing stuttering with this plugin: https://github.com/luivit/plugin.video.rivedila7
Unfortunately there is no android app for this channel, so my only chance is kodi app for android.

As far i can see the stuttering only occurs with live stream (unencrypted on demand streams work just fine). The stuttering gets more visible with the highest resolution (960x540, 1.7 Mbs).
My guess is that decryption is performed in software and the processor is no able to keep up with it (in fact I can see no problem with kodi on a pi4 which has a more beefy processor).

Can you clarify if this is a limitation of the plugin itself, inputstream.adaptive or kodi? Is there any configuration i can modify to force hardware decryption (which the tv is perfectly capable of since netflix app is in 4k)? If it is not possible at the moment is there any plan to implement it?

Thank you in advance for the clarification.

Testato:
Probably related to #638

It seems that the hw widivine decrypt is active, but it stuttering

glennguy:
As far as I know, on Android with secure decoder the decryption and decoding is done in Android's side of things. Kodi never gets to see decrypted data, IA sends Kodi encrypted data and Kodi hands it over to Android to decrypt, decode and display.

Maybe at that resolution it's still using Kodi for decoding the unencrypted data (widevine always does the decrypting)? If you can 
bring up the stream information (press 'o' key when playing, not sure how to do that on TV...) it should indicate if hardware or software decoding is happening.

Testato:
I checked it on a PC with Ubuntu, it isn't a L1 widevine certified device, but by O button Kodi tell that it do HW decoding.
So also on a non L1 device it is possibile Hw widevine decoding ?

When by O button we read (HW) it means Hw decoding and Hw decrypting ?

glennguy:
It means HW *decoding*, using AVC/h264, HEVC/h265 or whichever ***codec***. 

[...]

Issue #387, opened 2020-02-15
Call out to all stakeholders for testing v2.4.3 (Leia)


dagwieers:
@peak3d made clear he is unable to do Leia releases as he is using Matrix himself.

So this is a call out to all users and stakeholders to test **inputstream.adaptive** v2.4.3 with your add-on(s) and report back.

You can find various v2.4.3 builds at:
- [Most platforms](https://jenkins.kodi.tv/blue/organizations/jenkins/peak3d%2Finputstream.adaptive/detail/Leia_backport/3/artifacts) (Android, iOS, macOS and Windows)
- [LibreELEC v9.2.0 on RPi](http://users.telenet.be/peno/kodi/addons/inputstream.adaptive/)

If you want to do extensive testing with your add-on(s), you can find a list of tasks to complete to test the behaviour: https://github.com/peak3d/inputstream.adaptive/wiki/Test-plan

And you can report back success or failure at the Wiki at: https://github.com/peak3d/inputstream.adaptive/wiki/Test-results

*This relates to https://github.com/peak3d/inputstream.adaptive/issues/330 and https://github.com/michaelarnauts/plugin.video.vtm.go/issues/101*

dagwieers:
cc @Sandmann79 @micahg @emilsvennesson @herrnst @arvvoid @FraserChapman @CastagnaIT @Lunatixz @basrieter @skipmodea1 @enen92 @MrKrabat @phate89 @mediaminister @michaelarnauts @shycoderX

*(You are listed on the Wiki as Stakeholders: https://github.com/peak3d/inputstream.adaptive/wiki/Stakeholders)*

enen92:
@dagwieers first of all thanks for the heads up. In RTP play inputstream was used as a fallback but it looks like they just use hls streams now and ditched mpd/widevine. Either way, gave it a test on osx after installing from the zip provided above. I'm getting this error:

```
│2020-02-15 17:05:36.674 T:123145436524544   ERROR: Unable to load /Users/enen/Library/Application Support/Kodi/addons/inputstream.adaptive/inputstream.adaptive.2.4.3.dylib, reason: dlop│
en(/Users/enen/Library/Application Support/Kodi/addons/inputstream.adaptive/inputstream.adaptive.2.4.3.dylib, 1): no suitable image found.  Did find:                                     │
[...]

PR #662, opened 2021-04-27
Don't remove SECURE_DECODER from decrypter_caps

### LINKED ISSUE #638: Black Screen on Widevivene L1 Device (Samsung - Android10)

kszaq:
Many Android 10 L1 devices cannot use L1 Widevine decrypter to provide data to non-secure decoder. To prevent this, we should use the secure decoder if decrypter declares support in its capabilities.

This should be safe on Linux because `wvdecrypter.cpp` does not declare `SECURE_DECODER` in caps.

User-friendly alternative to https://github.com/xbmc/inputstream.adaptive/pull/661
May fix https://github.com/xbmc/inputstream.adaptive/issues/638 and https://github.com/xbmc/inputstream.adaptive/issues/582

Confirmed to fix the issue on my L1 Android device.

### DIFF
--- a/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
+++ b/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
@@ -67,6 +67,10 @@ msgctxt "#30121"
 msgid "Enable Pre-Release Features"
 msgstr ""
 
+msgctxt "#30122"
+msgid "Don't use secure decoder if possible"
+msgstr ""
+
 msgctxt "#30150"
 msgid "Max"
 msgstr ""
--- a/inputstream.adaptive/resources/settings.xml.in
+++ b/inputstream.adaptive/resources/settings.xml.in
@@ -96,6 +96,11 @@
           <default>false</default>
           <control type="toggle" />
         </setting>
+        <setting id="NOSECUREDECODER" type="boolean" label="30122">
+          <level>0</level>
+          <default>false</default>
+          <control type="toggle" />
+        </setting>
       </group>
     </category>
   </section>
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2080,6 +2080,9 @@ Session::Session(MANIFEST_TYPE manifestType,
   if (preReleaseFeatures)
     kodi::Log(ADDON_LOG_INFO, "PRERELEASEFEATURES enabled!");
 
+  allow_no_secure_decoder_ = kodi::GetSettingBoolean("NOSECUREDECODER");
+  kodi::Log(ADDON_LOG_DEBUG, "FORCENONSECUREDECODER selected: %d ", allow_no_secure_decoder_);
+
   int buf = kodi::GetSettingInt("MEDIATYPE");
   switch (buf)
   {
@@ -2493,8 +2496,9 @@ bool Session::InitializeDRM()
         {
           session.cdm_session_str_ = session.single_sample_decryptor_->GetSessionId();
           secure_video_session_ = true;
-          // Override this setting by information passed in manifest
-          if (!force_secure_decoder_ && !adaptiveTree_->current_period_->need_secure_decoder_)
+
+          if (allow_no_secure_decoder_
+              && !force_secure_decoder_ && !adaptiveTree_->current_period_->need_secure_decoder_)
             session.decrypter_caps_.flags &= ~SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_DECODER;
         }
       }
--- a/src/main.h
+++ b/src/main.h
@@ -212,4 +212,5 @@ private:
   bool ignore_display_;
   bool play_timeshift_buffer_;
   bool force_secure_decoder_;
+  bool allow_no_secure_decoder_;
 };


matthuisman:
so, I think the flow is

https://github.com/xbmc/inputstream.adaptive/blob/Matrix/wvdecrypter/wvdecrypter_android_jni.cpp#L482
sets SSD_SECURE_DECODER if the device is L1.

then IA was removing this flag later on here:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/main.cpp#L2492
if manifest doesn't specify its required or force_secure_decoder_  isn't set.

kszaq seems correct in saying it won't affect non-android platforms as SSD_SECURE_DECODER is only set in the Android wvdecrpyter. And SSD_SECURE_DECODER is only available on Android anyway (currently)

This would also fix the issue with Disney+ HLS which needs secure decoder on Android. Currently need to force.

Only concern is manifests that don't require secure playing back correctly with secure enabled on L1 devices
I suspect maybe that is always fine, but not the other way round which this PR fixes.
This just needs a lot of testing on L1 Android platforms.

Also, see this comment here: https://github.com/xbmc/inputstream.adaptive/issues/638#issuecomment-821818304
"On a nokia L1 device it's better without force_secure_decoder than with"
But maybe that was just a coincidence in testing due to that add-ons other issue.

@kszaq 
Please update the PR to also remove the force_secure_decoder_ property etc. as after this PR it does nothing :)
Also, need_secure_decoder_ looks to now not be needed either.

Or we could keep need_secure_decoder_ and then log a warning if it detects that is set but no SSD_SECURE_DECODER (non L1 device)
eg.
```
if (adaptiveTree_->current_period_->need_secure_decoder_ && (session.decrypter_caps_.flags & SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_DECODER))
 kodi::Log(ADDON_LOG_WARNING, "Manifest requires secure decoder which isn't available on this device");
```
(think up a better warning message)

An easy way to test is just to set force_secure_decoder in an add-on for all platforms.
That essentially does what this PR does without needing this specific build.

kszaq:
@matthuisman Thank you for your input!

I cannot agree that this PR is equal to enabling `force_secure_decoder`: for example, when the stream requires decryption but not secure decoder, it can be HW-decoded on L3 Android and Linux devices. If `force_secure_decoder` is set, it falls back to "secure" SW decoder, meaning using CPU for decoding.

The same goes for `need_secure_decoder`: an example is Netflix addon that signals it for >720p and uses either secure path on Android or secure SW decoder on Linux.

`SSD_SECURE_DECODER` is available on Linux in form of SW decoder provided by libinputstream.

My suggestion would be to keep this PR to only fix Android 10 L1 devices and if required, any further removals/cleanups can be PRd separately.

matthuisman:
need_secure_decoder_ is only used in that block you remove according to github search
https://github.com/xbmc/inputstream.adaptive/search?q=need_secure_decoder_&

same goes for force_secure_decoder
That simply stops the block you remove from executing - so it makes it the same behaviour...

So yes setting force_secure_decoder results in the same behaviour on current IA.
This just removes the need to do that.

session.decrypter_caps_.flags &= ~SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_DECODER;
That is just removing that flag. If that flag isn't set - it does nothing.

force_secure_decoder I guess then is a misleading name.
It doesn't force it. It simply stops it from being unset if IA doesn't detect it's needed in the manifest.

**This now makes me think peak must have had a reason to not use Secure Decoder when it's available but thought not needed**

So, where we need the tests is L1 Android devices using content that doesn't need secure decoder.
As long as that's not broken - we should be good.

lpuglia:
If you provide me the binaries I'm happy to test on my setup

glennguy:
@lpuglia binaries for this pr here: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-662/1/artifacts

lpuglia:
it says that the installation is successful (since the plugin can't be uninstalled I can't really check if the message is correct though). Unfortunately, i still experience stuttering, by the way my device is L1 certified and I never had the blank screen problem.

kszaq:
@matthuisman You are right, I have just checked on a LibreELEC (Linux) device and `force_secure_decoder` flag has no effect. I will add a commit to remove it.

@lpuglia Unfortunately this PR does not aim to fix stuttering with secure decoder. Actually on my Android device secure decoder works better (smooth playback) than decrypt+non-secure decoder (heavy stuttering) eventually we should aim to fix this.

Testato:
> So, where we need the tests is L1 Android devices using content that doesn't need secure decoder.
> As long as that's not broken - we should be good.

On my LA7 plugin only the two Live channels and some on-demand content need widevine, the other 95% of content is Drm free and this PR does not create any problems on my Samsung L1 device, so it is ok for me.
The stuttering instead i think is a different problem, it remain also with this PR

matthuisman:
@Testato
the stuttering isn't the same as your other buffering issue right (that is solved by rewinding a bit)?

Testato:
Yes the buffer problem is different.

CastagnaIT:
> 
> 
> need_secure_decoder_ is only used in that block you remove according to github search
> https://github.com/xbmc/inputstream.adaptive/search?q=need_secure_decoder_&
> 
> same goes for force_secure_decoder
> That simply stops the block you remove from executing - so it makes it the same behaviour...
> 
> So yes setting force_secure_decoder results in the same behaviour on current IA.
> This just removes the need to do that.
> 
> session.decrypter_caps_.flags &= ~SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_DECODER;
> That is just removing that flag. If that flag isn't set - it does nothing.
> 
> force_secure_decoder I guess then is a misleading name.
> It doesn't force it. It simply stops it from being unset if IA doesn't detect it's needed in the manifest.
> 
> **This now makes me think peak must have had a reason to not use Secure Decoder when it's available but thought not needed**
> 
> So, where we need the tests is L1 Android devices using content that doesn't need secure decoder.
> As long as that's not broken - we should be good.

only to provide some info about "robustness_level" used to change need_secure_decoder_ value,
on ntflx i always set the "robustness_level" to make sure to have it enabled on L1 devices,
but since there are cheap devices with fake widevine signed as L1, where they are not L1,
the only way to see the video images is set need_secure_decoder_  to false (by set "robustness_level" key to empty),
of course if need_secure_decoder_ is set to True on a L3 device this is ignored

^^ talking of android case only

lpuglia:
Hello @CastagnaIT, I don't want to drift off topic on this PR but i never had any stuttering problem with your plugin, even at higher than 1080p resolution. It has been a while since i was looking into the source of ntfx plugin but I have a question: is setting robustness_level something that is possible to do using python api in a basic python plugin? (sorry if I'm not clear enough but I have a very basic understanding of plugin programming) (you can read more about my issue here https://github.com/luivit/plugin.video.rivedila7/issues/40) Thanks!

glennguy:
@lpuglia setting the robustness_level will have the same effect as force_secure_decoder

matthuisman:
@CastagnaIT 
Do you have a setting to allow fake l1 devices to turn off secure decoder? With this PR, those fake l1 devices will no longer work I assume... Maybe we add a new property that can disable l1.

CastagnaIT:
@matthuisman yes currently i add `<widevine:license robustness_level="HW_SECURE_CODECS_REQUIRED" />` to every mpd
where the users that have cheap chinese android boxes (with fake L1) can force L3 in the addon settings, that in mpd will result as:  `<widevine:license robustness_level="" />`

matthuisman:
thanks. So maybe we need to keep a way to turn off secure decoder for cases like that? Do you have access to a user with a fake l1 device to test Netflix with this PR?

Testato:
How discover if a device have fake L1 certification ?
DRM info app is usefull ?

kszaq:
If I understand this correctly, "fake" L1 devices provide a "real" L1 environment, that is secure decoder will work but only with providers who don't check if the secure path is verified. In other words, Netflix, Disney (?), HBO Max (?) will probably not work with "fake" L1 device but other content will work.

I may check if I still have any of those in my drawer...

@Testato DRM Info app in not useful in this case because it will show that you have L1 device. You can tell that it's "fake" if it can't play Netflix.

CastagnaIT:
@kszaq exactly with fake L1 the videos are not playable (black screen) at least on netflx
@matthuisman at moment i don't know who to ask, i should dig up old Issue threads to see if someone answer
i have an fake L1 unit but now has been completely blacklisted by netflx... can not play video anymore...

matthuisman:
I wonder if there's an api call we can make to check secure path status? To try to detect real L1.

But I suspect maybe its the license server that requires it like on windows browsers and there will be no way to check via android api.

Well actually, if the flag is in the license request to the server, maybe we can detect that. As the blob passes us back the request data that we then post to the licence server. Or is it too late then to switch off secure decoder?

kszaq:
@matthuisman I believe IA does "try what works" on non-Android devices: https://github.com/xbmc/inputstream.adaptive/blob/Matrix/wvdecrypter/wvdecrypter.cpp#L539-L566 I am wondering if something similar could work to detect if the L1 secure path works correctly.

On the other hand, this would be working around hacks in those "fake" devices and maybe it would be easier to introduce an (expert) setting to disable signalling `SECURE_DECODER` cap to IA.

matthuisman:
An expert setting in IA to disable secure decoder would have my vote.

Testato:
Yes, it can be usefull for fake L1 device, and it is correct that you must do a manual setting in this case. But this PR solve the problem of blank screen on real L1 device, like mine Samsung and other. So probably is right merge it and only after open a new issue for  work on a manual new setting switch in expert sectiin for the fake devices

matthuisman:
i think the PR should be updated for the setting to be added.
There is no rush with this PR as we can just use force_secure_decoder which results in exact same behavior.

we merge it in as is - those chinese boxes break without a fix.
So may as well get it done now or else it could be forgotten.

Pretty easy to add a new setting.
In fact @kszaq already did something very similar on his original PR: https://github.com/xbmc/inputstream.adaptive/pull/661

Pretty much just need to do the opposite. "Force non-secure decoder"
and make it an expert setting. then pass it along like force_secure_decoder was but just invert.
```
if (force_non_secure_decoder_)
     session.decrypter_caps_.flags &= ~SSD::SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_DECODER;
```

glennguy:
>those chinese boxes break without a fix.

Yes that's pretty important, as mentioned elsewhere the plethora of fake L1 devices can still do L3 but not using the secure decoder. 
My question is what is the experience for users of those devices in the official apps for these services? Does it just fallback to L3 and user is able to watch (and if so how do we implement that here), do they get an error message, or do they get a black screen?

kszaq:
IIRC the official apps do not fallback, they fail with an error - at least that's the case with Netflix.

I'll update the PR with @matthuisman's suggestion soon. This should provide a one-stop fix for these devices.

CastagnaIT:
FYI if you use the official android ntflx app for smartphone (not android tv version) on a fake L1 the app is able to fallback to L3 automatically this without return errors (visible to user) but of course will have SD quality only
in the past i check with a reverse eng. there were several controls it's been too long, I don't remember...

CastagnaIT:
now I'm reminded one particular thing I'd seen in the rev eng.
when the app fallback to L3 it executed this command:
`drm.setPropertyString("securityLevel", "L3");`
but i don't know at what point of drm initialization

kszaq:
@Testato Yes, I believe this is correct. The reason for this PR is that most probably L1 devices running Android 10 and higher cannot provide decrypted data to non-secure decoder, my guess is to prevent dumping decrypted data to a file.

glennguy:
@CastagnaIT was your reverse engineering effort using Android APK/jadx? Any obfuscation on that section of code? Was it long ago? (might need to dig up an old APK)
It would be good to see the logic that leads to setting L3.

CastagnaIT:
@glennguy nf app use differents checks to fallback to L3 like android certification, device model etc...
but from the oldest apk vers (find "evolucionNetflix github" on google it already reversed) the things are a bit changed in new apk versions, so not use all the same behaviours of old versions. New apk versions (obfuscated) and are not easy to understand, ~also if deofuscated the methods/vars names are all wrongs... (if you want defuscate you can try use "deguard" website)~ but in the middle of all the mess you still manage to understand something
edit: `jadx-gui` seem has been fixed from last year now provide better results without crashing nice

kszaq:
I have tested this PR on the following devices:
* LibreELEC and CoreELEC: no change, everything working as before regardless of forcing disable secure decoder or not
* Android 10 L1 certified: no more black screen with secure streams when secure decoder is *not* disabled (default with this PR applied)
* Android "fake L1" device is banned from Google's certificate provisioning and does not work with or without this PR
* fake L1 device after downgrading to L3:  everything working as before regardless of forcing disable secure decoder or not

glennguy:
@kszaq thanks for all your work and testing on this, looking really good!

I've been looking at some decompiled android netflix code (v4.16.1 that @CastagnaIT was referring to above from ~2017) and their flow seems to be:
* If the app had a fatal error and device restarted - set an app preference to fallback to L3
* If there was a provisioning error with L1 - fallback to L3 (some code didn't decompile correctly but I'm pretty sure). It calls drm.provideProvisionResponse and if an exception is raised then sets the level to L3.

kszaq you tested with a fake L1 device correct? Is it printing `Exception on provideProvisionResponse` in the log? If so I have an idea to try. I'd do myself but I don't have a fake L1 device handy.

We could try setting a retry point at the start of this function:
https://github.com/xbmc/inputstream.adaptive/blob/99d72cdc57a783c1c88bc23ba231a2dd2bc16b60/wvdecrypter/wvdecrypter_android_jni.cpp#L497-L499

...and doing a goto, similar to how it's done here:
https://github.com/xbmc/inputstream.adaptive/blob/99d72cdc57a783c1c88bc23ba231a2dd2bc16b60/wvdecrypter/wvdecrypter_android_jni.cpp#L405-L420


What are your thoughts? I know it's outside the scope of this PR description, we can implement later.

kszaq:
@glennguy Apologies but I have deleted the logs now and downgraded the device to L3 by removing `liboemcrypto.so` - I will check if I can get it back to "fake L1" level. But please keep in mind that with L1 provisioning failing most other apps (from content providers) will fail to play and we would be working around issues with grey-area devices that IMHO users should deal with in the first place to make the device usable. 

I have amended the last commit to allow disabling secure decoder only for streams that do not have `need_secure_decoder_` - this would allow to disable secure decoder when possible but still get full-resolution L1 content - in other words, enabling this setting would force "old" IA behaviour. The setting to disable secure decoder is useful for devices that have better results - I'm referring to https://github.com/xbmc/inputstream.adaptive/issues/638#issuecomment-821818304

IMHO this PR is ready to go and we can try to work around "fake L1" in a separate PR - if I find a way to force L3 in IA.
Edit: I looked at the code and I think that using liboemcrypto in embedded in Android framework and we can't make it force to not use it, correct?

glennguy:
If we get a failed provisioning on fake L1 devices then we should be able to detect in wvdecrypter and do a `media_drm_.GetMediaDrm()->setPropertyString("securityLevel", "L3");` or similar
edit: and if so then we will no longer need to set the new setting, everyone wins!

kszaq:
Thank you! I will look into it and create a separate PR, because the device has the same provisioning error without this PR. For the reasons mentioned in my previous comment I would keep the expert setting that allows switching to pre-PR behaviour if user prefers to use it. Do you agree?

matthuisman:
i think with that new PR, we could remove the FORCENONSECUREDECODER setting.
Seems it wont be needed anymore? we have the fake L1 covered with that new PR.

Its just going to cause more issues than it solves
ie. user enables it by accident or is doing "try anything" approach to fix an issue
and makes this PR a lot simpler :)

matthuisman:
Ignore my last comment. Ive chatted with Glenn and agreed to keep the setting as a backup for now in case we have stuttering users. We could look at removing in future.

matthuisman:
@Testato 
Can you advise more about this stuttering issue.. Is it on all addons that use IA? Or just that single addon?
If we add a setting to force l3, then addons like Disney+ won't work on your device as it uses hls which has no way to indicate l1 required

If its a per addon issue, then we can make it a property instead of a setting so addons can turn off l1. 

@kszaq
What would be really cool and solve all our issues is if we just tried l3 always first. If that decodes ok, use it. If not, then try l1. But not sure if that's possible

lpuglia:
@matthuisman the stuttering problem is documented here: https://github.com/xbmc/inputstream.adaptive/issues/582

matthuisman:
@lpuglia 
please answer the other questions.
Does stuttering happen on other add-ons?
If you install IA with the above PR, and try other add-ons that use Widevine - do you get stuttering??

lpuglia:
just tested: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-662/5/artifacts/
the problem persists with both enbled and disabled "Disable secure decoder if possible" option.

about other apps, long story short: I wasn't able to find any other video stream with this specific combination of MPD and non-enforced widevine hardware decrypt, here is the manifest:
https://controlc.com/4eca64bc

The stuttering is visible only for the resolutions 1280x720.

matthuisman:
it doesnt matter if its forced or not.
this PR is same as forcing secure decoder.

all you need to do is install this version of IA then try various Widevine streams.
then see if they stutter or not...

If not - then it's specific to live streams on a single add-on....

lpuglia:
@matthuisman my problem is only with this specifc video stream. All the others i tried have always worked (and still work) flawlessly. But this doesn't mean anything, it may well be that other streams (which I'm not currently aware of) have the same problems.
N.B. This video stream has no stuttering on the same device using google exoplayer.

kszaq:
@matthuisman The latest commit in this PR is a workaround for users who have better results with non-secure decoder than secure. But for example, my 2 Android devices work better with 15Mbps 4K HEVC video when secure decoder is used.

If we tried to use L3 first, we may get working SD stream an until L1 is tried we cannot know if the device supports 1080p or 4K. I think we should always use the highest level device can initialize as it's easier to downgrade to L3.

In my opinion we can keep the workaround until a way to solve the stuttering is found.

glennguy:
Perhaps we need to consider having both the global (settings) option and the add-on specific (listitem properties) option. Maybe the add-on specific setting should override.

Consider the following:
User of LA7 add-on has stuttering with secure decoder 720p stream, but Disney+ and NF work fine in whatever resolution up to 4K. Disney+ needs to be able to force L1 as HLS has no 'robustness_level' equivalent to signal L1 to IA, so using the global "don't use secure decoder if possible" will stop user from viewing FHD/4K streams. We don't want user to have to change the setting each time they use a different add-on.

Edit: posted here: https://github.com/xbmc/inputstream.adaptive/issues/582#issuecomment-836638153 to how this issue seems to be quite widespread.

Changing default behaviour and forcing this on users with this PR could be inviting a lot of trouble.

kszaq:
@glennguy This PR is required to fix black screen on Android 10 devices. It needs to be default. I agree that we shouldn't break other devices, so maybe I'll change the PR to provide a setting that restores 1:1 "old" behaviour?

kszaq:
Force-pushed the PR branch to the last approach: don't turn off secure decoder by default but provide a setting to use "old" behaviour. IMHO secure decoder should be used by default as this is a fix for many Android 10 devices.

matthuisman (approved):
looks good :)

CastagnaIT:
@kszaq the two options should not exclude each other
has you done, this is no more per addon configurable,
each user has to enable/disable the settings every time...please let the addons to choose automatically

Testato:
@CastagnaIT 
For "automatically" you mean that IA should check if a device is L1 or L3, and automatically set SecureDecoder only of it found an L1 device ?

This is correct to me also, some with fake L1 device may have problem, but it is correct that him must manually change some setting, not me that have a real S9+ L1 device.

@lpuglia the stuttering problem is a different problem, this PR is for the blank video problem on real L1 device.
After this PR we can open a new issue with all the case where real L1 device have stuttering on widevine protected stream, it need separate test and separate issue imho

CastagnaIT:
@Testato no, i use "robustness_level" tag (on DASH file) to set it, but in this way a user have to enable/disable also the ISA setting to make it works
i think best way is add a new listitem property to allow every addon to force remove the caps when needed

matthuisman:
This code is removing secure decoder. Not setting it. I think youll find
logic is.actually ok. I'll post again soon once on a pc

CastagnaIT:
then the use of "robustness_level" tag on DASH because it is useless,
Peak has told me in the past to always set it to "HW"
and i currently set it to empty string in case of problems (black screen) with fake L1
so now i can delete this behaviour from my addon?
I can't do tests so will review in the future in case of problems

kszaq:
@CastagnaIT Without this PR the logic is: don't use SECURE_DECODER unless addon requests it either by setting `force_secure_decoder` or `need_secure_decoder`.

With this PR: use SECURE_DECODER when available, even if addon does not request it.
With this PR and expert setting turned on the old logic is restored, i.e. don't use SECURE_DECODER unless addon requests it either by setting `force_secure_decoder` or `need_secure_decoder`.

I have checked my "fake L1" device and it fails to initialize Widevine, there is a second PR that addresses it and downgrades to L3.

CastagnaIT:
ok then after this i can remove the use of "robustness_level" tag on DASH, it is right?

matthuisman:
Kind of. You may want to keep it to HW when known is needed in case users have that new addon setting enabled. Only reason for that setting is because of that stuttering issue. We hope to remove it once that's solved. Its just a backup. 

But no need to set to blank to stop secure. That other PR will automatically fallback to l3 if l1 fails regardless of any of these settings

Testato:
> With this PR: use SECURE_DECODER when available, even if addon does not request it.
> With this PR and expert setting turned on the old logic is restored, i.e. don't use SECURE_DECODER unless addon requests it either by setting `force_secure_decoder` or `need_secure_decoder`.

I agree

> I have checked my "fake L1" device and it fails to initialize Widevine, there is a second PR that addresses it and downgrades to L3.

Ok, this is the correct logic imho.

CastagnaIT:
@matthuisman thanks i understand now sorry for the noise...
Little OT:
i would just like to advise you to keep Wiki updated, also with future others PR's,
it would be better to put it as a thing "to do" for each PR author

matthuisman:
if it wasn't for that stuttering issue where a L1 device plays better with L3 on certain streams, this PR would be a lot simpler.

Eventually is will mean add-ons don't need to use force_secure or robust level.
IA will just use L1 if works, else use L3.

A few PRs lately are geared towards IA being a bit smarter like this.
eg. we no longer need to set manifest_update_param to "full" to make IA refresh manifest for live dash - it does this automatically now when detects its required.

After this PR is merged and a new version IA is released, we will see how many users have stuttering issues.
We can tell them to enable that setting to fix (stops angry mob)
At least we will be able to gauge how many devices have stuttering issues

But yeh, we don't know how widespread the stuttering issue is... so best to keep a backup setting for now :)

Maybe https://github.com/xbmc/inputstream.adaptive/pull/674 should have actually just been a separate commit on this PR to help reduce confusion. 
it is related and fixes an issue that this PR introduces with fake l1 devices

glennguy (approved):

glennguy:
@matthuisman @CastagnaIT @kszaq @Testato We're all happy now? :)

glennguy:
Thanks @kszaq

kszaq:
Fingers crossed we won't receive (m)any complaints because of this patch. 😃

Testato:
Bump to v2.6.15 ?

michaelarnauts:
> Bump to v2.6.15 ?

I would propose to first look at a pre-release or something that can be tested more globally. Like has been done in https://github.com/xbmc/inputstream.adaptive/issues/387

IA is a critical piece of the Kodi ecosystem, and releases really shouldn't be rushed IMHO.

kszaq:
Ideally repos should hold at least a few previous versions to allow users quickly downgrade...

matthuisman:
Maybe we could have an issue where users who want to test can subscribe to it. 

A release PR is created.
He posts in the issue so all users who have subscribed get notified. 

Users then have X days to test that commit and report any regressions only. All reports must have a new github issue with logs showing regression between last version and new.

After X days, it is released.

We can rapid release now as android uses repo. So any critical bugs that still sneak through can be hot fixed within 24 hours.

kszaq:
@matthuisman My point is that if you create a release PR, Jenkins will build it and you can get a few (power) users to test it. This is an obvious pre-release step. But after a release, the repo maintainer should keep previous versions available so that in case someone encounters an issue, they can revert from Kodi menu.

Sadly, on most if not all platforms only the latest version of binary add-ons is available.

And many people are not aware that Kodi stores previous versions in user's data folder. 😉

matthuisman:
that would require Kodi to update their repo pushing code so it keeps old versions listed.

it would affect all kodi repo addons so not sure how'd you go.
each version kept requires duplicate addons.xml code 
So if every add-on had 1x old version kept - you've doubled the size of addons.xml
double size addons.xml + millions of users = huge traffic increase to kodi servers

dagwieers:
If we make available binary add-ons for most popular platforms (e.g. Windows, macOS, Android, LibreELEC, Raspbian, Ubuntu) and involve our stakeholders (both add-on developers, as well as distributors) they can involve developers and users for wider testing.

We organized this once before (when retrofitting updates from Matrix to Leia) and it worked out quite well. https://github.com/xbmc/inputstream.adaptive/wiki/Test-results

I think we can organise this again for both Leia and Matrix and ask people to test each platform and report back. If we give it a week for people to do this, it is plenty of time to learn if anything is blocking or bisect to find the cause.

(We can also extend our list of testers every time we reach out, so we get better at this too)
https://github.com/xbmc/inputstream.adaptive/wiki/Stakeholders

matthuisman:
Waiting for every test will slow progress to a standstill.

It should be
upto the devs etc to be active and test when notified a release is pending.

And a time limit as well. 

 Not sure if kodi will want to provide arm builds as that's not one of their official platforms after all. Sure the devs can build arm. Or a 3rd party (yourself?) could always do it?

dagwieers:
> Waiting for every test will slow progress to a standstill.

Not sure what you mean. This would only be needed for (important) releases to catch any obvious regressions impacting one or more services/streams.

> And a time limit as well.

I proposed a week, to ensure wider testing but also to put all builds in the hands of testers.

> Sure the devs can build arm. Or a 3rd party (yourself?) could always do it?

That is what we did before, but it just takes more time to got the builds to testers. Automated builds would help get more people testing within the limited timeframe, or reduce the testing time for all.

Whatever works best.

glennguy:
Yes on one hand releases shouldn't be rushed for the sake of it, however I don't think it would be great to end up with too much changes in between releases either. Look at the activity in this repo right now compared to just over 2 months ago. Keep this in mind - while a manual testing process plays out there's going to be plenty of other commits going in and there's already a bunch since the last release. If we want to do a manual testing process every release then we're going to get hamstrung fairly quickly by it. 

> If we make available binary add-ons for most popular platforms...

Let's look at getting the GHA script updated to include more platforms and to store the artifacts as Jenkins only has Win/Mac/Android

I'm not against having some distributed runtime testing but I'm wary of doing this process every release.

Regarding this PR:
Yes it has been established that some devices will have issues with stuttering when using the secure decoder. There is a setting added to revert behaviour introduced to how it was before. On the plus side with this PR and #674 devs get to avoid dumpster fires like @matthuisman has had to put up with on Disney+ regarding black screen.

I'll create a release PR and copy the template from #387.  Please add whoever you'd like there.

matthuisman:
I must admit an arm build would actually be pretty handy if can be done relatively easily

Issue #580, opened 2020-12-30
NBA League Pass no sound in live games Fire Tv


scottydulton:
i can get sound on both live games and archived games running kodi 18.9 Leia on pc, however on kodi 18.9 Leia on fire tv i cannot get sound on live games where i can in archived games. can anyone notice why with the 2 logs?

[kodi live game.log](https://github.com/peak3d/inputstream.adaptive/files/5753138/kodi.live.game.log)
[archive game.log](https://github.com/peak3d/inputstream.adaptive/files/5753139/archive.game.log)

ivankokan:
Hi everyone, please check all the logs and relevant info within the mentioned issue.

Summary:
- Fire TV
- Live videos (NBA games, guys are using the latest from my fork https://github.com/ivankokan/xbmc.plugin.video.nba, not the original repository where the mentioned issue is submitted)
- The issue happens when IS.A Media is set to **All**, but it does not when set to **Audio**.

@CamiloHT @scottydulton @MRLB Please continue here.

MRLB:
@peak3d Can you help us? Is are bit sad to watch NBA without sound :-(

thanks in advance :-)

glennguy:
It appears that it is not recognising that the audio stream is encrypted on android/live. Note that the archived game is DASH whereas live is HLS so that would give us a good place to start looking :)

scottydulton:
@glennguy thanks mate, do u think there's anything we could manually copy from IS.H 2.6 to the Leia IS.H code, instead of waiting on awhole new release?

glennguy:
Have you tested live games on Android/Matrix? Does it work properly in that case?

scottydulton:
@glennguy I tried to fresh install matrix on both pc and android but I can't get the nba pass to install, I keep getting the error 'the dependency on xbmc.python version 2.20.0 could not be satisfied' I have no idea why

glennguy:
@ivankokan is there a matrix compatible version available or in the works?

scottydulton:
@glennguy i noticed on matrix there wasnt the script for simple xbmc downloader in services like in Leia, i tried to google i only found netflix compatible versions that had been adapted for matrix

glennguy:
@scottydulton would you like to test this build of inputstream.adaptive for me?
[android-armv7-leia-v2.4.6.zip](https://github.com/peak3d/inputstream.adaptive/files/5785372/android-armv7-leia-v2.4.6.zip)

[...]

PR #663, opened 2021-04-29
[Widevine] [HLS] Handle psshsets correctly

### LINKED ISSUE #592: [HLS] [Widevine] Streams with more than 1 unique psshset fail to play
HLS/Widevine seems to still be a relatively rare combination so I think that's why this problem hasn't been exposed until now.

There seems to be an issue with calling InitializeDRM() within the same period more than once. It's hard to know exactly what at this stage (I've looked a lot but no luck) but it looks to be causing widevinecdm.dll to crash in Windows, and no proper decryption in Android. I'll investigate Android again at a later date so I can check logcat for some clues, perhaps the error will be revealed there.

The proof for this is in a test stream which I've created and posted [here](https://github.com/glennguy/wvtest) - this stream is affected by the same issue as #393 but it doesn't have any bearing on this issue, just thought I'd mention that there's nothing actually wrong with the stream itself.

There are both mpd and hls manifests for the same content. The mpd plays fine as expected, hls will crash on Windows.

If I however hardcode the particular psshsets to be added on the first prepareRepresentation call i.e
```
if (adp->type_ == VIDEO)
        {
        periods_[0]->encryptionState_ |= ENCRYTIONSTATE_SUPPORTED;
        // 720p
        parseLine("#EXT-X-KEY:METHOD=SAMPLE-AES-CTR,URI=\"data:text/plain;base64,AAAAOHBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABgiEGZrajNsamFTZGZhbGtyM2pI49yVmwY=\",KEYID=0xEB676ABBCB345E96BBCF616630F1A3DA,KEYFORMATVERSIONS=\"1\",KEYFORMAT=\"urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed\"", 11, map);
        processEncryption(base_url, map);
        periods_[0]->adaptationSets_[0]->representations_[1]->pssh_set_ =
            insert_psshset(NOTYPE, periods_[0], periods_[0]->adaptationSets_[0]);
        // 480p
        parseLine("#EXT-X-KEY:METHOD=SAMPLE-AES-CTR,URI=\"data:text/plain;base64,AAAAOHBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABgiEGZrajNsamFTZGZhbGtyM2pI49yVmwY=\",KEYID=0x0294B9599D755DE2BBF0FDCA3FA5EAB7,KEYFORMATVERSIONS=\"1\",KEYFORMAT=\"urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed\"", 11, map);
        processEncryption(base_url, map);
        periods_[0]->adaptationSets_[0]->representations_[0]->pssh_set_ =
            insert_psshset(NOTYPE, periods_[0], periods_[0]->adaptationSets_[0]);
        // audio
        parseLine("#EXT-X-KEY:METHOD=SAMPLE-AES-CTR,URI=\"data:text/plain;base64,AAAAOHBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABgiEGZrajNsamFTZGZhbGtyM2pI49yVmwY=\",KEYID=0x639DA80CF23B55F3B8CAB3F64CFA5DF6,KEYFORMATVERSIONS=\"1\",KEYFORMAT=\"urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed\"", 11, map);
        processEncryption(base_url, map);
        periods_[0]->adaptationSets_[1]->representations_[0]->pssh_set_ =
            insert_psshset(NOTYPE, periods_[0], periods_[0]->adaptationSets_[1]);
        retVal = PREPARE_RESULT_DRMCHANGED;
      }
```
and comment out the reassignment of `retVal` during parsing EXT-X-KEY this will ensure that the cdmsessions are all set up on the first full run through of InitializeDRM (opening first video stream) and that the subsequent prepareRepresentation on the audio stream will not allow another call to InitializeDRM (PREPARERESULT_OK). 

The video then plays correctly.

Why hasn't this been an issue in the past?
With Disney+ the video representations all use the same key so switching between reps doesn't trigger any additions to psshsets.
With the NBA League Pass add-on it is a slightly different situation though as the audio reps are encrypted as well using the same key as all video reps.
This in combination with the fact in HLSTree.cpp we call insert_psshset with `NOTYPE` instead of the actual adaptationset->type_ causes both audio and video (and other video representations in case of manual streams) to use the same psshset, so only the first representation prepared will cause InitializeDRM to be invoked - second prepareRepresentation will return `PREPARERESULT_OK`.

I've tried a lot of things like inserting numerous log statements and also sleep statements at various points in order to safely attach the debugger. I've compared the states of cdm_sessions_ between the hls and mpd for any obvious differences but can't find anything that stands out.

From what I can tell Kodi crashes as soon as the first non-init segment has finished downloading (cdm is passed samples to decrypt) but I will try and verify.

glennguy:
Due to the workflow of HLS (explained in more detail in [this comment](https://github.com/xbmc/inputstream.adaptive/issues/580#issuecomment-817248878)) InitializeDRM() is called multiple times compared to the once with DASH. The cdm_sessions_ array already has a session with a reference to a single sample decrypter which will be lost on the second invocation of InitializeDRM() :
https://github.com/xbmc/inputstream.adaptive/blob/da68ef0983b21149d2b3401cada28c58cde24a64/src/main.cpp#L2304
https://github.com/xbmc/inputstream.adaptive/blob/da68ef0983b21149d2b3401cada28c58cde24a64/src/main.cpp#L2435-L2443

Now we're using a new SSD but there must be some point in the code that is still referring to the old SSD. I can only speculate though, due the the almost impossible nature of debugging widevine matters on any platform other than Android it has been hard to determine the exact point in which it hard crashes. It is pretty clear now though why IMO.

With this PR we assign a default value in the struct definition to avoid this.

A change has also been made to stop `OpenDRMSystem` from being needlessly invoked more than once.

Fixes #592, Should be a solution to #580.

Edit: 
The latest commit properly fixes #592. Where we had the same pssh set in use across multiple representations (NBA add-on case) the representations >1 (eg. audio) weren't being set as 'encrypted' which on would prevent a crypto session from being set up. Now this code path will happen correctly. 

Additionally the pssh type for widevine hls is now being correctly set to the adaptationset type rather than `NOTYPE` which is only logical for .ts/AES-128 streams with their combined audio/video.


glennguy:
@scottydulton @chamchenko test zips here: ~~https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-663/2/artifacts~~ outdated

chamchenko:
Hi @glennguy 
Thanks for your effort
At my first try I got sound but with some glitches, then my android box hang, and I had to reboot it
Then I retried and it hanged again
and the third try got no sound here is a debug log 
[LOG](https://paste.c-net.org/LoanedFragment)

glennguy:
Thanks for log @chamchenko 

I think I have identified the issue (or at least ***an*** issue). Also for future with logs - component logging is not normally needed, and if so probably only ffmpeg :)

glennguy:
@scottydulton @chamchenko updated zips: ~~https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-663/3/artifacts~~ outdated

chamchenko:
@glennguy 
I will try with tonight's games and let you know how it goes

glennguy:
Hard crash happening now (not on my test stream but on NBA) - investigating...

chamchenko:
I just tested and kodi crashes at playback start
here is a log
[LOG](https://paste.kodi.tv/woqomopota.kodi)

glennguy:
I've recreated the issue with a local stream, will hopefully get sorted later today

glennguy:
Fingers crossed: ~~https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-663/4/artifacts~~ outdated

My test stream which is now almost 1:1 with the NBA layout works well. We will all be able to see in another 8-9 hours :)

chamchenko:
I just tested and It works

scottydulton:
@glennguy i was at work so i tested on my phone on aarch64, played nba tv fine but hard crashed trying live games, what device did you test on @chamchenko

chamchenko:
> @glennguy i was at work so i tested on my phone on aarch64, played nba tv fine but hard crashed trying live games, what device did you test on @chamchenko
I tested on an amlogic android box kodi armv7

@scottydulton 
be aware start from the beginning does not work on the plugin (it's a plugin related issue I'm working on it) it's not ISA related
so to test use go live

scottydulton:
@chamchenko yep im aware of the start from beginning feature, hard crashed on me when selected "go live" i was on my phone though, samsung s20+

glennguy:
@scottydulton  make sure 100% that you're using the latest build that I posted above and not one of the earlier - file names are all the same

glennguy:
Hmmm might still be an issue regarding multiple periods/discontinuities, just crashed on open for me

chamchenko:
> Hmmm might still be an issue regarding multiple periods/discontinuities, just crashed on open for me

Could it be related to the no audio when provided with resume time that I mentioned  [here](https://github.com/xbmc/inputstream.adaptive/issues/664#issuecomment-830481183)

glennguy:
Maybe? We will see. I'll attempt to fix the crashing issue tonight. I'll also simulate my test stream being a live stream (and also setting a resume point to ffwd to the live edge) , it should hopefully expose if the issues are the same thing or not.

glennguy:
Testing done with a stream that is:
* HLS
* Widevine
* Multi-period (2 periods, both Widevine)
* Live (no new segments are brought in but it refreshes after each segment)
* 2 video representations - stream switching tested in 'manual' mode.

```
#KODIPROP:inputstream.adaptive.license_key=https://proxy.uat.widevine.com/proxy||R{SSM}|
https://raw.githubusercontent.com/glennguy/wvtest/mp-live/h264_master.m3u8
```

Tested on Android and Windows 10 with no issue.
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-663/5/artifacts

chamchenko:
@glennguy 
So as I mentioned I have been continuously working on an NBA plugin for the last few days, I now have a beta version that implements the resume time. so if you want you can use it to further investigate why there is no audio when provided with resume time
the code is [here](https://github.com/chamchenko/plugin.video.nbainternational)

Hey and thanks again for your hard work it's really appreciated.

Edit:
I noticed small difference between your manifest [here](https://github.com/glennguy/wvtest/blob/mp-live/h264_720p/main.m3u8)
and nba stream 
At the beginning of the manifest there an other discontinuity tag whichi is #EXT-X-DISCONTINUITY-SEQUENCE:0
ex:
`#EXTM3U

#Endeavor Streaming Server 6.4.14.20201223.32870

#EXT-X-VERSION:7

#EXT-X-PROGRAM-DATE-TIME:2021-04-29T22:17:19.737Z

#EXT-X-TARGETDURATION:6

#EXT-X-DISCONTINUITY-SEQUENCE:0

#EXT-X-MEDIA-SEQUENCE:269955773

#EXT-X-INDEPENDENT-SEGMENTS

#EXT-X-MAP:URI="br_3000_video-init.mp4`

I don't really know if it really have an impact on how kodi handles the chapter thing. I just wanted to let you know in case you haven't noticed

glennguy:
Oh I forgot to add that tag. Luckily since it is 0 and doesn't change I don't think it would affect anything, it is responsible for making sure that periods are removed once they drop off the playlist.

chamchenko:
What I noticed now, is that when seeking forward or backward between chapters with right and left keys of the keyboard it creates awkward behavior. It results on loss of audio in both ways (which explain why I'm not getting sound when setting resume time in my code). and when seeking backward it results on infinite replaying of the last segment of the first chapter and the player fail each time to make the transition to the next chapter.
What I noticed also is that in normal cases when seeking forward or backward there is these line that appears:
`inputstream.adaptive: PosTime (37681)

inputstream.adaptive: seekTime(37.7) for Stream:1 continues at 42.0 (PTS: 1620084654148115)

inputstream.adaptive: seekTime(42.0) for Stream:2 continues at 42.0 (PTS: 1620084654129501)

inputstream.adaptive: seekTime(37.7) for Stream:1 continues at 42.0 (PTS: 1620084654148115)

inputstream.adaptive: seekTime(42.0) for Stream:2 continues at 42.0 (PTS: 1620084654129501)`

But when jumping between chapters:

`inputstream.adaptive: PosTime (2002416)

inputstream.adaptive: SeekTime: seeking into new chapter: 2

inputstream.adaptive: EnableStream(1001: false)

inputstream.adaptive: EnableStream(1002: false)

inputstream.adaptive: Reusing DRM psshSets for new period!

inputstream.adaptive: DEMUX_SPECIALID_STREAMCHANGE

inputstream.adaptive: GetStreamIds()

inputstream.adaptive: GetStream(2001)

inputstream.adaptive: GetStream(2002)

inputstream.adaptive: OpenStream(2001)

inputstream.adaptive: Selecting stream with conditions: w: 0, h: 0, bw: 400000

inputstream.adaptive: Download https---- audio-init.mp4 finished, avg speed: 841168.68byte/s, current speed: 51370.00byte/s

inputstream.adaptive: Download https---- audio-init.mp4 finished, avg speed: 840126.59byte/s, current speed: 53346.00byte/s

inputstream.adaptive: UpdateSampleDescription: codec 1836069985

inputstream.adaptive: GetStream(2002)
here it does the same for video-int.mp4
inputstream.adaptive: seekTime(214.4) for Stream:1 continues at 210.0 (PTS: 1620086610135481)`

at this point I think the expected behaviour is to repeat the same line for the Stream 2

what really happens is the player from here on only downloads video segment as if audio stream is disabled


Edit
When I set media to audio only in inputstream adaptive settings, jumping between chapters does not result in loss of audio. 
and as expected I get that line in log
inputstream.adaptive: seekTime(920.0) for Stream:2 continues at 918.0 (PTS: 1620087318131678)

glennguy:
Confirmed, happening here too.

I can see in the test stream that the wrong audio segment is being seeked to. Once the stream starts, immediately seek 30s forward:
```
CurlFile::Open(0x2c1f272e870) https://raw.githubusercontent.com/glennguy/wvtest/mp-live/h264_720p/7.m4s
2021-05-04 21:07:15.082 T:24612   DEBUG <general>: CVideoPlayerAudio - CDVDMsg::GENERAL_PAUSE: 0
2021-05-04 21:07:15.083 T:24612   DEBUG <general>: CDVDAudio::Pause - pausing audio stream
2021-05-04 21:07:15.511 T:23880    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(19.2) for Stream:65537 continues at 18.0 (PTS: 36080000)
2021-05-04 21:07:15.511 T:11592   DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'connection: keep-alive'
2021-05-04 21:07:15.511 T:11592   DEBUG <general>: CurlFile::Open(0x2c1f272f8b0) https://raw.githubusercontent.com/glennguy/wvtest/mp-live/audio/10.m4s
2021-05-04 21:07:15.866 T:23880    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(18.0) for Stream:65538 continues at 35.9 (PTS: 54016000)
```

Here `seekTime` represents the time past the start of the chapter which is roughly correct, however the continue time is way off.

There is no sound after seeking/playback resumes, but wait 10 or so seconds and audio comes back.

This issue is present in non widevine streams as well (which is a blessing really, sooooo much easier to solve when you can properly debug), so not really to do with the subject of this PR. However I will come up with a solution as a separate work.


Edit:
Found where the seek is going wrong. I thought I had a solution, it fixed seeking into another chapter but broke seeking in chapter > 1. Will work on again tomorrow.

Also found another crash bug in this implementation which affects streams where there's different keys between periods, think I have that sorted though.

matthuisman:
did you say this PR should also fix https://github.com/xbmc/inputstream.adaptive/issues/667 ?

glennguy:
I need to confirm after applying the pts fix, I think it will

glennguy:
Will rebase this tonight and we can hopefully get some positive tests

glennguy:
@chamchenko could you please test?
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-663/7/artifacts

chamchenko:
@glennguy I will test with tonight's games and let you know.

chamchenko:
Hi @glennguy 
I just tested so when no resume time is provided the audio works fine on android armv7.
But when I provide a resume time still no audio here is the log

[LOG](https://paste.kodi.tv/tudowiviva.kodi)

glennguy:
Thanks for the update, I'll have a look again tonight, something is still not quite right

scottydulton:
I tested the games that started pre-midnight GMT they worked fine, then the same games after midnight GMT had issues

chamchenko:
@glennguy 
I just realised that in my first test that stream contained just one chapter, so no seeking across chapters border but yet the audio does not work.
On my second try (pass midnight GMT that discontinuity tag appears) the issue is different this time.
in the progressbar the position it start's from is negative
here is a screenshot and the log
[screenshot](https://i.ibb.co/YpZK3R2/screenshot00001.png) 
[LOG](https://paste.kodi.tv/delujuzeba.kodi)

glennguy:
OK I think I have figured out what is going wrong, I'll post a new issue. No solution just yet.

glennguy:
@chamchenko with regards to the really bad timeline position, could you please test with a game tonight - please set 'Media' in inputstream.adaptive settings to 'Video' and see if the same thing happens (start stream once passed midnight GMT).

glennguy:
Issue describing the problem - #681

scottydulton:
@glennguy @chamchenko I just tested the midnight gmt game and using your 'dev' version it kept going back to like -0.05 seconds over and over, then I used the 'test' code at 12.25am gmt time and when the player started I skipped to the next chapter which was 1hr35mins/2hours and then sound resumed and from there I could seek to any position between 1.35 and 2.00hrs with no issues

chamchenko:
Hi @glennguy 
It's the same when selecting video only it start's from a negative position.
@scottydulton 
with my dev version did it start at the end or at the beginning when you tested.
otherwise yes I already remarked that when you skip to next or previous chapter (using up/down keys) the audio resume

Edit:
scottydulton could you make sure you're testing with the last version of inputstream adaptive that glennguy posted.
In all my test on android with his last version I'm unable to go into the second chapter as each time I do It whether by going live or by seeking forward it seeks hours behind the start of the stream (the negative start point) that we are talking about

glennguy:
@chamchenko thanks for the confirmation.

I think I will try and get a local 'snapshot' of an NBA stream downloaded so I can test/debug at a convenient time.

Actually I'll try some alterations with my test streams to make it 'live' and it might give some clearer direction for which way to go in #681

Chamchenko - what happens if you set `listitem.setProperty('inputstream.adaptive.play_timeshift_buffer', 'true')`

chamchenko:
It does not help, always comes back to a negative point when I try to seek into the second chapter.

Otherwise what I noticed thank's to scottydulton comment that when no resume time is provided and the stream start at the beginning of the first chapter the video plays correctly but the progress keep reseting to 0 or to a very small negative amount of seconds ex: -00:00:05
I think scottydulton thought it was just going back over and over just because it was in the pregame blue screen (game will begin shortly) so there is no way he could have noticed that the video was continuing to play not replaying

And I think this is related to the fact that the player is reseting the progress to the position in the last update of the manifest, meaning that each time inputstream adaptive downloads the m3u8 list there are segments at the beginning that are dropped out of the list which means that the segment that is beeing played is at the start of the new list or dropped out of it.

glennguy:
Correct, the -0:00:05 is normal - every 4 or 5 seconds (whatever the length of the segments are) the playlists will refresh. Typically one segment is removed from the beginning of the stream, one added at the end, and the 'now' time should constantly play through the same 4 second section and reset back 4 seconds at the end. When you start the stream from 0, this is what happens too, just that it can go slightly negative.

glennguy:
Could I ask that someone get me a relatively close in time copy of the audio and video m3u8 (only need one video m3u8, pref the low bitrate one) and post files here? I can't access atm

chamchenko:
Here are the m3u8
[Audio](https://paste.kodi.tv/izubukikoj)
[Video](https://paste.kodi.tv/olifolagof)

glennguy:
Thankyou!
I'm hoping to write a small script to bring them down locally so I can add my own discontinuity and host somewhere. Might see about discovering the key too so I don't have to worry about WV!

chamchenko:
wouldn't you need the base url then?

glennguy:
I used the baseurl in the log you posted somewhere above and it downloads. Possibly a different feed but that's not important. Hopefully they're still valid later today.

scottydulton:
any luck @glennguy

glennguy:
Working on it later tonight and this weekend :) I'll keep you posted

glennguy:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-663/8/artifacts
Ok just for curiosity I've applied a hack that should make seeking work, but I don't want this to be the final solution as it forces downloading the first segment of the new chapter. This causes a longer than expected delay before playback resumes... much rather do it properly! At least though we can test this subject of this PR works properly.

@dagwieers @michaelarnauts @mediaminister could someone test VTM GO please :)

mediaminister:
Yes, with this hack I have always audio when seeking, skipping with mouse pointer or keyboard controls.

glennguy:
> Yes, with this hack I have always audio 

Awesome, thanks!

scottydulton:
@glennguy this is great!

chamchenko:
@glennguy 
So I tested with this build when I hit start from the beginning (seeking in the first chapeter) I have audio which was not working with old build. but I still have the negative point when seeking into the second chapter
Edit:
now it seems to work even when seeking into the second chapter no bad timeline position.

scottydulton:
@chamchenko  @glennguy 
I tested games starting pre midnight gmt and post midnight gmt and both are working correctly from live and from beginning seeking to any timeline or chapter 😀😀

glennguy:
Hack removed from PR, it is now implemented in a much more proper way in #685.

src/main.cpp
```diff
@@ -2287,11 +2293,12 @@ bool Session::InitializeDRM()
...
-    {
-      kodi::Log(ADDON_LOG_ERROR, "OpenDRMSystem failed");
-      return false;
-    }
+    if (!decrypter_->HasCdmSession())
+      if (!decrypter_->OpenDRMSystem(license_key_.c_str(), server_certificate_, drmConfig_))
+      {
+        kodi::Log(ADDON_LOG_ERROR, "OpenDRMSystem failed");
+        return false;
+      }
```
phunkyfish: If you are using an enclosed if statement always use brackets for the outer one. Then you can't run foul of a dangling else when extending the code.

phunkyfish (changes requested):

src/main.cpp
glennguy: No worries, sorted!

phunkyfish (approved):
LGTM

matthuisman (approved):
looks good. also run-time tested and found no regressions with other HLS streams

PR #668, opened 2021-05-02
[Dash] Fix Multiple SegmentTemplate Issues

matthuisman:
1) 
Fixes issue of wrong url calculated for init and media for the 2nd SegmentTemplate found.
 This is due to it using dash->current_representation_->url_ as the baseurl agument for ParseSegmentTemplate.
 This should actually be using dash->current_adaptationset_->base_url_

 dash->current_representation_->url_ is getting set to the previous initialization below when it starts on the Reprensentation
 https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/parser/DASHTree.cpp#L772
 So further on its using the previous initialisation  url as a baseurl for the new initialisation url

2) 
Fixes if startnumber provided in adaptionset SegmentTemplate but not provided in representation SegmentTemplate .
   Currently it will always use 1 if not found. With this PR, it will use the existing startnumber

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -90,9 +90,9 @@ static uint8_t GetChannels(const char** attr)
 static unsigned int ParseSegmentTemplate(const char** attr,
                                          std::string baseURL,
                                          std::string baseDomain,
-                                         DASHTree::SegmentTemplate& tpl)
+                                         DASHTree::SegmentTemplate& tpl,
+                                         unsigned int startNumber)
 {
-  unsigned int startNumber(1);
   for (; *attr;)
   {
     if (strcmp((const char*)*attr, "timescale") == 0)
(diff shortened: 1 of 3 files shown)


glennguy:
Looks really good, agreed on both issues.

`current_representation_->url_` is meant to be used for init segment when using SegmentTemplate (but baseurl for segment lists and ranged files - see [here](https://github.com/xbmc/inputstream.adaptive/blob/11412ab484d1f4efd62de68d1f3e7a7d09c488a9/src/common/AdaptiveStream.cpp#L306)) so very good fixing that.

 `startNumber_` is set to 1 by default at the Period level then propagates its way through AdaptationSet and Representation if not overridden by a template, so no need for setting a default of 1 in parsing the template :)

glennguy (approved):

Vikassm73:
Need similar fixing for Leia also

matthuisman:
Don't think Leia is getting any further IA updates sorry

PR #675, opened 2021-05-10
Refactor / Simplify effective (redirect) logic

### LINKED ISSUE #617: HLS master playlist redirect edge case

### LINKED ISSUE #672: 6Play via plugin CatchUp TV & More KO
Bonjour,

Un problème de DRM empêche le lancement des émissions en Replay du groupe 6Play via le plugin CatchUp TV ans More.

Le version utilisé est Kodi 19.0 Matrix.

Voici le lien vers le ticket avec des logs disponibles :
[Github Issues](https://github.com/Catch-up-TV-and-More/plugin.video.catchuptvandmore/issues/510) 

Merci pour votre aide,

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/617
fixes: https://github.com/xbmc/inputstream.adaptive/issues/672

Commit 1 Changes
- Remove media_renewal_url & media_renewal_time as no longer work with commit 2 changes due to relying on replacing domain strings (unreliable and often doesnt work). Peak himself said this was a workaround (hack) https://github.com/xbmc/inputstream.adaptive/issues/226#issuecomment-494290363. 
Correct approach would be manifest_update_param="full" or a Python proxy if that won't work.
Only add-on I could find using this was Ziggo IPTV. That author has now changed to using a proxy https://github.com/xbmc/inputstream.adaptive/pull/620#issuecomment-809036063
- rename "mpd" to "manifest" in main due to supporting more than mpd - also update log messages
- removed manifest_parameter_ as not used

Commit 2 Changes
- create a valid url from filename for tests that pass in empty url
- updated all .find("", 0); to .find(""); as 0 is default parameter
- merge SetEffectiveURL code into PreparePaths and remove SetEffectiveURL function as not used
- move base_url appending into BuildDownloadUrl. This removes a lot of duplicate code from elsewhere.
- update download function to call PreparePaths so the final redirected URL is correctly used
- add new PrepareManifestUrl function that is called before downloading manifest to correctly update_parameter_ and remove it from the url
- allow update_parameter_ to work correctly with redirect by moving ? or & check into RefreshLiveSegments
- make update_parameter_pos_ into a local variable as only needed in RefreshLiveSegments
- HLS: remove url parameter from processManifest as not used
- HLS: use BuildDownloadUrl to remove duplicate code
- HLS: when playing a substream playlist directly, use manifest_url_ as this will be the effective url 
This stops doing the redirect again when IA requests it a 2nd time
(could also use effective_url_ instead of manifest_url_ if that makes more sense?)
With some bigger changes, we could probably stop the 2nd request - but that's out of this PR scope and IMO be quite hacky
as it would skip the normal HLS flow
- HLS: use the effective_url_ as basis for calculating substreams
- update tests - enable HLS skipped test

**Test builds here (Matrix only):
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-675/36/artifacts**

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -56,7 +56,6 @@ namespace adaptive
     , updateInterval_(~0)
     , updateThread_(nullptr)
     , lastUpdated_(std::chrono::system_clock::now())
-    , lastMediaRenewal_(std::chrono::system_clock::now())
   {
   }
 
@@ -319,8 +318,6 @@ namespace adaptive
   {
     size_t paramPos = url.find('?');
     base_url_ = (paramPos == std::string::npos) ? url : url.substr(0, paramPos);
-    if (paramPos != std::string::npos)
-      manifest_parameter_= url.substr(paramPos);
 
     paramPos = base_url_.find_last_of('/', base_url_.length());
     if (paramPos == std::string::npos)
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -40,7 +40,6 @@ namespace adaptive
   AdaptiveTree::AdaptiveTree()
     : current_period_(nullptr)
     , next_period_(nullptr)
-    , update_parameter_pos_(std::string::npos)
     , parser_(0)
     , currentNode_(0)
     , segcount_(0)
@@ -170,7 +169,7 @@ namespace adaptive
   }
 
   void AdaptiveTree::OnDataArrived(unsigned int segNum, uint16_t psshSet, uint8_t iv[16], const uint8_t *src, uint8_t *dst, size_t dstOffset, size_t dataSize)
-  { 
+  {
     memcpy(dst + dstOffset, src, dataSize);
   }
 
@@ -314,15 +313,17 @@ namespace adaptive
           ++br;
   }
 
-  bool AdaptiveTree::PreparePaths(const std::string &url, const std::string &manifestUpdateParam)
+  bool AdaptiveTree::PreparePaths(const std::string &url)
   {
+    manifest_url_ = url;
+
     size_t paramPos = url.find('?');
     base_url_ = (paramPos == std::string::npos) ? url : url.substr(0, paramPos);
 
     paramPos = base_url_.find_last_of('/', base_url_.length());
     if (paramPos == std::string::npos)
     {
-      Log(LOGLEVEL_ERROR, "Invalid mpdURL: / expected (%s)", manifest_url_.c_str());
+      Log(LOGLEVEL_ERROR, "Invalid url: / expected (%s)", url.c_str());
       return false;
     }
     base_url_.resize(paramPos + 1);
@@ -338,6 +339,11 @@ namespace adaptive
     else
       base_domain_.clear();
 
+    return true;
+  }
+
+  void AdaptiveTree::PrepareManifestUrl(const std::string &url, const std::string &manifestUpdateParam)
+  {
     manifest_url_ = url;
 
     if (manifestUpdateParam.empty())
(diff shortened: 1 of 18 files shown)


matthuisman:
~~this pr causes issues with:
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_subtitles/hls_subtitles.strm
pauses at 6s and can't seek~~

Fix1.
I had been allowing the sub HLS playlist to updateParams when downloading...
but this resulted in the subtitles updating the base_url_ etc, and then all the audio / videos were wrong.

So keep how it was where a sub playlist keeps track of it's own base_url

~~oh, but that means a redirected sub-playlist will use old url as a base_url~~
Fix2:
For the above issue, setting effective_url_ in Download and then using that to create the base_url for the sub-playlist

matthuisman:
@glennguy 
I'm tempted to break this into 3x PR's
1) Rename mpd to manifest and update the logging and remove unused manifest_parameter_  (clean up PR)
2) Remove media_renewal_url & media_renewal_time (remove hacky method thats not compatible with #3)
3) Redirect redo

matthuisman:
@glennguy @anyone_else
Ready for review when you can :)

src/common/AdaptiveTree.cpp
```diff
@@ -319,18 +318,16 @@ namespace adaptive
...
     if (paramPos == std::string::npos)
     {
-      Log(LOGLEVEL_ERROR, "Invalid mpdURL: / expected (%s)", manifest_url_.c_str());
+      Log(LOGLEVEL_ERROR, "Invalid url: / expected (%s)", url.c_str());
       return false;
     }
     base_url_.resize(paramPos + 1);
 
-    paramPos = base_url_.find("://");
+    paramPos = base_url_.find("://", 0);
```
glennguy: You mentioned this change is for consistency with how it is used elsewhere. I think it may be better to go the other way though, the default argument for pos is 0 so specifying it seems redundant.

glennguy (changes requested):

src/common/AdaptiveTree.cpp
phunkyfish: I agree @glennguy, if anything the other uses should have the `, 0` removed. Best to use the default usage when ever possible.

matthuisman: Sounds good to me. I'll change the others as well and update PR :)

matthuisman: Done :)

src/test/TestHelper.cpp
```diff
@@ -22,15 +22,23 @@ void SetFileName(std::string& file, std::string name)
...
+                                      bool isManifest)
 {
   FILE* f = fopen(testHelper::testFile.c_str(), "rb");
   if (!f)
     return false;
 
-  if (scanEffectiveURL && !testHelper::effectiveUrl.empty())
-    SetEffectiveURL(testHelper::effectiveUrl);
+  effective_url_ = url;
+  if (!testHelper::effectiveUrl.empty())
```
glennguy, line 32: Rather than assign twice here a if/else would be clearer

glennguy (changes requested):
Fantastic work with this clean up, just one more minor change :)

glennguy:
Btw @phunkyfish the failing clang builds everywhere are likely due to this: https://github.com/actions/virtual-environments/issues/3376

thanks @matthuisman for the research on that

src/common/AdaptiveTree.h
```diff
@@ -412,8 +412,7 @@ class ATTRIBUTE_HIDDEN AdaptiveTree
   }*current_period_, *next_period_;
 
   std::vector<Period*> periods_;
-  std::string manifest_url_, base_url_, effective_url_, base_domain_, effective_domain_,
-      update_parameter_;
+  std::string manifest_url_, base_url_, effective_url_, base_domain_, update_parameter_;
```
phunkyfish: Just a comment on this. Creating multiple variables like this is usually best avoided. The reasoning is probably something like, "Oh but those variables are related so I create them on the same line". 

Doing it this way makes it hard to add a comment and someone new to the code doesn't know where they should add their new variable, as a type like this one has several places where multiple vars are created. If multiple groups of separate types are needed then use a function to denote their relationship instead.

BTW, I'm not familiar with the code, maybe there is a very good reason for this.

matthuisman: I was just keeping with what was already there. Should I change?

UPDATE:
I did change :)
May as well fix stuff like this as we go!
It also means if a variable is removed - it's easier to tell in the diff in the future

src/test/TestHelper.cpp
matthuisman: My reasoning for this was to keep it as close as the real function. This line is where the effective url would be retrieved from curl. 

We need keep the real and mocked function as close as possible to keep the tests accurate?

Wait. Ignore me. Lol. Your right. All these 3 lines are for simulating that one line. Il fix :)

UPDATE: Done

src/common/AdaptiveTree.h
phunkyfish: I would just define one on each line.

matthuisman: Done :)

matthuisman:
just found something with the $START_NUMBER$ stuff (looks like Youtube uses it)
will need to rework a bit. please hold :)

phunkyfish:
The changes look ok to me. I would suggest in future to separate the refactoring into its own commit before the other changes. Largely I will ignore refactoring commits as they don't change anything. Then the core logic changes are easier to review.

Let me know when you have solved the issue you found and I can do a full review.

glennguy:
@iranl are you still using media renewal? I found some old issues where you made a PR with changes to this section of code.
The view is that this method of retrying after 403 and requesting information from the add-on and getting a listitem in return is pretty hacky and an add-on side proxy is preferred.

matthuisman:
**ok guys, this one is ready for review :)**

Fixing the Youtube issue actually made it cleaner in the end.
No more needing to pass manifestUpdateParam to download which meant yucky empty "" parameters in HlsTree when downloading a sub-playlist.

Basically there is some code that needs to run before the manifest request (PrepareManifestUrl) to pull out any template tags and then code that needs to run after the request (PreparePaths) so the final url can be used. 

Flow is:
1) url comes into the manifest open function
2) PrepareManifestUrl is called. This strips out any template tags (just $START_NUMBER$ currently)
    then it sets the url without that to manifest_url_. This is the url we want to download
3) open function passes manifest_url_ to download()
4) after the curl request, we call PreparePaths with the final redirect url
5) this sets up all the base_url correctly off the final url. 
   It also updates the manifest_url_ to the final url

Now as we don't know the final manifest_url_ in PrepareManifestUrl, I moved the code that checks for & and ? of update_parameter_ into the refresh segements code so correctly works with any redirected url. 
I then noted without this code, we only use update_parameter_pos inside that function. 
So moved that to a local variable.

I've tested on Youtube Live, Youtube VOD, Youtube purchased movie.
All my tests from here: https://github.com/matthuisman/ia_tests
and all my add-ons including iptv add-ons that use a heap of different HLS live streams

Wasnt sure if PrepareManifestUrl  should return the new url, modify one passed in or use a class variable.
In the end I just used the class variable manifest_url_

phunkyfish (approved):
Looks good to me.

glennguy (approved):
Love it!

PR #676, opened 2021-05-10
[HLS] Don't reset pts to 0 on new period

glennguy:
## Description
*This is my third attempt to solve this issue, tries 1 and 2 were not getting at the root of the problem and involved:*
* *In OpenStream(): Forcing the fmp4 sample reader to start and download the first segment of the period to set a correct PTS diff/offset*
* *Using a class bool to track when we seek across chapters to conditionally subtract the start time of the new chapter*

Finally after many hours I present a one liner 😄 

Upon creation of a new period in HLS manifests (#EXT-X-DISCONTINUITY tag encountered) currently the pts is set back to 0, this PR keeps the pts value so that the first segment of the new period picks up where left off.

## Motivation and context

So far resetting pts to 0 hasn't caused issues when playing from one chapter to the next, but seeking into another chapter (using seek steps i.e left and right buttons, not up/down) causes issues with fmp4 streams. In addition to manually seeking the cross chapter seek is called when resuming in any chapter > 1. TS seems to be unaffected due to video/audio in the one file.

When seeking across chapters the `SeekTime` function in main.cpp is calculating incorrect times to instruct the stream readers to advance to, in particular here:
https://github.com/xbmc/inputstream.adaptive/blob/f8bb7435ff72281e4ec7affbd341173cb08e1e32/src/main.cpp#L2942
 - the absolute PTS offset is always being set to 0 resulting in the following scenario:
```
 CurlFile::Open(0x2c1f272e870) https://raw.githubusercontent.com/glennguy/wvtest/mp-live/h264_720p/7.m4s
 2021-05-04 21:07:15.082 T:24612   DEBUG <general>: CVideoPlayerAudio - CDVDMsg::GENERAL_PAUSE: 0
 2021-05-04 21:07:15.083 T:24612   DEBUG <general>: CDVDAudio::Pause - pausing audio stream
 2021-05-04 21:07:15.511 T:23880    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(19.2) for Stream:65537 continues at 18.0 (PTS: 36080000)
 2021-05-04 21:07:15.511 T:11592   DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'connection: keep-alive'
 2021-05-04 21:07:15.511 T:11592   DEBUG <general>: CurlFile::Open(0x2c1f272f8b0) https://raw.githubusercontent.com/glennguy/wvtest/mp-live/audio/10.m4s
 2021-05-04 21:07:15.866 T:23880    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(18.0) for Stream:65538 continues at 35.9 (PTS: 54016000)
```
You can see the video and audio segment numbers are out of alignment. The audio stream has seeked well past where it should have, and the user is left without audio until the video stream reaches the audio streams time. In many cases this could be past the end of the chapter or stream and audio will never recover by itself within the chapter.

## How has this been tested?

Tested on DASH (multi period) and HLS (multi period, multi period live) streams in Windows 10.
* Seeking forwards into new chapter
* Seeking backwards into new chapter
* Seeking within the current chapter, chapter == 1 and chapter > 1

Audio streams will continue when seeking forward through chapters.
In particular multi period live streams (sports) where users will want to start the stream at the live edge. Due to a limitation in the inputstream interface, inputstream adaptive must always start in the first period. The workaround in add-ons is to set a resume time, which relies on seeking across a chapter border.

Fixes issue mentioned here: https://github.com/xbmc/inputstream.adaptive/pull/663#issuecomment-831688263

## Types of change

## Checklist:

- [x] I have added tests to cover my change


matthuisman:
looks good!
got to love the 1-line fixes :)

matthuisman (approved):
looks good. I have run-time tested with my various test streams and add-ons (Disney+ etc)

dagwieers:
That may be related to reports we got (about audio being lost during/after commercials).

phunkyfish (approved):
LGTM

matthuisman:
> That may be related to reports we got (about audio being lost during/after commercials).

I've also heard similar reports with 19 and STIRR for example that uses x-discontinuity.
Each ad / commercial is a period - so I can see how the original setting to 0 could impact that quite a lot.
If they are on Leia - it is probably be because of x-discontinuity as well. 
Only 19 version of IA has glenn's x-discontinuity code.

Issue #497, opened 2020-08-13
IA does not expose playback errors using onPlayBackError()


dagwieers:
This is an important problem for many add-ons, if playing a stream fails by IA for whatever reason (authentication issue, license issue, any HTTP issue, format error or Widevine library issue) this is not reported back to the add-on. As a result an add-on cannot report this failure properly to the user, or implement a work-around (in case this is a known/expected issue). add-ons/plugin.video.vrt.nu#727

ISH (InputStream Helper) could use specific license or Widevine issues as an event for e.g. updating the Widevine library, and therefore would not need to check for newer Widevine libraries on every call (or after some TTL has expired). This would be a much better user experience. emilsvennesson/script.module.inputstreamhelper#69)

CastagnaIT:
i have investigated on this problem, because i have similar problem on netflix

the results are that ISA provide the right value to Kodi when going in error then the problem is not in ISA,
after check che Kodi Player core i see that completely is missing the error notification event

i built a Kodi W64 build, from a first test with netflix Monitor seem now works, please test with your VRT addon:
https://www.dropbox.com/s/kvdr8j7cgkrhbma/Kodi19_Win64_Patch_PlayerEventError.zip?dl=0

if go well i will try to do a PR
let me know

dagwieers:
@CastagnaIT This is an important one indeed, it would be nice to see this fixed in Matrix. We may have to discuss the best interface for add-ons in this case.
- What information do we return, in what format?
- Do we have different callback handlers for different types of errors or not?

It would be nice to look at all the use-cases to ensure that our design is the most convenient for add-on developers. Maybe I am a bit over-concerned here, but I am very sensitive to consider all use-cases and at least try to have the best design from the very start.

CastagnaIT:
> We may have to discuss the best interface for add-ons in this case.
> What information do we return, in what format?
> Do we have different callback handlers for different types of errors or not?

what are you referring to?
I don't think i understand well what you're trying to explain

dagwieers:
You have different kinds of errors, and they have different information they can return, and we should have a good idea of the range of errors and the type of information that is returned, and how add-ons would be interacting with it. Look at the description of the issue and two specific use-cases we have. Maybe there are  more?

CastagnaIT:
Ah your idea is would return a code and/or description of the error in the onPlayBackError() notification
like
```
onPlayBackError(data)
code = data['err_code']
```

could be useful, but not easy to implement,
[...]

PR #679, opened 2021-05-10
cdm_adapter: log error if LoadNativeLibrary fails

kszaq:
In case loading libwidevine fails, we should print the reason to log. This simple patch allows to print the error to the debug log.

Related to https://github.com/xbmc/inputstream.adaptive/issues/678

### DIFF
--- a/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
+++ b/wvdecrypter/cdm/media/cdm/cdm_adapter.cc
@@ -160,7 +160,11 @@ void CdmAdapter::Initialize()
   library_ = base::LoadNativeLibrary(cdm_path_, &error);
 
   if (!library_)
+  {
+    std::string log_error = "CDM LoadNativeLibrary error: " + error.ToString();
+    client_->CDMLog(log_error.c_str());
     return;
+  }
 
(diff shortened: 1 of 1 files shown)


phunkyfish (commented):
Appears perfectly sane to me. @glennguy?

matthuisman:
out of interest, what is the error you get with the latest troublesom Widevine blob?

kszaq:
`cannot allocate memory in static TLS block`
No idea how to fix.

matthuisman:
quick google I get this:
https://github.com/opencv/opencv/issues/14884
aarch64 issue...

There is no way we are actually getting a new AARCH64 widevine blob are we?
As in - they have finally released one...

Are the Chrome HW  they using ARM7 or were they AARCH64 and using ARM7 widevine.
They did just recently do an ARM64 for Mac - so.... maybe ARM64 for Chrome finally?

glennguy (approved):

dagwieers:
Any chance we can return this via the player as an onPlaybackError event as well?

kszaq:
I'm afraid no until https://github.com/xbmc/inputstream.adaptive/issues/497

PR #680, opened 2021-05-12
Increased 'extra_data' buffer to avoid a crash with streams containing large VPS/SPS/PPS

gizmocuz:
Took me a while to find this one... Testing with a UHD HEVC stream with the following NAL sizes was causing a crash

VPS 135
SPS 148
PPS 117

This caused the 'extra_data_size' to be 400 bytes instead of the maximum of 256

I also added a extra check in ES_hevc.cpp to avoid this from happening in the feature and added a debug log in case the buffer size is to small in the feature

### DIFF
--- a/lib/mpegts/ES_hevc.cpp
+++ b/lib/mpegts/ES_hevc.cpp
@@ -183,9 +183,16 @@ void ES_hevc::Parse_HEVC(int buf_ptr, unsigned int NumBytesInNalUnit, bool &comp
     case NAL_VPS_NUT:
       if (m_NeedVPS)
       {
-        memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
-        stream_info.extra_data_size += NumBytesInNalUnit;
-        m_NeedVPS = false;
+        if (stream_info.extra_data_size + NumBytesInNalUnit <= sizeof(stream_info.extra_data))
+        {
+          memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
+          stream_info.extra_data_size += NumBytesInNalUnit;
+          m_NeedVPS = false;
+        }
+        else
+        {
+          DBG(DEMUX_DBG_INFO, "HEVC fixme: stream_info.extra_data too small! %i\n", stream_info.extra_data_size + NumBytesInNalUnit);
+        }
       }
        break;
 
@@ -200,9 +207,16 @@ void ES_hevc::Parse_HEVC(int buf_ptr, unsigned int NumBytesInNalUnit, bool &comp
       Parse_SPS(buf, NumBytesInNalUnit, hdr);
       if (m_NeedSPS)
       {
-        memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
-        stream_info.extra_data_size += NumBytesInNalUnit;
-        m_NeedSPS = false;
+        if (stream_info.extra_data_size + NumBytesInNalUnit <= sizeof(stream_info.extra_data))
+        {
+          memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
+          stream_info.extra_data_size += NumBytesInNalUnit;
+          m_NeedSPS = false;
+        }
+        else
+        {
+          DBG(DEMUX_DBG_INFO, "HEVC fixme: stream_info.extra_data too small! %i\n", stream_info.extra_data_size + NumBytesInNalUnit);
+        }
       }
       break;
     }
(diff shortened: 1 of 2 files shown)


glennguy (approved):

glennguy:
Thanks @gizmocuz, I'm not really across this part of the project but it sounds fine to me.

lib/mpegts/ES_hevc.cpp
```diff
@@ -183,9 +183,15 @@ void ES_hevc::Parse_HEVC(int buf_ptr, unsigned int NumBytesInNalUnit, bool &comp
     case NAL_VPS_NUT:
       if (m_NeedVPS)
       {
-        memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
-        stream_info.extra_data_size += NumBytesInNalUnit;
-        m_NeedVPS = false;
+        if (stream_info.extra_data_size + NumBytesInNalUnit < sizeof(stream_info.extra_data))
```
arnova: I think this should be "<=", same for the ones below.

arnova (changes requested):

lib/mpegts/ES_hevc.cpp
```diff
@@ -183,9 +183,15 @@ void ES_hevc::Parse_HEVC(int buf_ptr, unsigned int NumBytesInNalUnit, bool &comp
...
+        {
+          memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
+          stream_info.extra_data_size += NumBytesInNalUnit;
+          m_NeedVPS = false;
+        }
+        else {
+          DBG(DEMUX_DBG_INFO, "HEVC fixme: stream_info.extra_data to small! %i\n", stream_info.extra_data_size + NumBytesInNalUnit);
```
arnova: "to" => "too", same for the ones below.

matthuisman (approved):
looks good once @arnova requested changes made

matthuisman:
Could you do the "too" for the log message? And also squash all into a single commit?
Please :))

gizmocuz:
Yeah sorry about this, this is a coding habit of mine... always on the safe side ;)
These days I use a std::vector<uint8_t> or std::string for small undefined buffers that do not change/increase often

@matthuisman , I have to do some research how to squash all into a single commit... I know you can 'Squash and Merge' here ... Let me check if I can do this remotely without recreating this PR

matthuisman:
You can just soft revert your local commit. Do changes. Then git add . again. Then git commit --amend. Then force push git push -f

git reset --soft HEAD~2
git add .
git commit --amend 
git push -f

Or look into rebase. But I usually use above as makes more sense in my head. Haha

Awesome, see you got it sorted :)

gizmocuz:
Hope I did it right... I don't see much difference here. But feel free to accept this PR with Github's squash and merge option

image

matthuisman:
Yup. All in one commit now.
I forgot about that github button :)

phunkyfish:
We actually generally avoid squash and commit. Much better to have commits and their descriptions appear as the author intended.

matthuisman:
Makes sense. It was ok to ask them to be squashed though as two of them where just small fix ups? Wouldn't want commits like "fix typo" being merged?

phunkyfish:
> Makes sense. It was ok to ask them to be squashed though as two of them where just small fix ups? Wouldn't want commits like "fix typo" being merged?

Yes, of course. We always want the author to prepare the commits correctly and squash in any fixups. We pretty much just want the PR commits to be the same once merged.

gizmocuz:
@phunkyfish , I am new to squashing, but when you squash via Github, isn't that the same as the above instructions ? Making it one commit ?

phunkyfish:
> @phunkyfish , I am new to squashing, but when you squash via Github, isn't that the same as the above instructions ? Making it one commit ?

In this case sure but if you had the order incorrect then it could look very different in the upstream repo. Also if you intended to have more than one commit then you lose some of the commit descriptions. Hence we never do this.

We want to always make sure that the commits are named and ordered the same in both the PR and the upstream repo. This makes it easier to find the reasoning behind changes and avoids any large PRs from being squashed in error. When this happens it can be impossible to find which change causes a regression which wouldn't have occurred if we don't squash and merge.

lib/mpegts/ES_hevc.cpp
```diff
@@ -183,9 +183,15 @@ void ES_hevc::Parse_HEVC(int buf_ptr, unsigned int NumBytesInNalUnit, bool &comp
...
-        memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
-        stream_info.extra_data_size += NumBytesInNalUnit;
-        m_NeedVPS = false;
+        if (stream_info.extra_data_size + NumBytesInNalUnit <= sizeof(stream_info.extra_data))
+        {
+          memcpy(stream_info.extra_data + stream_info.extra_data_size, es_buf + (buf_ptr - 4), NumBytesInNalUnit);
+          stream_info.extra_data_size += NumBytesInNalUnit;
+          m_NeedVPS = false;
+        }
+        else {
```
arnova: { on a new line please, same for the ones below.

matthuisman:
@phunkyfish 
Is there anyway to get Jenkins to do the clang formatting check patch thing it does on main repo?

phunkyfish:
> @phunkyfish
> Is there anyway to get Jenkins to do the clang formatting check patch thing it does on main repo?

Probably a better conversation to have on kodi slack. I’m sure it is but would be down to addon maintainers if they want it or not.

matthuisman:
I just noticed a user had requested formatting changes above, so was more wondering what should be enforced / encouraged? 

As this is now in xbmc github, im sort of viewing it as an extension of kodi?

phunkyfish:
No, addon formatting and style is down the maintainers of an addon. However in this case they are team members. If you would like to abide by the same format just copy the .clang_format file from xbmc to here. If you want to can run it on the source you can to see how much would change. But you would need to install clang format of course.

matthuisman:
already is using a .clang-format:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/.clang-format
which looks the same as xbmc's.

Just wondering if the team members would like it enforced?
If yes, then could we see if Jenkins could warn us if it's not and provide the patch?

lib/mpegts/ES_hevc.cpp
gizmocuz: Done. I also did a quick scan in the project, you might want to have a look in src/helpers.cpp

matthuisman:
The check error isnt related to your code. Its something else that just showed up today

phunkyfish:
> already is using a .clang-format:
> https://github.com/xbmc/inputstream.adaptive/blob/Matrix/.clang-format
> which looks the same as xbmc's.
> 
> Just wondering if the team members would like it enforced?
> If yes, then could we see if Jenkins could warn us if it's not and provide the patch?

I will ask.

phunkyfish (approved):
LGTM

arnova (approved):
LGTM

Issue #667, opened 2021-04-30
No audiostream 2.6.14.1 addon VTM GO


kengithub9:
I have problems (no audio) since I installed version 2.6.14.1 in CoreElec. I can see a picture. 
With version 2.6.7.3. everything worked.
In addon VTM GO.

What I also still have is that just after the start he gives a black image, also during the commercials, 
I noticed that it is due to the framerate changes. 

How can I restore the inputstream version, the zip you find on github cannot be installed?

System

     Addon version: 1.25
     Kodi version: 19.0
     Input stream adaptive version: 2.6.14.1
     Operating System: CoreElec 19.1rc4
     Special Hardware: Beelink GT-King

glennguy:
Zip can't be found here, you'd need to ask at CoreELEC.

What might be better is if you could provide a full debug log (no component logging pls) and we can fix for the future?

kengithub9:
How can I make such a debug log?

freynder:
@kengithub9 : You need to activate debug log in Kodi settings. 

@glennguy : I also experience the same problem, attached you can find the log.
[kodi.log.gz](https://github.com/xbmc/inputstream.adaptive/files/6407720/kodi.log.gz)

The first advertisement plays, then the screen turns black and logs shows lots of timeout waiting for buffer error messages:
2021-04-30 21:44:43.158 T:5626  WARNING <general>: OutputPicture - timeout waiting for buffer
2021-04-30 21:44:43.209 T:5626    DEBUG <general>: CVideoPlayerVideo::OutputPicture - ttd:-504ms pts:20.559 Clock:21.064 Level:-1
after a while it plays the next section, but without audio.

Also attaching the mpd retrieved from https://dcs-vod.apis.anvato.net/vod/p/session/manifest.mpd?i=i176422922-nf77c271a-30f2-4a84-a9c1-5d06249d0994
[manifest.mpd.gz](https://github.com/xbmc/inputstream.adaptive/files/6407763/manifest.mpd.gz)

Also maybe interesting: when trying to play on [Dash javascript player](https://reference.dashif.org/dash.js/nightly/samples/dash-if-reference-player/index.html), it fails after the advertisement with an error. Not sure if it is related:
[...]

Issue #648, opened 2021-04-14
Audio Stream doesn't change in Live channels with multiple languages


nikhilk0:
While watching live channels with multiple language audio streams in Discovery+ addon, when toggling audio from one language to another (e.g. English to regional language), instead to switching to that language the audio stops playing and there is no audio.
Current workaround is to toggle the audio stream to your language preference, stop the stream and play it the stream again as kodi remembers your last audio stream.
This issue is only present with live channels. Discovery+ episodes with multiple language audio play fine.
I have mentioned it on discovery+'s repo, but the issue seems to be with ISA.

Tell me if this is legit issue and if you want any debug logs for reference. Thanks

matthuisman:
@glennguy I wonder if this is also fixed with
https://github.com/xbmc/inputstream.adaptive/pull/685

nikhilk0:
Unfortunately, the issue is still there even after this PR.

nikhilk0:
@glennguy @matthuisman I tried again with [this](https://jenkins.kodi.tv/job/xbmc/job/inputstream.adaptive/view/change-requests/job/PR-685/lastSuccessfulBuild/artifact/cmake/addons/build/zips/inputstream.adaptive+windows-x86_64/inputstream.adaptive-2.6.15.zip) build and the audio stream does change. But there is a slight issue.  When I start playing the stream, the video is hung and audio takes 5-10 secs to start playing with video still hung. The video doesn't start playing until I switch the audio to another language.

nikhilk0:
Okay. The latest build of https://github.com/xbmc/inputstream.adaptive/pull/685 has fixed the issue. Thanks for the excellent work @glennguy.

glennguy:
@nikhilk0 my pleasure, thank you so much for testing and reporting back!

PR #685, opened 2021-05-15
[fmp4] ensure streams are started when seeking into a new chapter

### LINKED ISSUE #681: FMP4 seeking between chapters is broken - SeekTime logic failing
### Describe the bug

So, I thought that #676 would solve this problem but it seems not for all streams, live streams (HLS only at this point, DASH live multiperiod doesn't really work at all) in particular.

Unfortunately the test streams that I have built using shaka packager work very well. I believe this is due to the stream as a whole having a continuous pts time line throughout the presentation from 0 until finish. The current implementation seems to fit this scenario well.

The abridged flow when seeking into another chapter is:
* Call SeekTime()
* SeekTime determines that we're seeking into a new chapter, sets a flag and returns early
* Kodi opens the new chapter, streams are opened/setup
* During this stream setup the initialization segments are downloaded
* Inputstream.adaptive seeks to the requested time due to flag
* The below logic does not function correctly (when conditions aren't as described above) and bad things happen
Namely, the pts diff hasn't yet been set by the reader. The first stream seeks to the correct segment, but on second iteration a pts diff *is* set and we end up offset by that amount.

https://github.com/xbmc/inputstream.adaptive/blob/4c1a0e163b1da312896758294ff1e69f18a3c27a/src/main.cpp#L2939-L2960

## Expected Behavior

Stream resumes at correct point with audio and video

## Actual Behavior
Video stream resumes at the correct point, audio stream does not.

## Possible Fix

One possible fix is to emulate the behavior that .ts streams have and 'initialize' the reader:
https://github.com/xbmc/inputstream.adaptive/blob/4c1a0e163b1da312896758294ff1e69f18a3c27a/src/main.cpp#L3628
The result of this action is reader reads the first segment of the stream.
When seeking into another chapter in a multiperiod .ts stream the first segment of the new chapter is always read.

We can do the same when creating the fmp4 reader by calling `bool bStarted; stream->reader_->Start(bStarted);`
The first segment will always be read when seeking into a new chapter.

This helps us by getting the pts diff set in the sample reader so the correct value (not default of 0) is returned from each call to `reader_->GetPTSDiff()`

Advantage of this method:
* Easy to implement
* Don't have to change code in SeekTime and risk introducing bugs

Downsides:
* An extra segment for each stream is downloaded that may not be presented.
* Feels like a hack.

Ideal solution - rewrite SeekTime.
Off the top of my head this could be done my setting a flag in the reader that indicates 'have we yet set an initial pts diff?'
If this condition isn't true, then we handle and rewrite accordingly. 

@peak3d if you have any spare time your input would be much appreciated.

### To Reproduce

The following test stream gives an indication as to what happens. The first ~6 second segment has been removed from the playlists:
```
#KODIPROP:inputstream.adaptive.license_key=https://proxy.uat.widevine.com/proxy||R{SSM}|
https://raw.githubusercontent.com/glennguy/wvtest/mp-live-clear-missing-from-start/h264_master.m3u8
```
1. Save this as a .strm file and open from Kodi's file manager.
2. Immediately seek to chapter 2 or 3
3. Observe audio stream is not there for ~6 seconds
4. View log and see similar to the below, showing the audio stream has seeked to ~6 seconds after video:
```
2021-05-12 22:06:22.639 T:6708    DEBUG <general>: CurlFile::Open(0x20211065a00) https://raw.githubusercontent.com/glennguy/wvtest/mp-live-clear-missing-from-start/h264_480p/18.m4s
2021-05-12 22:06:22.639 T:19064   DEBUG <general>: CVideoPlayerAudio - CDVDMsg::GENERAL_PAUSE: 0
2021-05-12 22:06:22.639 T:19064   DEBUG <general>: CDVDAudio::Pause - pausing audio stream
2021-05-12 22:06:22.961 T:21148    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(28.9) for Stream:131073 continues at 24.0 (PTS: 99080000)
2021-05-12 22:06:22.961 T:7984    DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'connection: keep-alive'
2021-05-12 22:06:22.961 T:7984    DEBUG <general>: CurlFile::Open(0x20208c1ddb0) https://raw.githubusercontent.com/glennguy/wvtest/mp-live-clear-missing-from-start/audio/19.m4s
2021-05-12 22:06:23.099 T:6708    DEBUG <general>: AddOnLog: inputstream.adaptive: Download https://raw.githubusercontent.com/glennguy/wvtest/mp-live-clear-missing-from-start/h264_480p/18.m4s finished, avg speed: 1323446.49byte/s, current speed: 1216304.00byte/s
2021-05-12 22:06:23.269 T:21148    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(24.0) for Stream:65538 continues at 29.9 (PTS: 105002666)
```
If our starting mp4 PTS is different then we have different results. This case here is manageable however when we jump into a live stream with a time shift buffer smaller than the stream, we're unlikely to have such a nice situation to recover from.

### Debuglog

https://paste.kodi.tv/eqazojayuf.kodi

### Screenshots 

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Kodi version: 19.0

*note: Once the issue is made we require you to update it with new information or Kodi versions should that be required.

glennguy:
## Description

As documented in #681 there is a current issue when seeking across chapter boundaries, this PR addresses that issue.

fmp4 stream readers are not 'started' when a new chapter is loaded. This doesn't happen until either the first sample is read - called by DemuxRead() from Kodi, or through the SeekTime function itself during the seek process. The problem is that to function correctly, SeekTime needs to have accurate data from the stream reader, which is only populated once the stream reader has started, at this point in the function it is too late and the typical net effect is the audio stream advances to a segment far ahead of the video stream - playback then has no audio.

This change will check in SeekTime if the reader has started before the critical information is required and start it accordingly. Seeking can now take place in the same fashion whether seeking within a chapter or outside of.

I've also added some comments throughout SeekTime to help explain the inner workings for others benefit.

## Motivation and context

This change is required in conjunction with #663 to solve #580 and similar issues that have been mentioned in #667

Fixes #681

## How has this been tested?

Tested on this stream:
```
#KODIPROP:inputstream.adaptive.license_key=https://proxy.uat.widevine.com/proxy||R{SSM}|
https://raw.githubusercontent.com/glennguy/wvtest/mp-live-clear-alt-2/h264_master.m3u8
```
* No DRM
* Internal PTS of start of stream (first segment) is not 0
* 2 chapters

Also tested on NBA add-on (in conjunction with changes in PR #663):
* Widevine
* Internal PTS of start of stream (first segment) is not 0
* 2 chapters
* Live

Tested on Windows 10 Matrix 19.0

Users can seek without losing their audio stream
Users can use resume points (and add-on authors can use resume point workaround to make multi chapter live streams start in chapter > 1) and not lose audio stream

## Types of change

## Checklist:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1057,6 +1057,7 @@ public:
   virtual void AddStreamType(INPUTSTREAM_TYPE type, uint32_t sid){};
   virtual void SetStreamType(INPUTSTREAM_TYPE type, uint32_t sid){};
   virtual bool RemoveStreamType(INPUTSTREAM_TYPE type) { return true; };
+  virtual bool IsStarted() const = 0;
 };
 
 /*******************************************************
@@ -1087,6 +1088,7 @@ public:
   void AddStreamType(INPUTSTREAM_TYPE type, uint32_t sid) override{};
   void SetStreamType(INPUTSTREAM_TYPE type, uint32_t sid) override{};
   bool RemoveStreamType(INPUTSTREAM_TYPE type) override { return true; };
+  bool IsStarted() const override { return true; }
 } DummyReader;
 
 /*******************************************************
@@ -1265,6 +1267,7 @@ public:
   }
 
   bool EOS() const override { return m_eos; };
+  bool IsStarted() const override { return m_started; };
   uint64_t DTS() const override { return m_dts; };
   uint64_t PTS() const override { return m_pts; };
   AP4_UI32 GetStreamId() const override { return m_streamId; };
@@ -1591,6 +1594,7 @@ public:
       m_codecHandler = new TTMLCodecHandler(nullptr);
   }
 
+  bool IsStarted() const override { return true; };
   bool EOS() const override { return m_eos; };
   uint64_t DTS() const override { return m_pts; };
   uint64_t PTS() const override { return m_pts; };
@@ -1708,6 +1712,7 @@ public:
     return m_typeMask == 0;
   };
 
+  bool IsStarted() const override { return m_started; }
   bool EOS() const override { return m_eos; }
   uint64_t DTS() const override { return m_dts; }
   uint64_t PTS() const override { return m_pts; }
@@ -1805,6 +1810,7 @@ public:
   ADTSSampleReader(AP4_ByteStream* input, AP4_UI32 streamId)
     : ADTSReader(input), m_streamId(streamId), m_stream(dynamic_cast<AP4_DASHStream*>(input)){};
 
+  bool IsStarted() const override { return m_started; }
   bool EOS() const override { return m_eos; }
   uint64_t DTS() const override { return m_pts; }
   uint64_t PTS() const override { return m_pts; }
@@ -1890,6 +1896,7 @@ public:
   WebmSampleReader(AP4_ByteStream* input, AP4_UI32 streamId)
     : WebmReader(input), m_streamId(streamId), m_stream(dynamic_cast<AP4_DASHStream*>(input)){};
 
+  bool IsStarted() const override { return m_started; }
   bool EOS() const override { return m_eos; }
   uint64_t DTS() const override { return m_dts; }
   uint64_t PTS() const override { return m_pts; }
@@ -2862,6 +2869,25 @@ uint64_t Session::GetTimeshiftBufferStart()
     return 0ULL;
 }
 
+void Session::StartReader(
+    STREAM* stream, uint64_t seekTimeCorrected, int64_t ptsDiff, bool preceeding, bool timing)
+{
+  bool bReset = true;
+  if (timing)
+    seekTimeCorrected += stream->stream_.GetAbsolutePTSOffset();
+  else
+    seekTimeCorrected -= ptsDiff;
+  stream->stream_.seek_time(
+      static_cast<double>(seekTimeCorrected / STREAM_TIME_BASE),
+      preceeding, bReset);
+  if (bReset)
+    stream->reader_->Reset(false);
+  bool bStarted = false;
+  stream->reader_->Start(bStarted);
+  if (bStarted && (stream->reader_->GetInformation(stream->info_)))
+    changed_ = true;
+}
+
 SampleReader* Session::GetNextSample()
 {
   STREAM *res(0), *waiting(0);
@@ -2927,6 +2953,7 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
 
   seekTime -= chapterTime;
 
+  // don't try to seek past the end of the stream, leave a sensible amount so we can buffer properly
   if (adaptiveTree_->has_timeshift_buffer_)
   {
     uint64_t curTime, maxTime(0);
@@ -2940,11 +2967,19 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
     }
   }
 
+  // correct for starting segment pts value of chapter and chapter offset within program
   uint64_t seekTimeCorrected = static_cast<uint64_t>(seekTime * STREAM_TIME_BASE);
+  int64_t ptsDiff = 0;
   if (timing_stream_)
   {
+    // after seeking across chapters with fmp4 streams the reader will not have started
+    // so we start here to ensure that we have the required information to correctly
+    // seek with proper stream alignment
+    if (!timing_stream_->reader_->IsStarted())
+      StartReader(timing_stream_, seekTimeCorrected, ptsDiff, preceeding, true);
+
     seekTimeCorrected += timing_stream_->stream_.GetAbsolutePTSOffset();
-    int64_t ptsDiff = timing_stream_->reader_->GetPTSDiff();
+    ptsDiff = timing_stream_->reader_->GetPTSDiff();
     if (ptsDiff < 0 && seekTimeCorrected + ptsDiff > seekTimeCorrected)
       seekTimeCorrected = 0;
     else
@@ -2955,7 +2990,12 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
     if ((*b)->enabled && (*b)->reader_ &&
         (streamId == 0 || (*b)->info_.GetPhysicalIndex() == streamId))
     {
-      bool bReset;
+      bool bReset = true;
+      // all streams must be started before seeking to ensure cross chapter seeks
+      // will seek to the correct location/segment
+      if (!(*b)->reader_->IsStarted())
+        StartReader((*b), seekTimeCorrected, ptsDiff, preceeding, false);
+      // advance adaptiveStream to the correct segment (triggers segment download)
       if ((*b)->stream_.seek_time(
               static_cast<double>(seekTimeCorrected - (*b)->reader_->GetPTSDiff()) /
                   STREAM_TIME_BASE,
@@ -2963,6 +3003,7 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
       {
         if (bReset)
           (*b)->reader_->Reset(false);
+        // advance reader to requested time
         if (!(*b)->reader_->TimeSeek(seekTimeCorrected, preceeding))
           (*b)->reader_->Reset(true);
         else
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -2956,13 +2976,27 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
         (streamId == 0 || (*b)->info_.GetPhysicalIndex() == streamId))
     {
       bool bReset;
+      // all streams must be started before seeking to ensure cross chapter seeks
```
phunkyfish, line 2994: Not for this PR but variables called `b` and `e` are not easy to follow. Maybe upon a larger refactor some of these can be fixed.

src/main.cpp
```diff
@@ -2956,13 +2976,27 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
...
+      // will seek to the correct location/segment
+      if (!(*b)->reader_->IsStarted())
+      {
+        (*b)->stream_.seek_time(
+            static_cast<double>(seekTimeCorrected - ptsDiff) / STREAM_TIME_BASE,
+            preceeding, bReset);
+        if (bReset)
+          (*b)->reader_->Reset(false);
+        bool bStarted;
+        (*b)->reader_->Start(bStarted);
```
phunkyfish: This makes no sense and is bad code. `bStarted` could have any value as it's a basic type. It should be initialised. Plus it's copied in the call to start so is not used at all!

src/main.cpp
```diff
@@ -2940,11 +2943,28 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
...
+    {
+      bool bReset;
+      timing_stream_->stream_.seek_time(
+          static_cast<double>(seekTimeCorrected + timing_stream_->stream_.GetAbsolutePTSOffset()) /
+              STREAM_TIME_BASE,
+          preceeding, bReset);
+      if (bReset)
+        timing_stream_->reader_->Reset(false);
+      bool bStarted;
+      timing_stream_->reader_->Start(bStarted);
```
phunkyfish: This makes no sense and is bad code. bStarted could have any value as it's a basic type. It should be initialised. Plus it's copied in the call to start so is not used at all!

phunkyfish (changes requested):

src/main.cpp
glennguy: Yes definitely agreed, particularly this nightmare from here onwards: https://github.com/xbmc/inputstream.adaptive/blob/1247d4bcff218db54c3c2ec30be020913731f09d/src/parser/DASHTree.cpp#L1664

src/main.cpp
glennguy: Ok, so I will make sure to initialise bStarted. I'm using it based off what happens here:
https://github.com/xbmc/inputstream.adaptive/blob/1247d4bcff218db54c3c2ec30be020913731f09d/src/main.cpp#L2870-L2877

The `bStarted` bool parameter inside `Start` is by reference so I can't send in a temporary variable (much like `bReset` above) but on looking at the snippet above perhaps I should be checking `bStarted` afterwards and setting `changed_`. And also could save on code duplication by splitting this off as a function.
Does that sound good?

phunkyfish: Yes, if you don’t need to do anything with the value Then maybe a start function without it is better. Always ask yourself what should my code be doing with that value afterwards.

glennguy: All sorted (I think!)

src/main.cpp
```diff
@@ -2862,6 +2864,25 @@ uint64_t Session::GetTimeshiftBufferStart()
...
+  if (timing)
+    seekTimeCorrected += stream->stream_.GetAbsolutePTSOffset();
+  else
+    seekTimeCorrected -= ptsDiff;
+  stream->stream_.seek_time(
+      static_cast<double>(seekTimeCorrected / STREAM_TIME_BASE),
+      preceeding, bReset);
+  if (bReset)
+    stream->reader_->Reset(false);
+  bool bStarted(false);
```
phunkyfish: `bool bStarted = false;` is easier to understand. Especially for basic types, otherwise they can appear like constructors.

src/main.cpp
```diff
@@ -1057,6 +1057,7 @@ class ATTRIBUTE_HIDDEN SampleReader
   virtual void AddStreamType(INPUTSTREAM_TYPE type, uint32_t sid){};
   virtual void SetStreamType(INPUTSTREAM_TYPE type, uint32_t sid){};
   virtual bool RemoveStreamType(INPUTSTREAM_TYPE type) { return true; };
+  virtual bool IsStarted() const { return true; };
```
phunkyfish: Why is the default true?

phunkyfish (changes requested):

src/main.cpp
glennguy: I set it up like this so the other stream readers don't go through the `StartReader` block as there's no issue with how they work.
Having said that I can see now that name of the function isn't true to how it behaves, it should really return `m_Started`, correct? And then we either check the stream type as well i.e (pseudo code) `if (reader == fmp4 && !timing_stream_->reader_->IsStarted()) ...` or if possible even better to implement IsStarted() for the other readers.

phunkyfish: Yes, much better!

glennguy: Ok I hope we're good now.
No default implementation for `IsStarted`. 
Implementations for all sample readers.
Subtitle reader always returns true - there is no `m_started` that we keep track of in this reader.
bReset is also initialized now to `true`. There is a code path in `SeekTime` where bReset doesn't get modified so that's why I went with true.

phunkyfish (approved):
Nice

matthuisman (approved):
very nice. works great with provided test stream.
also tested with a bunch of my test streams / addons and noticed no regressions

glennguy:
@phunkyfish 
I reverted the change where the TS reader used Initialize() to set m_started on construction - this turned out to be setting m_started true at a time before it should be. Viewing wasn't affected but this was happening on the timeline during playback of the first segment: 
screenshot00002_720
Thanks to @matthuisman for finding this!
TS streams will now unnecessarily call StartReader(), effectively the AdaptiveStream will seek to the same segment twice in a row. No real harm (doesn't download the same segment twice) 

Wondering your thoughts on what the correct option is:
* Leave code as it is now with changes just made - TS stream gets asked to move the reader to the same segment twice in a row when we seek into a new chapter; or
* Make IsStarted() in TSReader always return true, ignoring m_started and disrespecting the name of the function.

phunkyfish:
It is perfectly ok to have a particular case where virtual function override returns something else. That is after all one of the benefits from using a object hierarchy. If you feel like the override is not obvious just add a comment.

matthuisman:
i've actually just found a Live Dash stream that this PR fixes that doesn't have chapters :)
It does have a single period with a start > 0 though.

[in.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/6511405/in.mpd.txt)
If I use ResumeFrom - it will have no audio until this PR is included.

**Tested your latest changes and those 2x little issues I had are now fixed :)**

Once you get this in, I can rebase my subtitle update to use the m_started to report back when the subtitle sample ready fails to download the sub and thus disable the stream.

scottydulton:
Any idea when this PR will be merged? Excellent work by all!

glennguy:
@phunkyfish sure thing, I'll make the change tonight and merge in

glennguy:
> Leave code as it is now with changes just made - TS stream gets asked to move the reader to the same segment twice in a row when we seek into a new chapter

Turns out we should go with this, and it's my fault for not realising... All of the TS streams I've been testing have included audio i.e 1 stream, every segment is video/audio muxed together so with only one stream to advance on seek we never end with 2nd/3rd etc streams seeking to the wrong location. 

As @matthuisman predicted in #648 we have the same problem because there are separate audio streams. Therefore we need to be going through the same process for TS streams to catch this case.

PR #688, opened 2021-05-16
Updated german translation

tweimer:


phunkyfish:
Hey, I'm afraid we only take translation for en_GB. Otherwise they will be overwritten by transifix. Also, we will migrating to Weblate soon so once we have we can give instruction on how to contribute.

phunkyfish:
Translations have moved to Weblate. Please follow the instructions here to sign up and contribute:

https://kodi.tv/article/switching-translations-to-weblate/

PR #689, opened 2021-05-18
[DASH] Allow merging video adaption sets

matthuisman:
[Discussion around criteria required]

Current criteria is
```
if (a->type_ == VIDEO && b->type_ == VIDEO && a->default_ == b->default_ &&
 a->language_ == b->language_ && a->startPTS_ == b->startPTS_ && a->group_ == b->group_)
```
Both video adaption sets
Both have the same default value (role == main or no role)
Both are the same language
Both have the same startPTS
Both are in same group (not sure if we need this?)

I've tried, and IA and Kodi both seem capable of even switching between reps using different content encryption

I think our main concern is that they are the same "timing" so when IA switches - the video keeps going at the same position.

Other things to consider.
- Two reps in different sets with same width / height but different codec (eg. H264 and H265)
  (we may meed better rep sorting that picks based on codec + width + height vs just bandwidth currently)
  
The 2nd commit here makes the sorting of the sets a bit better.
For one - no need to sort Lang A above Lang Z... so just leave them in the order they come.
Also, no need to sort based on type (video, audio, subtitle). Again - just leave them in order they come.
Also makes tests [index] numbers match the manifest

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -383,12 +383,10 @@ namespace adaptive
   {
     for (std::vector<Period*>::const_iterator bp(periods_.begin()), ep(periods_.end()); bp != ep; ++bp)
     {
-      std::stable_sort((*bp)->adaptationSets_.begin(), (*bp)->adaptationSets_.end(), AdaptationSet::compare);
-
-      // Merge AUDIO streams, some provider pass everythng in own Audio sets
+      // Merge VIDEO & AUDIO adaption sets
       for (std::vector<AdaptationSet*>::iterator ba((*bp)->adaptationSets_.begin()), ea((*bp)->adaptationSets_.end()); ba != ea;)
       {
-        if ((*ba)->type_ == AUDIO && ba + 1 != ea && AdaptationSet::mergeable(*ba, *(ba + 1)))
+        if (((*ba)->type_ == VIDEO || (*ba)->type_ == AUDIO) && ba + 1 != ea && AdaptationSet::mergeable(*ba, *(ba + 1)))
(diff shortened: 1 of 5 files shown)


kszaq:
I have tested this with one addon providing multiple adaptiation sets and it looks good! Now automatic stream selection works properly. Manual selection is also possible if selected in IA settings.

matthuisman:
I've also got another branch that just merges based on the switch xml element:
https://github.com/matthuisman/inputstream.adaptive/commit/a92dda06728f7893b07bff4a71010f4cf4f90a
(Needs a bit of a clean, but you should see the idea)

Not sure what approach we should take.. this or that?

@kszaq
Did your addon manifest have the switching xml element?
I'm also not sure if the adaption sets do need to share the same group. Couldn't find anything saying they must

kszaq:
Here is my mpd (pssh cut, they were all the same):
[live.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/6503643/live.mpd.txt)
It looks like it does have the switch xml element, doesn't it?

matthuisman:
Yup
`<SupplementalProperty schemeIdUri="urn:mpeg:dash:adaptation-set-switching:2016" value="1,2"/>`
So maybe we do that then. Im worried blindly merging video adaption sets could cause unknown issues. At least that element tells us those adaption sets are compatible to switch between

Also, I think IDs are only unique per group. So I can't see that element working across groups as it has no way to specify the group number and Id. So I'll keep the group check

matthuisman:
closed in favour of https://github.com/xbmc/inputstream.adaptive/pull/694

PR #694, opened 2021-05-19
[Dash] Support urn:mpeg:dash:adaptation-set-switching:2016

### LINKED ISSUE #555: [MPD] Support multiple video AdaptationSet (urn:mpeg:dash:adaptation-set-switching:2016)
Seeing a few mpds that have multiple video AdaptationSet.
eg.
```
<AdaptationSet id="9" group="2" contentType="video" mimeType="video/mp4"....>
        .............
       <SupplementalProperty schemeIdUri="urn:mpeg:dash:adaptation-set-switching:2016" value="10"/>
</AdaptationSet>
<AdaptationSet id="10" group="2" contentType="video" mimeType="video/mp4"....>
        .............
       <SupplementalProperty schemeIdUri="urn:mpeg:dash:adaptation-set-switching:2016" value="9"/>
</AdaptationSet>
```
maybe could be incorporated into new adaptive bitrate switching code by @spyderboy92

my current work around uses a proxy to move which ever adaption set has highest qualities to the top so IA uses that.

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/555

**Note: This is not implementing adaptive bitrate switching....**

Currently IA will just pick whichever video set is at the top of the file. 
With this pr, it will merge whichever sets are allowed and therefore be able to access all the qualities either manually or automatically.

When adaptive switching is implemented, I'm pretty sure we will still need this PR so all representations are available for it to choose from

1st
Support urn:mpeg:dash:adaptation-set-switching:2016
Pretty straight-forward
1) store a list of ids (switching_ids_) pulled from urn:mpeg:dash:adaptation-set-switching:2016
   this can be a single id, or multiple with comma delimiter
2) update SortTree() to also merge VIDEO sets
3) add new check to mergeable for VIDEOs with same group and both contain each others id in switching_ids_
4) add tests

2nd
Move the sorting to after merging sets
if we had merged, then less sets to sort = slightly faster

3rd
I also made the sort ignore type and language 
it doesn't affect anything and makes the tests easier to view as they come in same order as the file.
Update the few tests that expected the order to change to the new ordered as in file
= less confusion

4th
:dash: looks like a fart

**TEST BUILDS HERE:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-694/7/artifacts**

**TEST STREAMS:**
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_multi_vid_sets/mpd_multi_vid_sets_highest_first.strm
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_multi_vid_sets/mpd_multi_vid_sets_lowest_first.strm
https://raw.githubusercontent.com/matthuisman/ia_tests/master/mpd_multi_vid_sets/mpd_multi_vid_sets_different.strm

Only thing I wasn't sure about is:
https://github.com/xbmc/inputstream.adaptive/pull/694/files#diff-f5e439696055d30c282716f0d0fc54ee6612edb6e4f29ef91b4c3a00cd426679R618
that split() helper functions returns a new vector array. 
is it ok to assign that directly to our std::vector<std::string> switching_ids_ ?

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -599,6 +599,25 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
             attr += 2;
           }
         }
+        else if (strcmp(el, "SupplementalProperty") == 0)
+        {
+          const char *schemeIdUri(0), *value(0);
+
+          for (; *attr;)
+          {
+            if (strcmp((const char*)*attr, "schemeIdUri") == 0)
+              schemeIdUri = (const char*)*(attr + 1);
+            else if (strcmp((const char*)*attr, "value") == 0)
+              value = (const char*)*(attr + 1);
+            attr += 2;
+          }
+
+          if (schemeIdUri && value)
+          {
+            if (strcmp(schemeIdUri, "urn:mpeg:dash:adaptation-set-switching:2016") == 0)
+              dash->current_adaptationset_->switching_ids_ = split(value, ',');
+          }
+        }
         else if (dash->currentNode_ & MPDNODE_BASEURL)
         {
         }
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -384,12 +384,10 @@ namespace adaptive
   {
     for (std::vector<Period*>::const_iterator bp(periods_.begin()), ep(periods_.end()); bp != ep; ++bp)
     {
-      std::stable_sort((*bp)->adaptationSets_.begin(), (*bp)->adaptationSets_.end(), AdaptationSet::compare);
-
-      // Merge AUDIO streams, some provider pass everythng in own Audio sets
+      // Merge VIDEO & AUDIO adaption sets
       for (std::vector<AdaptationSet*>::iterator ba((*bp)->adaptationSets_.begin()), ea((*bp)->adaptationSets_.end()); ba != ea;)
       {
-        if ((*ba)->type_ == AUDIO && ba + 1 != ea && AdaptationSet::mergeable(*ba, *(ba + 1)))
+        if (((*ba)->type_ == AUDIO || (*ba)->type_ == VIDEO) && ba + 1 != ea && AdaptationSet::mergeable(*ba, *(ba + 1)))
         {
           for (size_t i(1); i < (*bp)->psshSets_.size(); ++i)
             if ((*bp)->psshSets_[i].adaptation_set_ == *ba)
@@ -403,6 +401,8 @@ namespace adaptive
         else
           ++ba;
       }
+
+      std::stable_sort((*bp)->adaptationSets_.begin(), (*bp)->adaptationSets_.end(), AdaptationSet::compare);
 
       for (std::vector<AdaptationSet*>::const_iterator ba((*bp)->adaptationSets_.begin()), ea((*bp)->adaptationSets_.end()); ba != ea; ++ba)
       {
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -302,6 +302,7 @@ public:
     std::string codecs_;
     std::string audio_track_id_;
     std::string name_;
+    std::vector<std::string> switching_ids_;
     std::vector<Representation*> representations_;
     Representation* best_rep_;
     Representation* min_rep_;
@@ -313,12 +314,10 @@ public:
       return *segment_durations_[pos];
     };
 
-    static bool compare(const AdaptationSet* a, const AdaptationSet *b)
+    static bool compare(const AdaptationSet* a, const AdaptationSet* b)
     {
       if (a->type_ != b->type_)
-        return a->type_ < b->type_;
-      if (a->language_ != b->language_)
-        return a->language_ < b->language_;
+        return false;
       if (a->default_ != b->default_)
         return a->default_;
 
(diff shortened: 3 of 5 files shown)


matthuisman:
@kszaq 
Want to give this a try and review please?

kszaq:
Thank you, @matthuisman - the new code looks sane, I will give it a go this week.

If I may add my 2 cents, I would prefer if you kept the previous formatting in `AdaptiveeTree.h`, i.e. every condition on a new line. This helps readability and with `&&` at the beginning of the line it's easier to blame when adding/removing conditions.

matthuisman:
I'd argue that clang is based on best rules for readability for majority. Everyone has their own style they like, so I think it's best just to keep with clang? If we make an exception for this, then where does it stop? Oh, this also looks a bit nicer than clang thinks etc. 

Its quite nice having something else make the decision for us :)

I do agree looks nicer the old way tho, so maybe @phunkyfish can advise if we can make exceptions? What would happen in the main repo in this case?

~~Unrelated note:
I also see potential for a bug. Think the existing code would segfault if adaption set had no representations.. But possibly only sets with representations make it to this point.~~

Looked into above - DashTree removes any adaptionsets without reprensentations here:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/parser/DASHTree.cpp#L1385
So looks like we are safe to access representations_[0] later on

phunkyfish:
On clang from looking at the code it just appears unclean.

If there was a function to compare a and b I don’t think we’d be having this debate.

matthuisman:
thats basically what this function is doing.
eg. we could change to something like this
```
    bool canMerge(const AdaptationSet *other)
    {
      if (type_ != other.type_)
        return false;

      if (type_ == VIDEO)
      {
        if (group_ != other->group_)
          return false;

        if (std::find(switching_ids_.begin(), switching_ids_.end(), other->id_) == switching_ids_.end())
          return false;

        if (std::find(other.begin(), other.end(), id_) == other.end())
          return false;

        return true;
      }
      else if(type_ == AUDIO)
      {
        if (timescale_ != other->timescale_)
          return false;

        if (duration_ != other->duration_)
          return false;

        13+ more if statements
      }

      return false;
    }
```
but then we just going to have a bunch of if statements?
What's more evil? Long single statement vs multiple statements all returning false?

I think even with the large if statement, the diff still will quite clearly show what is added / removed in the future?

matthuisman:
ok, tried a bit more to get it look a bit cleaner and still clang 
Hopefully a happy middle ground

kszaq:
Checked with my (only) adaptation-set-switching provider: works as expected. Good job. 👍

kszaq:
Hi @matthuisman. What is the plan regarding this PR?

matthuisman:
> Hi @matthuisman. What is the plan regarding this PR?

Pretty sure plan was to put it into Kodi N once it branches and only do bug fixes in Matrix. @glennguy can confirm

glennguy:
A lot of these things are hard to put squarely in a 'bug fix' or 'feature' column.

Yes adding the support for the tag can be looked at as a 'feature' but also it fixes issues encountered when users play a playlists with these tags. Not show stopping of course and it can be worked around by a proxy.

Wrt to Matrix and N: branching for binary add-ons is imminent-ish and the idea is to hold on features, code cleanups/refactoring for the N branch with backports to follow into Matrix.

As for the code formatting - I agree with @phunkyfish, it looks bad. I can see that clang-format has `BinPackArguments: true` but I really think that it should be as the code that was removed, one line per argument. If you don't want clang-format to 'fix' it every time then you can wrap the section with `// clang-format off` and `// clang-format on` comments.

kszaq:
What is the reason for feature freeze if N is far away? I understand this strategy for Kodi but as long as the addon API in Kodi stays the same, shouldn't new features be pushed also to Matrix branch? I can compile things so I don't care that much but I can imagine a PR with some brilliant new feature and users having to choose between using a stable Kodi without this new IA feature or having to switch to unstable Kodi nightly only to get an updated IA. I am fully aware that features (can) introduce bugs and that's why I feel distribution maintainers should keep at least one previous addon release available, then it's really simple to downgrade in case a new IA release breaks something.

phunkyfish:
We are in the process of branching all the binary addons right now of which there are hundreds. So we avoid merging any new changes as we wait for this to complete.

Once that is done we then have place to test fixes more throughly before they go to Matrix.

Its up to maintainers whether or not to backport features however there is usually a time when so much refactoring has happened where the cost of this is too high.

glennguy (approved):

glennguy:
@matthuisman could you kindly rebase to resolve the confilct :)

src/parser/DASHTree.cpp
```diff
@@ -599,6 +599,25 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
...
+              value = (const char*)*(attr + 1);
+            attr += 2;
+          }
+
+          if (schemeIdUri && value)
+          {
+            if (strcmp(schemeIdUri, "urn:mpeg:dash:adaptation-set-switching:2016") == 0)
+              dash->current_adaptationset_->switching_ids_ = split(value, ',');
+          }
+        }
```
phunkyfish, line 620: Fully qualified calls to functions please, e.g. `std::strcmp`.

Also, C++ casts only.

phunkyfish (changes requested):

src/parser/DASHTree.cpp
matthuisman: i was just following the same code used above and below this
Can you please confirm you still want it changed?
Keep the same and then do a big refactor on all later?
Or fix as we go?

matthuisman: i was just following the same code used above and below this
Can you please confirm you still want it changed?
Keep the same and then do a big refactor on all later?
Or fix as we go?

phunkyfish: If it’s fixed as we go it’s less work for the big refactor. But @glennguy’s call as it will fall to him.

matthuisman: yeh, i'm just worried if he does a search and replace, having 1x different with a bunch of others may lead to a bad replace. Plus, I'm not too familiar with C++, so not sure what I'd change the casts too. haha

matthuisman:
rebased so no more conflics

src/parser/DASHTree.cpp
glennguy: I'll sort these as part of the big refactor, for now lets get it in :)

PR #697, opened 2021-05-20
[Android] remove old ndk as not used

matthuisman:
remove wvdecrypter_android.cpp and CMakeLists.txt.ndk

rename wvdecrypter_android_jni.cpp to wvdecrypter_android.cpp
update CMakeLists.txt

Pretty sure we dont use ndk anymore as kodi was moved to jni


glennguy:
Is there a way you can do this in git so that it will show as the file getting renamed? Maybe it would work across 2 commits? If not maybe just better to delete wvdecrypter_android and not rename wvdecrypter_android_jni.

I'm thinking about being able to use blame and history, that's all.

matthuisman:
Haha. I had the same thought. I then did over two commits but when making the pull request, it showed as the same. But now I think that's because the pr window shows the changes all together! I'll try again.

matthuisman:
@glennguy 
looks better now over 2 commits - thank you

matthuisman:
@glennguy 
this also good to go for Nexus

glennguy (approved):
LGTM

phunkyfish (approved):

matthuisman:
argh conflicts again... i'll fix
then can we get it in?

matthuisman:
@glennguy  ready now :)

PR #699, opened 2021-05-21
[Curl] Improvements

matthuisman:
**Test builds:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-699/2/artifacts**

**first:**
make parseheader store the header names in lower-case.
This will make user provided header names lowercase. 
This allows user to override any headers set in IA 
(as long as IA uses lowercase as well - which it does)

**second:**
n need to check if headers contain 'connection' before setting connection 
if the user has provided it, it will be overridden with their value when it loops the media headers

**third:**
add a new method to session to get the media headers so we can use them in SubtitleSampleReader
and anywhere else needed in future.

**fourth:**
remove file.CURLAddOption(ADDON_CURL_OPTION_PROTOCOL, "failonerror", "false");
this does nothing. The default value of failonerror is 0 = off.
https://curl.se/libcurl/c/CURLOPT_FAILONERROR.html

**fifth:**
update SubtitleSampleReader to use 
file.CURLOpen(ADDON_READ_CHUNKED | ADDON_READ_NO_CACHE) instead of file.CURLOpen(0);
with the current Open(0) - kodi tries to fetch the result from filecache and you get weird EOF errors in the log. 
Who knows what other issues that could be causing

**sixth:**
update SubtitleSampleReader to also use the same mediaheaders that Video & Audio already use

**seventh:**
noticed an issue if SubtitleSampleReader returns without setting a codec_handler - Kodi will segfault and crash.
Until now - it wasn't really an issue as there was no check for the failed url download.
So I used @glennguy's new m_started to detect if the reader actually downloaded the subs. 
If it failed, then we disable the stream.

**eighth:**
added keep-alive to Mainfest download and Subtitle download.
This may help speed up manifest refreshes as the socket may stay open.
If they aren't needed - the server can always respond with "close"

**ninth:**
remove setting accept-encoding so IA will just use curls default (deflate, gzip, br)

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -21,6 +21,7 @@
 #include "oscompat.h"
 #include <stdlib.h>
 #include <sstream>
+#include <algorithm>
 
 #ifndef BYTE
 typedef unsigned char BYTE;
@@ -526,7 +527,11 @@ void parseheader(std::map<std::string, std::string>& headerMap, const std::strin
   {
     std::string::size_type pos(b->find('='));
     if(pos != std::string::npos)
-      headerMap[trimcp(b->substr(0, pos))] = url_decode(trimcp(b->substr(pos+1)));
+    {
+      std::string name = trimcp(b->substr(0, pos));
+      std::transform(name.begin(), name.end(), name.begin(), ::tolower);
+      headerMap[name] = url_decode(trimcp(b->substr(pos+1)));
+    }
   }
 }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -272,7 +272,7 @@ bool adaptive::AdaptiveTree::download(const char* url,
     return false;
 
   file.CURLAddOption(ADDON_CURL_OPTION_PROTOCOL, "seekable", "0");
-  file.CURLAddOption(ADDON_CURL_OPTION_PROTOCOL, "acceptencoding", "gzip");
+  file.CURLAddOption(ADDON_CURL_OPTION_HEADER, "connection", "keep-alive");
 
   for (const auto& entry : manifestHeaders)
   {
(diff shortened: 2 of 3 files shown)


kszaq:
Regarding `acceptencoding`, I think we can use Kodi defaults, that is "accept all supported encodings by default": https://github.com/xbmc/xbmc/blob/Matrix/xbmc/filesystem/CurlFile.cpp#L430

I've been running IA with this line deleted and encountered no issues so far:
https://github.com/kszaq/inputstream.adaptive/commit/145463f97dda56ffc8e571e06446bd69a8382b46

matthuisman:
need to keep seekable = "0" otherwise we may get partial files returned.

I agree on acceptencoding, be nice to use just the default.
have you checked debug logs with curl debug logging enabled to make sure accept encoding is still getting sent?
i'm sure it is, i just want to be sure :)

kszaq:
Let me double-check later. I'll report back.

matthuisman:
thanks. Im actually trying now as well :)
by the way - great work on the PRs for LibreELEC and CoreELEC!

UPDATE:
Confirmed we still get accept encoding.
I got "Accept-Encoding: deflate, gzip, br"

So yeh - I like that - I'll remove those lines from PR :)
If curl can handle uncompromising it - then why not accept it :)

user can always overwrite it with their own header if really needed

glennguy:
Looks like a good clean up of some things.

> first:
make parseheader store the header names in lower-case.

Servers aren't supposed to care about case, but from personal experience there are some that do... this could be a breaking change. Kodi does it like this:
https://github.com/xbmc/xbmc/blob/68bf07ee4fab9f844a9bef3ddeeb443be45d9453/xbmc/filesystem/CurlFile.cpp#L852


2,3,4,5 and 6 look good

> seventh:
noticed an issue if SubtitleSampleReader returns without setting a codec_handler - Kodi will segfault and crash.
Until now - it wasn't really an issue as there was no check for the failed url download.

Nice

> So I used @glennguy's new m_started to detect if the reader actually downloaded the subs.
If it failed, then we disable the stream.

You'll need to see if it introduces any strange behaviour with the seek fix changes brought in on #685. Also keep in mind some multiperiod subs are broken atm anyway.

> eighth:
added keep-alive to Mainfest download and Subtitle download.

Again a small thing but might cause issues with some servers to start sending different headers?

9 looks good but also changing default behaviour


Are there any outstanding issues that any of these fix?
With Kodi 20 binary branch so close it might be worth holding off until then for merging and backport required changes after we know we're not causing any breakages.

matthuisman:
the code in xbmc isn't slightly different.
What we are needing to do is set a header if not already set... or at least set it and allow it to be overriden.
If we don't no the casing of the user provided headers, there isn't an easy way to do this.

i guess would need to go back to
`if (mediaHeaders.find("connection") == mediaHeaders.end())`
but somehow make it case in-sensitive to stop Connection and connection...

I think I'll just close this PR and do a really simple one liner to stop it using file cache and we can look at this down the road as I need to get back onto add-ons for a bit :)

PR #705, opened 2021-06-06
Clear cdm_session_str_ on DRM reinitialization

glennguy:
## Description

This change fixes an oversight in #663
`cdm_session_str_` is only set when `SSD_CAPS::SSD_SECURE_PATH` is set which in turn is usually only on Android.
The issue that this change fixes occurs in the following situation/conditions:
1. Multi period stream
2. New period that has different psshsets (eg. different default KID) starts
3. As the vector of cdm sessions is iterated over to set the decrypters again https://github.com/xbmc/inputstream.adaptive/blob/461202194c01d6089e2ccfb25643b47d331dc12c/src/main.cpp#L2420 the cdm_session_str_ from previous period(s) still exists
4. We are now set up to use a secure path for streams that use this psshset whether it is required or not

This wouldn't likely be a problem when every period has the same setup wrt no. of audio/video streams and content protection, but consider for example a program with ad insertion where the ads are different KIDs to the main content, main content has 2 video adaptationsets... psshset_[2] is for video in main program and audio in the ads.

## Motivation and context

Discovered during testing setting SSD_CAPS::SSD_SECURE_PATH on non-android platform (Windows). 

## How has this been tested?

Tested on Windows 10, VTM Go add-on.

No change currently, if in the future secure path is needed on non-android platforms this will avoid issues.

## Types of change

## Checklist:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2170,8 +2170,11 @@ void Session::GetSupportedDecrypterURN(std::string& key_system)
 void Session::DisposeSampleDecrypter()
 {
   if (decrypter_)
+  {
     for (std::vector<CDMSESSION>::iterator b(cdm_sessions_.begin()), e(cdm_sessions_.end()); b != e;
          ++b)
+    {
+      b->cdm_session_str_ = nullptr;
       if (!b->shared_single_sample_decryptor_)
       {
         decrypter_->DestroySingleSampleDecrypter(b->single_sample_decryptor_);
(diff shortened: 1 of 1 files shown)


glennguy:
@matthuisman could I get a quick review when you have a chance?

matthuisman (approved):
Makes sense to me :)

phunkyfish (approved):
LGTM

PR #709, opened 2021-06-08
wvdecrypter: remove clearbytes even for 1 subsample

### LINKED ISSUE #393: Image artefacts (pixel) when streaming video DRM
I have some video playback problem, a lot of miss pixel are falling apart like big artefacts...

i test all the settings available but nothing change, and it's always on the same parts on same video, for exemple in video 1 : at 1m9s i get artefacts, i change params and restart kodi and i get same artefacts at 1m9s ...

some example : 
https://i.imgur.com/av3NDNV.png
https://i.imgur.com/8fuO9Px.png

i have 1 Gbps down and up btw.
Thanks in advance.

kszaq:
When SINGLE_DECRYPT is used, CDM should not get any cleartext bytes, we should remove them even if there is only 1 subsample.

Else if CDM receives 1 subsample with some cleartext data, it cannot decrypt and we get `Decrypt Sample returns failure!` in the log.

Fixes https://github.com/xbmc/inputstream.adaptive/issues/393

### DIFF
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -1141,6 +1141,10 @@ AP4_Result WV_CencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 pool_id,
 
   bool useSingleDecrypt(false);
 
+  // CDM should get 1 block of encrypted data per sample, encrypted data
+  // from all subsamples should be formed into a contiguous block.
+  // Even if there is only 1 subsample, we should remove cleartext data
+  // from it before passing to CDM.
   if ((fragInfo.decrypter_flags_ & SSD_DECRYPTER::SSD_CAPS::SSD_SINGLE_DECRYPT) != 0)
   {
     decrypt_in_.Reserve(data_in.GetDataSize());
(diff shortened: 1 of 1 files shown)


matthuisman (approved):
looks good to me.
Have tested on all Widevine content I have on Windows without issues.
May be good idea for other platforms to be tested as well :)

glennguy:
@kszaq great work!

Could you put in the PR a comment above this block of code giving a quick rundown of what happens similar to [your comment](https://github.com/xbmc/inputstream.adaptive/issues/393#issuecomment-857151008)?

kszaq:
Sure, I will try to add something today evening.

NB that there are some good explanations in Chromium code:
https://github.com/chromium/chromium/blob/d7da0240cae77824d1eda25745c4022757499131/media/cdm/cenc_decryptor.h#L18-L37
https://github.com/chromium/chromium/blob/d7da0240cae77824d1eda25745c4022757499131/media/cdm/cenc_decryptor.cc#L122-L126

kszaq:
Comment added.

glennguy:
Thanks!

glennguy (approved):

matthuisman:
@kszaq
are you able to commit --amend and force push so it tries to build again?
its missing the windows x86_64 build

glennguy:
@matthuisman @kszaq don't bother with rebuilding yet. I can force it to rebuild and have just done that - failure again. The previous build failure was to do with filename too long for another addon and seems to have affected triggers from other PRs

PR #710, opened 2021-06-09
[Dash] Override timeShiftBufferDepth with SegmentTimeline (if present)

### LINKED ISSUE #278: Stream gets stalled because of startNumber not getting incremented
Hi,

It's more like a question than a bug report so sorry for that. I ran across an issue which is bothering me. I've noticed that a stream gets stuck because of startNumber not getting incremented for a while, I've got lots of following logs repeating:
```
2019-06-10 21:27:11.613 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 720p current_start:1
2019-06-10 21:27:11.613 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 1080p current_start:1
2019-06-10 21:27:11.613 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 576p current_start:1
2019-06-10 21:27:11.613 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: tv current_start:1
2019-06-10 21:27:11.613 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: mob1 current_start:1
2019-06-10 21:27:11.613 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 96kbps current_start:1
```

After a few minutes it seems to be working fine:
```
2019-06-10 21:30:08.065 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 720p current_start:2
2019-06-10 21:30:08.065 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 1080p current_start:2
2019-06-10 21:30:08.065 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 576p current_start:2
2019-06-10 21:30:08.065 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: tv current_start:2
2019-06-10 21:30:08.065 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: mob1 current_start:2
2019-06-10 21:30:08.065 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 96kbps current_start:2
2019-06-10 21:30:10.855 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 720p current_start:2
2019-06-10 21:30:10.855 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 1080p current_start:2
2019-06-10 21:30:10.855 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 576p current_start:2
2019-06-10 21:30:10.855 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: tv current_start:2
2019-06-10 21:30:10.855 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: mob1 current_start:2
2019-06-10 21:30:10.855 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 96kbps current_start:2
2019-06-10 21:30:12.991 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 720p current_start:3
2019-06-10 21:30:12.991 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 1080p current_start:2
2019-06-10 21:30:12.991 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: 576p current_start:3
2019-06-10 21:30:12.991 T:139965325371136   DEBUG: AddOnLog: InputStream Adaptive: DASH Full update (w/o startnum): repid: tv current_start:3
```

I've noticed that my MPD file has availabilityStartTime (2016-05-04T00:00:00Z) much more older than publishTime (2019-06-10T21:25:58Z).

It seems that for calculating range_end, availabilty time is used:
```cpp
if (!timeBased && dash->has_timeshift_buffer_ && dash->available_time_)
                    seg.range_end_ += (static_cast<int64_t>(dash->stream_start_ - dash->available_time_ - dash->overallSeconds_ - dash->current_period_start_)*tpl.timescale) / tpl.duration;
```

Which is later used for evaluating startNumber:
```cpp
                //TODO: check if first element or size differs
                  unsigned int segmentId((*brd)->getCurrentSegmentNumber());
                  if ((*br)->segments_[0]->startPTS_ == (*brd)->segments_[0]->startPTS_)
                  {
                    uint64_t search_re = (*br)->segments_[0]->range_end_;
                    for (const auto &s : (*brd)->segments_.data)
                    {
                      if (s.range_end_ >= search_re)
                        break;
                      ++(*brd)->startNumber_;
                    }
                  }
```

When I've forced parser to use publishTime instead of availablityTime by commenting:
```cpp
//    if (!~dash->available_time_)
      dash->available_time_ = dash->publish_time_;
```
it seemed to workaround the issue.

So, I've got few questions (forgive me lack of knowledge):
1) Is server reporting publishTime and availabilityStartTime properly? Aren't they swapped?
2) Do we handle properly calculating range_end by using availabilityStartTime instead of publishTime?
3) Do you have an idea how to fix the issue properly?

I've attached example MPD. Please let me know if you need more logs from Kodi. Hopefully I've described the issue understandable.
[live.txt](https://github.com/peak3d/inputstream.adaptive/files/3273882/live.txt)

dobo90:
In attached mpd timeShiftBufferDepth is translated into 43200 seconds.
But the real length of the stream is 43504. Using the first value makes it
unable to seek in the last 5 minutes of the stream.

Fixes #278

I'm not sure at all whether it is a proper solution. Please see discussion in #278.

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1067,7 +1067,6 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
     AddDuration(mpt, dash->overallSeconds_, 1);
     dash->has_overall_seconds_ = dash->overallSeconds_ > 0;
 
-    uint64_t overallsecs(dash->overallSeconds_ ? dash->overallSeconds_ + 60 : 86400);
     dash->minPresentationOffset = ~0ULL;
 
     dash->currentNode_ |= MPDNODE_MPD;
@@ -1519,6 +1518,23 @@ static void XMLCALL end(void* data, const char* el)
     }
     else if (strcmp(el, "MPD") == 0)
     {
+      if (dash->has_overall_seconds_)
+      {
+        uint64_t overallSeconds = 0;
+
+        // calculate overall seconds from SegmentTimeline
+        for (std::vector<AdaptiveTree::Period*>::iterator b(dash->periods_.begin());
+             b != dash->periods_.end(); ++b)
+        {
+          overallSeconds += (*b)->duration_ / (*b)->timescale_;
+        }
+
+        // override overall seconds from SegmentTimeline as
+        // sometimes timeShiftBufferDepth may contain an invalid value
+        if (overallSeconds)
+          dash->overallSeconds_ = overallSeconds;
+      }
+
       //cleanup periods
       for (std::vector<AdaptiveTree::Period*>::iterator b(dash->periods_.begin());
            b != dash->periods_.end();)
--- a/src/test/TestDASHTree.cpp
+++ b/src/test/TestDASHTree.cpp
@@ -453,3 +453,10 @@ TEST_F(DASHTreeTest, CalculateRedirectSegTpl)
   EXPECT_EQ(tree->periods_[0]->adaptationSets_[1]->representations_[0]->segtpl_.initialization, "https://foo.bar/A48/init.mp4");
   EXPECT_EQ(tree->periods_[0]->adaptationSets_[1]->representations_[0]->segtpl_.media, "https://foo.bar/A48/$Number$.m4s");
 }
+
+TEST_F(DASHTreeTest, OverrideTimeShiftBufferDepthWithSegmentTimeline)
+{
+  OpenTestFile("mpd/segmenttimeline_vs_tsbd.mpd", "http://foo.bar/stream.mpd", "");
+
+  EXPECT_EQ(tree->overallSeconds_, 43504);
+}
--- a/src/test/manifests/mpd/segmenttimeline_vs_tsbd.mpd
+++ b/src/test/manifests/mpd/segmenttimeline_vs_tsbd.mpd
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:cenc="urn:mpeg:cenc:2013" xmlns:mspr="urn:microsoft:playready" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:scte35="urn:scte:scte35:2013:xml" profiles="urn:mpeg:dash:profile:isoff-live:2011" availabilityStartTime="2016-05-04T00:00:00Z" maxSegmentDuration="PT6S" minBufferTime="PT12S" minimumUpdatePeriod="PT18S" publishTime="2021-06-09T14:20:31Z" suggestedPresentationDelay="PT32.12S" timeShiftBufferDepth="PT12H" type="dynamic">
+  <BaseURL>http://foo.bar</BaseURL>
+  <Period id="0" start="PT0S">
+    <AdaptationSet id="0" group="1" bitstreamSwitching="true" segmentAlignment="true" contentType="video" mimeType="video/mp4" startWithSAP="1" maxWidth="1920" maxHeight="1080" frameRate="50">
+      <SegmentTemplate timescale="90000" media="/$RepresentationID$/$Time$.mp4?t=dash-seg" initialization="/$RepresentationID$/init.mp4?t=dash-init" presentationTimeOffset="0">
+        <SegmentTimeline>
+          <S t="14479649308800" d="460800" r="8496"/>
+        </SegmentTimeline>
+      </SegmentTemplate>
+      <Representation id="1080p" codecs="avc1.64002a" bandwidth="8000000" width="1920" height="1080"/>
+    </AdaptationSet>
+  </Period>
+</MPD>


L2501:
imo, trusting the mpd/m3u8 playlists to be correct will become more and more unreliable as this plugins tries to support every stream, many plugins need to add a correcting proxy and end up in a mess of different versions and compatibility

matthuisman:
Did you forget to attach the mpd?

matthuisman:
@dobo90 
looks like IA already does what your doing here:
https://github.com/xbmc/inputstream.adaptive/blob/b2e9078cd4a9306cd2dc30d605421f8c0dbf26bc/src/parser/DASHTree.cpp#L1559

but its using it as a fallback if there is no has_overall_seconds_.
I suspect if you remove mediaPresentationDuration / timeShiftBufferDepth from your MPD - you will already get the correct time.

So really - the question is - should IA change the behavior to use the calculated time instead?
If yes, then all that needs to be done is stop parsing the mediaPresentationDuration / timeShiftBufferDepth  so has_overall_seconds_ becomes false.

has_overall_seconds_ actually could actually be removed as this is the only place that flag is used (detect if we should calculate or not)

dobo90:
@matthuisman that was my initial attempt:
```
From a98a26062839d88342e58980d7a71f369189df79 Mon Sep 17 00:00:00 2001
From: Dobroslaw Kijowski <dobo90@gmail.com>
Date: Wed, 9 Jun 2021 11:35:42 +0200
Subject: [PATCH] DASHTree: don't rely on timeShiftBufferDepth

* Remove also unused overallsecs variable
---
 src/parser/DASHTree.cpp | 7 +------
 1 file changed, 1 insertion(+), 6 deletions(-)

diff --git a/src/parser/DASHTree.cpp b/src/parser/DASHTree.cpp
index 9294720..b4fbba6 100644
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1022,7 +1022,7 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
   }
   else if (strcmp(el, "MPD") == 0)
   {
-    const char *mpt(0), *tsbd(0);
+    const char *mpt(0);
 
     dash->firstStartNumber_ = 0;
 
@@ -1043,7 +1043,6 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       }
       else if (strcmp((const char*)*attr, "timeShiftBufferDepth") == 0)
       {
-        tsbd = (const char*)*(attr + 1);
         dash->has_timeshift_buffer_ = true;
       }
       else if (strcmp((const char*)*attr, "availabilityStartTime") == 0)
@@ -1061,13 +1060,9 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       attr += 2;
     }
 
-    if (!mpt)
-      mpt = tsbd;
-
     AddDuration(mpt, dash->overallSeconds_, 1);
     dash->has_overall_seconds_ = dash->overallSeconds_ > 0;
 
-    uint64_t overallsecs(dash->overallSeconds_ ? dash->overallSeconds_ + 60 : 86400);
     dash->minPresentationOffset = ~0ULL;
 
     dash->currentNode_ |= MPDNODE_MPD;
-- 
2.31.1

```
(taken from https://github.com/xbmc/inputstream.adaptive/issues/278#issuecomment-857571095)

It worked fine in my case but two unit tests are failing (CalculateCorrectSegmentNumbersFromSegmentTemplateWithPTO and CalculateCorrectSegmentNumbersFromSegmentTemplateWithOldPublishTime). Mpds used in those tests are missing ```<S t="X" d="Y" r="Z"/>``` tags. I have no clue what to do in that situation (maybe we should get the time from SegmentTemplate duration?). I'm totally unfamiliar with the MPD specification.

I've just wanted to make a discussion about a bug I've noticed. If you consider that MPD in my pull request is faulty please close this ticket.

matthuisman:
oh, don't get me wrong - I really like the idea.
Much more accurate to calculate it rather than take it from a "estimate" value.
just need to think of a few things

1) do timelines always have duration?
2) can older segments be removed and still exist in the timeline?
   resulting in trying to seek to a segment that has been removed
   eg. timeline calcs we have 10mins of segments - but last 5 mins of those segments is deleted from the server before deleted from the manifest. This manifest then uses a buffer duration of 5mins to stop that.
3) Maybe we use the calculated value if its within a certain % of the buffer time value?
eg. if it's within 10% we use the calculated value. This makes sure we aren't setting it to a way wrong value and respecting the manifests value where we can. But just using the calculated value to be a bit more accurate.
   
Usually in cases like this, we look at other players (eg. Exoplayer) and see what they do

dobo90:
It's not needed anymore in my case.

PR #712, opened 2021-06-11
Create bug report issue template

glennguy:
As the title says.

This is almost 100% verbatim to the Kodi template, happy to take suggestions if anyone thinks there should be additional information collected.


matthuisman:
Nice. Debug log and manifest are the big ones. 

Curl debug component enabled is only thing that may be nice to have

phunkyfish (approved):
LG

Issue #713, opened 2021-06-12
[DRM] License-server key-request fails


jaylinski:
### Description of the issue

When I'm trying to play a DRM protected video via `inputstream.adaptive`, the request to the license-server fails.

I'm experiencing this issue with the following stream:

```
#KODIPROP:inputstream=inputstream.adaptive
#KODIPROP:inputstream.adaptive.manifest_type=mpd
#KODIPROP:inputstream.adaptive.license_type=com.widevine.alpha
#KODIPROP:inputstream.adaptive.license_key=https://drm.ors.at/acquire-license/widevine?BrandGuid=13f2e056-53fe-4469-ba6d-999970dbe549&userToken=https://drm.ors.at/acquire-license/widevine?BrandGuid=13f2e056-53fe-4469-ba6d-999970dbe549&userToken=<token>
https://orf1.mdn.ors.at/out/u/orf1/drmqxa/manifest.mpd
```

*Important*: The license-server URL requires a fresh token for the `userToken`-param which can be requested by calling https://api-stage-tvthek.orf.at/livestream/_token and copying the `base64`-value.

> This stream may be geo-blocked to Austrian IPs.

### Expected behaviour

Kodi (inputstream.adaptive) plays the stream.

> This stream may be geo-blocked to Austrian IPs.

### Actual behaviour

Kodi doesn't play the stream and shows an error. (See `License server returned failure` error in logs below)

```
2021-06-12 16:31:41.742 T:4941    DEBUG <general>: OnPlayMedia /home/osboxes/Downloads/iptv.strm
2021-06-12 16:31:41.759 T:4941    DEBUG <general>: easy_acquire - Created session to https://orf1.mdn.ors.at
2021-06-12 16:31:41.838 T:4941    DEBUG <general>: CCurlFile::GetMimeType - https://orf1.mdn.ors.at/out/u/orf1/drmqxa/manifest.mpd -> application/dash+xml
2021-06-12 16:31:41.842 T:4941    DEBUG <general>: CPlayerCoreFactory::GetPlayers(https://orf1.mdn.ors.at/out/u/orf1/drmqxa/manifest.mpd)
2021-06-12 16:31:41.842 T:4941    DEBUG <general>: CPlayerSelectionRule::GetPlayers: considering rule: system rules
2021-06-12 16:31:41.842 T:4941    DEBUG <general>: CPlayerSelectionRule::GetPlayers: matches rule: system rules
[...]

PR #716, opened 2021-06-13
[WV] More sensible defaults if | not in license url

matthuisman:
Will fix issues like: https://github.com/xbmc/inputstream.adaptive/issues/713 and https://forum.kodi.tv/showthread.php?tid=363338&pid=3046463

Currently if no | is provided in the license url, IA defaults to some Amazon convention that I don't think I've actually seen ever used in production.

I assume it's some legacy value that used to be more common years ago.

Content-Type=application/octet-stream (binary file)
&
R{SSM} (Binary payload)

Are much much more common these days

This should help reduce issues :)

Eventually I think this default code needs updating.
eg. if the url is http://example.com/license_request|Myheaders
then I'm not sure what the payload type will be.... as the existing code here only runs if | not in url.
We would want to check when actually parsing the |. Thus letting us set defaults for headers and payload independently

### DIFF
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -359,9 +359,9 @@ WV_DRM::WV_DRM(const char* licenseURL, const AP4_DataBuffer &serverCert, const u
   if (serverCert.GetDataSize())
     wv_adapter->SetServerCertificate(0, serverCert.GetData(), serverCert.GetDataSize());
 
-  // For backward compatibility: If no | is found in URL, make the amazon convention out of it
+  // For backward compatibility: If no | is found in URL, use the most common working config
   if (license_url_.find('|') == std::string::npos)
-    license_url_ += "|Content-Type=application%2Fx-www-form-urlencoded|widevine2Challenge=B{SSM}&includeHdcpTestKeyInLicense=true|JBlicense;hdcpEnforcementResolutionPixels";
+    license_url_ += "|Content-Type=application%2Foctet-stream|R{SSM}|";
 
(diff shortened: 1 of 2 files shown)


ghost:
> Are much much more common these days

Absolutely

> Eventually I think this default code needs updating.
> eg. if the url is [http://example.com/license_request|Myheaders](http://example.com/license_request%7CMyheaders)
> then I'm not sure what the payload type will be.... as the existing code here only runs if | not in url.
> We would want to check when actually parsing the |. Thus letting us set defaults for headers and payload independently

if there is only one "|" it will fails at https://github.com/xbmc/inputstream.adaptive/blob/461202194c01d6089e2ccfb25643b47d331dc12c/wvdecrypter/wvdecrypter.cpp#L586-L593

then SendSessionMessage() will return `false`, and it will exit here : https://github.com/xbmc/inputstream.adaptive/blob/461202194c01d6089e2ccfb25643b47d331dc12c/wvdecrypter/wvdecrypter.cpp#L475-L483
so... it's ok

matthuisman:
@HawkLiking
awesome, thanks for checking :)
You can review the PR if you like via "Files changed" tab and then "Review changes" button

glennguy (approved):

glennguy:
Looks good, will leave this for Nexus branch though as we're altering the default behaviour and it's not any sort of bug fix

matthuisman:
Perfect :) thank you

ghost (approved):

matthuisman:
@glennguy 
this should be good to go into nexus branch now :)

Issue #703, opened 2021-05-24
Audio sync error on playing specific DASH streams


ratcashdev:
I have a list of 14 channels. On a handful of them (but never on others) the audio gets every time out of sync (and repeatedly playing the same audio section, stuttering) quite soon, after a couple of seconds.

Kodi: 19.1
inputstream.adaptive: 7.6.5
OS: Manjaro Linux and/or Android TV
Cache: default. Customized cache size (up to 200Mb) had no effect.

The log shows:

```
2021-05-24 20:05:02.348 T:466320    INFO <general>: VideoPlayer::OpenFile: https://stream-host.com/64026/2788c6ee5f4f2a004800fdce15fba9f7/60abeace/manifest.mpd?v=5sip:185.29.156.34
2021-05-24 20:05:02.349 T:466950    INFO <general>: Creating InputStream
2021-05-24 20:05:02.355 T:466947    INFO <general>: Python interpreter stopped
2021-05-24 20:05:02.359 T:466950    INFO <general>: AddOnLog: inputstream.adaptive: SetVideoResolution (1920 x 1080)
2021-05-24 20:05:02.503 T:466950    INFO <general>: AddOnLog: inputstream.adaptive: Successfully parsed .mpd file. #Periods: 1, #Streams in first period: 2, Type: live, Download speed: 0.0000 Bytes/s
2021-05-24 20:05:03.335 T:466950    INFO <general>: Creating Demuxer
2021-05-24 20:05:03.335 T:466950    INFO <general>: Opening stream: 1001 source: 256
2021-05-24 20:05:03.375 T:466950    INFO <general>: Creating video codec with codec id: 27
2021-05-24 20:05:03.375 T:466950    INFO <general>: CDVDVideoCodecFFmpeg::Open() Using codec: H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10
2021-05-24 20:05:03.375 T:466950    INFO <general>: Creating video thread
2021-05-24 20:05:03.375 T:466953    INFO <general>: running thread: video_thread
2021-05-24 20:05:03.375 T:466950    INFO <general>: Opening stream: 1002 source: 256
2021-05-24 20:05:03.421 T:466950    INFO <general>: Finding audio codec for: 86018
2021-05-24 20:05:03.422 T:466950    INFO <general>: CDVDAudioCodecFFmpeg::Open() Successful opened audio decoder aac
[...]

PR #724, opened 2021-06-27
Workaround for misaligned SegmentTimeline refreshes

glennguy:
Addresses #703, since we know now there is at least one live manifest packager out there that produces incorrect manifests and a backend that serves up segments with arbitrary timestamps.

The most informative comment in that thread is https://github.com/xbmc/inputstream.adaptive/issues/703#issuecomment-860048289 . It shows how subsequent manifest refreshes are providing timestamps that are slightly out of alignment to where they should be. As the times are slightly ahead the current code in IA will increment one time too many and the new segments are mapped incorrectly against the old.

The way this 'fix' works is that if our segment timestamp (provided by manifest refresh) does not exactly line up with an existing segment timestamp then as long as we're very close (within 2% compared to next segment length) to the existing timestamp we assume that was the intended one.

As far as I can tell from DASH-IF documents SegmentTimeline is meant to be *exact*, not an estimate and the previous code acted on that assertion. New code allows a small amount of flexibility and allows something to work that would otherwise be broken, while not affecting properly constructed streams.


ratcashdev:
THIS WORKS!

ratcashdev:
@glennguy  Is this part of the latest release (.19)?

glennguy:
No, it needs to be merged first before it will become part of a release.  I just need to update the OP slightly then I'll ask for a review. Once accepted then it can be merged and made part of a release. Not long to go now :)

glennguy:
@matthuisman could you take a look and review for me please?

matthuisman (commented):
Looks good!

matthuisman (approved):

src/common/AdaptiveStream.h
```diff
@@ -90,6 +90,8 @@ namespace adaptive
     bool write_data(const void *buffer, size_t buffer_size);
     bool prepareDownload(const AdaptiveTree::Segment *seg);
     adaptive::AdaptiveTree& GetTree() { return tree_; };
+    virtual void SetLastUpdated(std::chrono::system_clock::time_point tm){};
```
phunkyfish: Space before braces: `tm) {};`

phunkyfish (approved):
Just the one, very minor, minor.

Issue #666, opened 2021-04-30
Needed DRM key request (challenge) and session ID to manifest proxy callback


CastagnaIT:
Ntflx service recently implemented new changes in order to request the Manifest,
it appears necessary to provide these info:
- DRM key request (challenge)
- DRM session ID

Currenlty ISA do not provide these info in any way, and i do not know how to do in future if it becomes mandatory for all platforms

The Key request is already mandatory to have HD content on ARM linux devices
and currently i use a hack workaround by hardcoding the value in the addon (of my device) for the first manifest request, and i sniff the value from the license callback for all others manifest callbacks.
This hack is really ugly and is not correct to provide data from different devices and sniffed from the license calllback
I wish to have this fixed in ISA a day...

A good solution would be provide these info directly when ISA make the proxy manifest callback (http request to the addon to get the manifest)
but seem that ISA create the DRM session only after the manifest (addon) callback
and i have no idea on how this can be done

I tried to investigate the situation, but i am not able to make these changes by myself

glennguy:
Yes drm session needs pssh/kid to begin, which we only know from the manifest. 

Off the top of my head (and without seeing what you're doing in nf add-on) you could make a bare-bones version of InitializeDRM():

Roughly:
* ensure ssd_wv is loaded (`GetSupportedDecrypterURN`)
* call `decrypter_OpenDRMSystem(....)`
* create `CDMSESSION` in `cdm_sessions_[1]`
* assign this session's ssd through `decrypter_->CreateSingleSampleDecrypter(...)`
* call `ssd->getSessionId()`
* do other stuff to get the challenge.

Getting this data back to your add-on though would be very hacky using a plugin call. And probably even worse trying to signal back to IA the mpd.

Please correct me if my thinking is wrong. I think the proper solution is implementing a way for binary add-ons to call back to Kodi. This would be useful for much more situations as well, like signalling that playback failed and how/why.

CastagnaIT:
> Getting this data back to your add-on though would be very hacky using a plugin call. And probably even worse trying to signal back to IA the mpd.

with nflx addon i use ISA proxy callbacks to provide manifest and license, as wiki guide:
https://github.com/xbmc/inputstream.adaptive/wiki/How-to-provide-custom-manifest-and-license
i do not think that is needed implement others type of plugin calls or kodi interactions

[...]

PR #725, opened 2021-06-28
Add support to pre-initialize the DRM

CastagnaIT:
This PR add the support to pre-initialize the DRM with a PSSH+KID provided by an add-on via ListItem property.

The purpose is to allow add-ons to make licensed manifest requests via ISA manifest proxy callback.
In order to do licensed manifest requests is needed two data:
- The challenge
- The session id

As discussed previously on #666 and https://github.com/CastagnaIT/plugin.video.netflix/pull/1129#issuecomment-830779762 i have implemented all the things.

In brief:
- The add-on add in the play callback this property (to the list item passed to xbmcplugin.setResolvedUrl) :
`setProperty('inputstream.adaptive.pre_init_data', 'pssh|kid')`
by providing the PSSH and the KID values both encoded as base64 and splitted by a `|`
- When ISA see the `pre_init_data` property data, pre initialize the DRM with the PSSH+KID provided
At this point we have the DRM challenge and the session id generated
- When ISA do the manifest proxy callback to the add-on, attach to the request these two headers:
`challengeB64` (encoded as base64) and `sessionId`
- The add-on when receive the manifest proxy callback can read the challenge and the session id from the headers

NOTE: This not change in any way the current use of the manifest request

The tests were conducted with ntfx add-on, where if the challenge/sid were wrong, video playback would not take place,
and on ARM devices it is the only way to achieve 1080P resolution.

Tested on:
- ARM device with last CoreELEC nightly with Kodi 19 ✅
- Windows 10 64 bit with latest Kodi 19 release ✅
- Android Widevine L3 with latest Kodi 19 release ✅
- Android Widevine L1 with latest Kodi 19 release ✅
- Linux Mint with latest Kodi 19 release ✅

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -454,7 +454,8 @@ public:
   AdaptiveTree();
   virtual ~AdaptiveTree();
 
-  virtual bool open(const std::string &url, const std::string &manifestUpdateParam) = 0;
+  virtual bool open(const std::string& url, const std::string& manifestUpdateParam) = 0;
+  virtual bool open(const std::string& url, const std::string& manifestUpdateParam, std::map<std::string, std::string> additionalHeaders) = 0;
   virtual PREPARE_RESULT prepareRepresentation(Period* period,
                                                AdaptationSet* adp,
                                                Representation* rep,
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1982,7 +1982,8 @@ Session::Session(MANIFEST_TYPE manifestType,
                  uint16_t display_height,
                  const std::string& ov_audio,
                  bool play_timeshift_buffer,
-                 bool force_secure_decoder)
+                 bool force_secure_decoder,
+                 const std::string& drmPreInitData)
   : manifest_type_(manifestType),
     manifestURL_(strURL),
     manifestUpdateParam_(strUpdateParam),
@@ -2005,7 +2006,8 @@ Session::Session(MANIFEST_TYPE manifestType,
     chapter_start_time_(0),
     chapter_seek_time_(0.0),
     play_timeshift_buffer_(play_timeshift_buffer),
-    force_secure_decoder_(force_secure_decoder)
+    force_secure_decoder_(force_secure_decoder),
+    drmPreInitData_(drmPreInitData)
 {
   switch (manifest_type_)
   {
@@ -2222,10 +2224,31 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
     kodi::Log(ADDON_LOG_DEBUG, "Supported URN: %s", adaptiveTree_->supportedKeySystem_.c_str());
   }
 
+  // Preinitialize the DRM, if pre-initialisation data are provided
+  std::map<std::string, std::string> additionalHeaders = std::map<std::string, std::string>();
+
+  if (!drmPreInitData_.empty())
+  {
+    std::string challengeB64;
+    std::string sessionId;
+    // Pre-initialize the DRM allow to generate the challenge and session ID data 
+    // used to make licensed manifest requests (via proxy callback)
+    if (PreInitializeDRM(challengeB64, sessionId))
+    {
+      additionalHeaders["challengeB64"] = challengeB64;
+      additionalHeaders["sessionId"] = sessionId;
+    }
+    else
+    {
+      kodi::Log(ADDON_LOG_ERROR, "%s - DRM pre-initialization failed", __FUNCTION__);
+      return false;
+    }
+  }
+
   // Open manifest file with location redirect support  bool mpdSuccess;
   std::string manifestUrl =
       adaptiveTree_->location_.empty() ? manifestURL_.c_str() : adaptiveTree_->location_;
-  if (!adaptiveTree_->open(manifestUrl.c_str(), manifestUpdateParam_.c_str()) || adaptiveTree_->empty())
+  if (!adaptiveTree_->open(manifestUrl.c_str(), manifestUpdateParam_.c_str(), additionalHeaders) || adaptiveTree_->empty())
   {
     kodi::Log(ADDON_LOG_ERROR, "Could not open / parse manifest (%s)", manifestUrl.c_str());
     return false;
@@ -2240,6 +2263,91 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
   maxUserBandwidth_ = max_user_bandwidth;
 
   return InitializePeriod();
+}
+
+bool Session::PreInitializeDRM(std::string& challengeB64, std::string& sessionId)
+{
+    std::string psshData;
+    std::string kidData;
+    // Parse the PSSH/KID data
+    std::string::size_type posSplitter(drmPreInitData_.find("|"));
+    if (posSplitter != std::string::npos)
+    {
+      psshData = drmPreInitData_.substr(0, posSplitter);
+      kidData = drmPreInitData_.substr(posSplitter + 1);
+    }
+
+    if (psshData.empty() || kidData.empty())
+    {
+      kodi::Log(ADDON_LOG_ERROR, "%s - Invalid DRM pre-init data, must be as: {PSSH as base64}|{KID as base64}", __FUNCTION__);
+      return false;
+    }
+
+    cdm_sessions_.resize(2);
+    memset(&cdm_sessions_.front(), 0, sizeof(CDMSESSION));
+    // Try to initialize an SingleSampleDecryptor
+    kodi::Log(ADDON_LOG_DEBUG, "%s - Entering encryption section", __FUNCTION__);
+
+    if (license_key_.empty())
+    {
+      kodi::Log(ADDON_LOG_ERROR, "%s - Invalid license_key", __FUNCTION__);
+      return false;
+    }
+
+    if (!decrypter_)
+    {
+      kodi::Log(ADDON_LOG_ERROR, "%s - No decrypter found for encrypted stream", __FUNCTION__);
+      return false;
+    }
+
+    if (!decrypter_->HasCdmSession())
+    {
+      if (!decrypter_->OpenDRMSystem(license_key_.c_str(), server_certificate_, drmConfig_))
+      {
+        kodi::Log(ADDON_LOG_ERROR, "%s - OpenDRMSystem failed", __FUNCTION__);
+        return false;
+      }
+    }
+
+    AP4_DataBuffer init_data;
+    const char* optionalKeyParameter(nullptr);
+
+    // Set the provided PSSH
+    init_data.SetBufferSize(1024);
+    unsigned int init_data_size(1024);
+
+    b64_decode(psshData.c_str(), psshData.size(), init_data.UseData(), init_data_size);
+    init_data.SetDataSize(init_data_size);
+
+    // Decode the provided KID
+    uint8_t buffer[32];
+    unsigned int buffer_size(32);
+    b64_decode(kidData.c_str(), kidData.size(), buffer, buffer_size);
+    const char* decodedKid = reinterpret_cast<const char*>(buffer);
+
+    CDMSESSION& session(cdm_sessions_[1]);
+
+    char hexkid[36];
+    AP4_FormatHex(reinterpret_cast<const AP4_UI08*>(decodedKid), 16, hexkid), hexkid[32] = 0;
+    kodi::Log(ADDON_LOG_DEBUG, "%s - Initializing session with KID: %s", __FUNCTION__, hexkid);
+
+    if (decrypter_ && init_data.GetDataSize() >= 4 &&
+        (session.single_sample_decryptor_ = decrypter_->CreateSingleSampleDecrypter(
+             init_data, optionalKeyParameter, (const uint8_t*)decodedKid, true)) != 0)
+    {
+      session.cdm_session_str_ = session.single_sample_decryptor_->GetSessionId();
+      sessionId = session.cdm_session_str_;
+      challengeB64 = decrypter_->GetChallengeB64Data(session.single_sample_decryptor_);
+    }
+    else
+    {
+      kodi::Log(ADDON_LOG_ERROR, "%s - Initialize failed (SingleSampleDecrypter)", __FUNCTION__);
+      cdm_sessions_[1].single_sample_decryptor_ = nullptr;
+      return false;
+    }
+
+  DisposeSampleDecrypter();
+  return true;
 }
 
 bool Session::InitializeDRM()
@@ -2459,7 +2567,7 @@ bool Session::InitializeDRM()
       if (decrypter_ && init_data.GetDataSize() >= 4 &&
           (session.single_sample_decryptor_ ||
            (session.single_sample_decryptor_ = decrypter_->CreateSingleSampleDecrypter(
-                init_data, optionalKeyParameter, (const uint8_t*)defkid)) != 0))
+                init_data, optionalKeyParameter, (const uint8_t*)defkid, false)) != 0))
       {
 
         decrypter_->GetCapabilities(session.single_sample_decryptor_, (const uint8_t*)defkid,
@@ -3325,7 +3433,7 @@ bool CInputStreamAdaptive::Open(const kodi::addon::InputstreamProperty& props)
 {
   kodi::Log(ADDON_LOG_DEBUG, "Open()");
 
-  std::string lt, lk, ld, lsc, mfup, ov_audio;
+  std::string lt, lk, ld, lsc, mfup, ov_audio, drmPreInitData;
   std::map<std::string, std::string> manh, medh;
   std::string url = props.GetURL();
   MANIFEST_TYPE manifest(MANIFEST_TYPE_UNKNOWN);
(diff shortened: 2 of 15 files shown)


matthuisman:
awesome work.

one thing I noticed after a quick look is no need to make changes to wvdecrypter_android.cpp
That actually isn't used anymore.

wvdecrypter_android_jni.cpp is the one used

I have a PR open to clean that up https://github.com/xbmc/inputstream.adaptive/pull/697
One of the main reasons because it isn't clear which is actually used :)

oh, and I think you need to bump the ssd version number here too if changes are made to the decrypter:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/SSD_dll.h#L24

CastagnaIT:
> wvdecrypter_android_jni.cpp is the one used

each time i did not understand which one was used.. i will move the changes to the other file

src/main.cpp
```diff
@@ -2242,6 +2265,110 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
...
+      kidData = drmPreInitData_.substr(posSplitter + 1);
+    }
+
+    if (psshData.empty() || kidData.empty())
+    {
+      kodi::Log(ADDON_LOG_ERROR, "%s - Invalid DRM pre-init data, must be as: {PSSH as base64}|{KID as base64}", __FUNCTION__);
+      return false;
+    }
+
+    cdm_sessions_.resize(2); //(adaptiveTree_->current_period_->psshSets_.size());  TODO: One session for a single period/PSSH, why is needed size of 2?
```
CastagnaIT: @matthuisman i have a question on this line,
as you can also see in the `Session::InitializeDRM` method the cdm_sessions_ variable is resized, but, why the first value of the array is not used?

glennguy: Periods always have a `psshSets[0]` by default (I think it's just used for AES-128) - lots of logic is relying on this i.e https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/src/main.cpp#L2291
https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/src/main.cpp#L2436
https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/src/common/AdaptiveTree.cpp#L286-L297

CastagnaIT:
I have done the suggested changes, now is ready for review,
please wait to merge until i will get a feedback from a WV L1 android device with 4K

lib/libbento4/Core/Ap4CommonEncryption.h
```diff
@@ -523,7 +523,7 @@ class AP4_CencSingleSampleDecrypter
     virtual AP4_UI32 AddPool() { return 0; };
     virtual void RemovePool(AP4_UI32 poolid) {};
     virtual const char* GetSessionId() { return NULL; };
-
+    virtual const char* GetChallengeB64Data() { return NULL; };
```
ksooo: Is it really good practice to directly patch a third party library instead of PR-ing the change upstream and to use a patch file here until the upstream PR got merged and libbento used here was updated?

glennguy: Good point however unfortunately we're already sitting on top of several years of direct patches (lots) on a very outdated libbento, it really needs to be brought up to date. I tried last year sometime but it introduced bugs :/

To address this issue @CastagnaIT you should be able to accomplish what you want without changes to libbento - introduce the new function in SSD_dll.h

src/main.cpp
```diff
@@ -2242,6 +2265,112 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
...
+    if (!decrypter_->HasCdmSession())
+    {
+      if (!decrypter_->OpenDRMSystem(license_key_.c_str(), server_certificate_, drmConfig_))
+      {
+        kodi::Log(ADDON_LOG_ERROR, "%s - OpenDRMSystem failed", __FUNCTION__);
+        return false;
+      }
+    }
+
+    for (size_t ses(1); ses < cdm_sessions_.size(); ++ses)
```
glennguy: Since we know the size of cdm_sessions_ already and we're only interested in [1] you can remove this and the other for loops.

src/main.cpp
```diff
@@ -2242,6 +2265,112 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
...
+      {
+        session.cdm_session_str_ = session.single_sample_decryptor_->GetSessionId();
+        sessionId = session.cdm_session_str_;
+        challengeB64 = std::string(session.single_sample_decryptor_->GetChallengeB64Data());
+      }
+      else
+      {
+        kodi::Log(ADDON_LOG_ERROR, "%s - Initialize failed (SingleSampleDecrypter)", __FUNCTION__);
+        for (unsigned int i(ses); i < cdm_sessions_.size(); ++i)
+          cdm_sessions_[i].single_sample_decryptor_ = nullptr;
```
glennguy: Just `DisposeSampleDecrypter();` should be fine here

src/main.cpp
```diff
@@ -2242,6 +2265,112 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
...
+        kodi::Log(ADDON_LOG_DEBUG, "%s - Initializing session with KID: %s", __FUNCTION__, hexkid);
+
+        for (unsigned int i(1); i < ses; ++i)
+          if (decrypter_->HasLicenseKey(cdm_sessions_[i].single_sample_decryptor_,
+                                        (const uint8_t*)decodedKid))
+          {
+            session.single_sample_decryptor_ = cdm_sessions_[i].single_sample_decryptor_;
+            session.shared_single_sample_decryptor_ = true;
+            break;
+          }
```
glennguy: This should be able to be removed entirely

glennguy (changes requested):

CastagnaIT:
i have a problem discovered only now (and reported also by others) because i was not aware that was needed reboot kodi every time i update ISA to make libreelec works with the updated binary

the problem in my theory after a lot of tests, is that in the challenge data sometimes there are some special chars that i am not able to send to the http headers, results that in the data added in the "challengeB64" header is ignored?! and leaves the value blank

i have tried also update the b64_encode helper with new code, with no improvement.
i have tried also enclose the challenge two times as b64_encode, with no improvement.
i have tried send the challenge data by making a manifest Post request by add the param:
`file.CURLAddOption(ADDON_CURL_OPTION_PROTOCOL, "postdata", postData.c_str());`
in the manifest download method, in this way seem works better, but after 1 or 2 playback the problem happen again, blank challenge value...

and the challenge data there is always, this is the data size reported of
challenge_ in wvdecrypter.cpp
when works and also when not works (blank data sent):
`GetChallengeB64Data CHALLENGE DATA SIZE: (1773)`

I ran out of ideas on how to proceed

UPDATE: in the
`file.CURLAddOption(ADDON_CURL_OPTION_PROTOCOL, "postdata", postData.c_str());`
the data is passed then the problem could be in curl

lib/libbento4/Core/Ap4CommonEncryption.h
glennguy: > introduce the new function in SSD_dll.h

I'm wrong about this (was thinking of `decryptor_` not `single_sample_decrypter_`), ~~changes need to stay in libbento for now.~~ nope wrong again haha

CastagnaIT:
found a way to keep Ap4CommonEncryption.h unchanged
and in this way i have cast with the WV_CencSingleSampleDecrypter  class the decripter to get the value
but unfurnately the problem is not resolved sometimes works other no

this is an example of the log:
```
2021-07-03 15:18:31.744 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: PreInitializeDRM - Entering encryption section
2021-07-03 15:18:31.786 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: CDM version: 4.10.2252.0
2021-07-03 15:18:31.967 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: cdm::OnInitialized: true
2021-07-03 15:18:32.063 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: PreInitializeDRM - Initializing session with KID: 0000000003D267490000000000000000
2021-07-03 15:18:32.404 T:5105    DEBUG <general>: ------ Window Deinit (DialogNotification.xml) ------
2021-07-03 15:18:32.786 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: CDMMessage: 1 arrived!
2021-07-03 15:18:32.823 T:5297    ERROR <general>: AddOnLog: inputstream.adaptive: GetChallengeB64Data: is executed
2021-07-03 15:18:32.823 T:5297    ERROR <general>: AddOnLog: inputstream.adaptive: etChallengeB64Data: converted data: ¤F5à¦ÚÃÿÐ0ºaêí
2021-07-03 15:18:32.823 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: GetChallengeB64Data: data set x&Ò
2021-07-03 15:18:32.848 T:5297    DEBUG <general>: AddOnLog: inputstream.adaptive: CDMMessage: 4 arrived!
2021-07-03 15:18:32.871 T:5297    DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'challengeB64: '
2021-07-03 15:18:32.871 T:5297    DEBUG <general>: CurlFile::ParseAndCorrectUrl() adding custom header option 'sessionId: CB1CB479F8DCDF56CC2F3EF54744E32B'
2021-07-03 15:18:32.871 T:5297    DEBUG <general>: CurlFile::Open(0xea61b430) http://127.0.0.1:53821/msl/get_manifest?videoid=80230399
2021-07-03 15:18:32.871 T:5297    DEBUG <general>: easy_acquire - Created session to http://127.0.0.1
2021-07-03 15:18:32.873 T:5299    DEBUG <general>: [plugin.video.netflix (0)] HTTP Server: received GET request /msl/get_manifest?videoid=80230399
2021-07-03 15:18:32.873 T:5299    ERROR <general>: ----------------------------------------
2021-07-03 15:18:32.873 T:5299    ERROR <general>: 
                                                   
2021-07-03 15:18:32.873 T:5299    ERROR <general>: Exception happened during processing of request from
###cutted addon raised exception due to None value###
```

glennguy:
In GetChallengeB64() you're returning a c_str() of a local variable, the local variable is destroyed soon after the function exits so sometimes you're able to read the information in time, other times it's too late and the pointer is junk.
One way would be to get it to return std::string instead then call c_str() where you need it in main.cpp

CastagnaIT:
@glennguy it seems to have worked!!! I played 6 videos without any problems, great 🤩
I'll have the other guys run some tests to confirm

CastagnaIT:
i have done all tests with success
this PR is done you can review

CastagnaIT:
I need to do some changes but can only be done after the freeze fix PR will be merged

wvdecrypter/wvdecrypter_android_jni.cpp
```diff
@@ -564,48 +570,61 @@ bool WV_CencSingleSampleDecrypter::KeyUpdateRequest(bool waitKeys)
...
-  optParams_.clear();
-
-  std::vector<char> keyRequestData = keyRequest.getData();
+  keyRequestData = keyRequest.getData();
   Log(SSD_HOST::LL_DEBUG, "Key request successful size: %lu", keyRequestData.size());
+  return true;
+}
 
-  if (!SendSessionMessage(keyRequestData))
+bool WV_CencSingleSampleDecrypter::KeyUpdateRequest(bool waitKeys, bool sendSessionMsg /* = true */)
```
glennguy: ~~comment left over from debugging?~~
as above

wvdecrypter/wvdecrypter.cpp
```diff
@@ -397,7 +400,7 @@ void WV_DRM::OnCDMMessage(const char* session, uint32_t session_size, CDMADPMSG
 |   WV_CencSingleSampleDecrypter::WV_CencSingleSampleDecrypter
 +---------------------------------------------------------------------*/
 
-WV_CencSingleSampleDecrypter::WV_CencSingleSampleDecrypter(WV_DRM &drm, AP4_DataBuffer &pssh, const uint8_t *defaultKeyId)
+WV_CencSingleSampleDecrypter::WV_CencSingleSampleDecrypter(WV_DRM &drm, AP4_DataBuffer &pssh, const uint8_t *defaultKeyId, bool sendSessionMsg /* = true */)
```
glennguy: Are these to let know about the default in the declaration? Would it be better to remove the default from the declaration and put here (and the other one) instead?

src/common/AdaptiveTree.h
```diff
@@ -454,7 +454,8 @@ class ATTRIBUTE_HIDDEN AdaptiveTree
   AdaptiveTree();
   virtual ~AdaptiveTree();
 
-  virtual bool open(const std::string &url, const std::string &manifestUpdateParam) = 0;
+  virtual bool open(const std::string& url, const std::string& manifestUpdateParam) = 0;
+  virtual bool open(const std::string& url, const std::string& manifestUpdateParam, std::map<std::string, std::string> additionalHeaders) = 0;
```
glennguy, line 458: Just a quick question - would the additionalHeaders ever be required for a manifest update: https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/src/parser/DASHTree.cpp#L1669

wvdecrypter/wvdecrypter_android_jni.cpp
CastagnaIT: i have add the `/* = true */` comment the same way code is handled in Kodi sources
to make more easy to understand when an method argument has a default value set

wvdecrypter/wvdecrypter.cpp
CastagnaIT: if you prefer I remove all default values from method argument and assign them to each method call

glennguy: To be honest I'm not really sure if there is a right/wrong way, maybe just preference... @phunkyfish do you have a suggestion for these situations?

src/common/AdaptiveTree.h
CastagnaIT: for my use case no, but perhaps it is better add it?

wvdecrypter/wvdecrypter.cpp
CastagnaIT: I think there are not many ways,
i can remove the defalut value, but at this point i will change the argument from `sendSessionMsg` to `skipSessionMessage`, so it becomes more understandable

phunkyfish:
> I need to do some changes but can only be done after the freeze fix PR will be merged

Please post when this is the case and I will review.

phunkyfish:
Apologies, hit wrong button. No intention to close this.

src/common/AdaptiveTree.h
glennguy: Looking at it now I can't think of a reason we would need to redo this mid stream - I don't think I've seen more than one session open at a time, let alone changing sessions mid stream. Leave for now and if it comes up in the future it's not a big change.

CastagnaIT:
@phunkyfish i have rebased and done the last changes

wvdecrypter/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -328,16 +328,23 @@ void CdmAdapter::UpdateSession(uint32_t promise_id,
       response, response_size);
 }
 
-void CdmAdapter::SetSessionActive()
+void CdmAdapter::SetSessionActive(bool isActive)
 {
-  session_active_ = true;
+  std::lock_guard<std::mutex> lock(session_active_mutex_);
```
Portisch: it's an atomic var, no need for mutex

CastagnaIT: did not know the meaning, now learned

wvdecrypter/wvdecrypter.cpp
phunkyfish: I think leave the default value. Honestly, bools are poor choice in this case. Using an enum is better.

src/main.cpp
```diff
@@ -3401,7 +3509,14 @@ bool CInputStreamAdaptive::Open(const kodi::addon::InputstreamProperty& props)
                 max_user_bandwidth);
     }
     else if (prop.first == "inputstream.adaptive.play_timeshift_buffer")
+    {
       m_playTimeshiftBuffer = stricmp(prop.second.c_str(), "true") == 0;
+    }
+    else if (prop.first == "inputstream.adaptive.pre_init_data")
+    {
+      kodi::Log(ADDON_LOG_DEBUG, "found inputstream.adaptive.pre_init_data: [not shown]");
```
phunkyfish, line 3522: We really need to explain here what pre_init_data means. Is this documented somewhere?

src/main.cpp
```diff
@@ -3423,10 +3538,9 @@ bool CInputStreamAdaptive::Open(const kodi::addon::InputstreamProperty& props)
 
   kodihost->SetProfilePath(props.GetProfileFolder());
 
-  m_session = std::shared_ptr<Session>(new Session(manifest, url.c_str(), mfup, lt, lk, ld, lsc,
-                                                   manh, medh, props.GetProfileFolder(),
-                                                   m_width, m_height, ov_audio,
-                                                   m_playTimeshiftBuffer, force_secure_decoder));
+  m_session = std::shared_ptr<Session>(new Session(
+      manifest, url.c_str(), mfup, lt, lk, ld, lsc, manh, medh, props.GetProfileFolder(), m_width,
+      m_height, ov_audio, m_playTimeshiftBuffer, force_secure_decoder, drmPreInitData));
```
phunkyfish, line 3548: For a future refactor. This function call makes little to no sense at all.

wvdecrypter/wvdecrypter.cpp
```diff
@@ -469,7 +471,7 @@ WV_CencSingleSampleDecrypter::WV_CencSingleSampleDecrypter(WV_DRM &drm, AP4_Data
     reinterpret_cast<const uint8_t *>(pssh_.GetData()), pssh_.GetDataSize());
 
   int retrycount=0;
-  while (session_.empty() && ++retrycount < 100)
+  while (!drm.GetCdmAdapter()->IsSessionActive() && ++retrycount < 100)
```
phunkyfish, line 474: Also not for this PR but this is not a good way to do this. A wait for condition would be much better. Then no need to thread::sleep_for which is expensive.

phunkyfish (changes requested):

src/main.cpp
CastagnaIT: in the wiki i can document it when it is merged,
I can add a short comment line there as well

CastagnaIT: done (in the source code)

phunkyfish (approved):

matthuisman:
This fixes: https://github.com/xbmc/inputstream.adaptive/issues/666

glennguy (approved):

Issue #728, opened 2021-07-04
[CoreELEC] Kodi freeze when stop the played video


CastagnaIT:
This happen to me very rarely, on my CoreELEC matrix nightly (ARM)
and only after play/stopped a lot of videos, after about 6 or 10 videos depends, sometimes this problem not happen and all works good

When this problem happen, can happen these behaviours:
1) At video stopped, kodi freeze, and after some minuts resolves itself and you can play again as usual
2) At video stopped, kodi freeze, and after some minuts, kodi reboot itself
3) At video stopped, kodi freeze, and you need to hard reboot the device

After tried a lot of times i have managed to restrict the field on the problem:
[kodi.old.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6759643/kodi.old.log.txt)
[kodi_crashlog_20210704080408.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6759644/kodi_crashlog_20210704080408.log.txt) (same log but with coreleec crash report)

on first log file, line 9150 `Session::DisposeDecrypter()`
`2021-07-04 08:02:05.088 T:5707    ERROR <general>: AddOnLog: inputstream.adaptive: Session::DisposeDecrypter(): ENTERED`
is run and the code freeze inside this method, seem to always wait at cdm close session message,
then if cdm session close message:
- arrive late, then happen behaviour 1
- not arrive,  then happen behaviour 2 or 3

you can see from the log that in this case Kodi is freezed 1 minute:

`2021-07-04 08:03:05.484 T:5707    DEBUG <general>: AddOnLog: inputstream.adaptive: CDMMessage: 4 arrived!`

fritsch:
Looks like the CoreElec specific code tries to read / write to the sys fs most likely race when changing refreshrates, see:
```
2021-07-04 08:03:05.525 T:5105    DEBUG <general>: AMLUtils::aml_handle_display_stereo_mode old mode 3doff new mode 3doff
2021-07-04 08:03:05.525 T:5105    DEBUG <general>: AMLUtils::aml_handle_display_stereo_mode - no change needed
2021-07-04 08:03:05.525 T:5105    ERROR <general>: GetString: error reading /sys/class/amhdmitx/amhdmitx0/custom_mode

``` 
Sorry - this code is not in kodi at all anymore, please continue in the coreelec forum.

The error should go away when you plain disable AML in the settings, right?

CastagnaIT:
can you explain me how to disable AML?

[...]

PR #729, opened 2021-07-07
Fix kodi freeze by session close

Portisch:
Two points do cause Kodi to freeze randomly on session end:

* CloseSession should not be called in dispose function
* SetTimer callback did reset `exit_thread_flag` status to `false` while being in CloseSession function cause waiting endless that `timer_thread_running` become false

More than 150 start/stop cycles (30s cycle time) where successful done with this modification. Tested with CoreELEC 19.2-Matrix_nightly and last Amazon VOD. Before these changes Kodi got randomly frozen between 1-30 cycles.

This should solve issue #728

### DIFF
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -174,6 +174,7 @@ public:
 
   void GetCapabilities(const uint8_t* key, uint32_t media, SSD_DECRYPTER::SSD_CAPS &caps);
   virtual const char *GetSessionId() override;
+  void SetSessionActive();
   void CloseSessionId();
   void SetSession(const char* session, uint32_t session_size, const uint8_t *data, size_t data_size)
   {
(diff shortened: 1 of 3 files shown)


wvdecrypter/wvdecrypter.cpp
```diff
@@ -1418,7 +1438,11 @@ class WVDecrypter : public SSD_DECRYPTER
   virtual void DestroySingleSampleDecrypter(AP4_CencSingleSampleDecrypter* decrypter) override
   {
     if (decrypter)
+    {
+      // close session before dispose
+      decrypter->CloseSessionId();
```
CastagnaIT: you can leave untouched libbento (Ap4CommonEncryption.h)
by removing virtual to CloseSessionId method,
and replace this with:
`static_cast<WV_CencSingleSampleDecrypter*>(decrypter)->CloseSessionId();`

Portisch: Yes, it's just because of base class. Will change it

glennguy:
Great work @Portisch, thanks for finding and fixing!

glennguy (approved):

matthuisman:
@MisterD81
are you able to please review this?
as it relates to your original work here: https://github.com/xbmc/inputstream.adaptive/pull/589
maybe even runtime test with a Youtube movie if you have available to make sure license renewal still works?

I'm about to test with Binge to confirm license renewal still works

phunkyfish (approved):
LGTM

MisterD81:
> @MisterD81
> are you able to please review this?
> as it relates to your original work here: #589
> maybe even runtime test with a Youtube movie if you have available to make sure license renewal still works?
I'm using latest LE (Leia) version of ISA already which includes this change already. I haven't recognized some particular problems with it. Youtube still work.
This patch does not affect #589, only how the session is stopped, but not the timer/license renewal itself.

Issue #392, opened 2020-03-12
VOD video problems - Missing support for Kodi stream buffering


CastagnaIT:
Since always, and especially in recent months many users complain about problems in the playback

These problems are due to latencies in the VOD service or networks that are not always stable due to the ISP or WIFI.

The problem is the lack of a buffer,
getting the data in real time is not a good way to VOD playback

I do not know if it is already supported, in this case closed the Issue, i just need to know how to customize it and increase the size.

Myself have buffering problems and it starts to get worse:
- at best, playback pauses and resumes after about 10/15 seconds
- at worst, the playback is stopped and exit to the list screen, and you have to resume it

This can also happen several times in a single playback

It would be a good idea to use Kodi buffer to load as much data as possible in as little time as possible, in order to suppress these problems

glennguy:
This has come up a few times before, see #8 and #14 for example.

It would be great to have but it's not trivial to implement. At the moment from what I can see segments are only downloaded at Kodi's request (to fill 8 second internal buffer - cannot be changed) so there would need to be a whole new section built to have our own buffer within inputstream adaptive, associated logic, also buffering for different streams (for adaptive bitrate switching)... I'm sure more things!

If waiting for this isn't an option you could do a proxy server in python to pre-download future segments

zero972:
I came across this issue also, 8 sek. internal buffer is propably to small for flawless playback, especially with 4k content. Adaptive bitrate switching is nice to have, but not a deal breaker since you can switch the streams by yourself if needed. 

Maybe adding a buffer for the actual stream first (to fix the playback issues) would be a lot less work to do as to implement both features at once?

yoshimo:
How does this compare to https://kodi.wiki/view/HOW-TO:Modify_the_video_cache `?
That page describes how to increase the cache for 

> all video files, including local network, internet, and even the local hard drive. Default value is 0 and will only cache videos that use internet file paths/sources.

dagwieers:
I have asked for better triage in this project so we can make open feature requests more upfront, rather than having the same thing being brought up ever so often in another issue. It doesn't help to converge to a proper fix (or get the proper attention from those that can implement it).

dagwieers:
We can already use the Wiki to document and track features or known issues, so don't let that keep us from collaborating ;-)

Besur:
are you guys talking about artefacts with DRM vod playback like this https://imgur.com/8fuO9Px ? cause i don't know if i have to open an issuer for my problem

[...]

Issue #141, opened 2018-04-16
Is there a way to buffer ahead?


peterpan192:
I know, there have already been issues regarding the buffering situation. 
However, the best time of the year has started (NBA Playoffs) and the NBA has changed their streaming system to MPEG Dash. 
My internet connection is very slow so that I cannot even stream the games on lowest quality. 
It would be so sweet if KODI was able to buffer ahead so that I could start the playback of games in the morning and watch them after work. 
Is there any workaround for me here?

peak3d:
Unfortunately still no buffering implemented,

glennguy:
duplicate

Issue #14, opened 2017-02-28
It's possible to add a manual size for cache ??


DjDiabolik:
The inputstream.adaptive on kodi-17.x it's been used for example for youtube video.
The currently video addons 5.3.10 whitout inpustream activated have a your personal size for cache setted by default at 20MB....... 

Apparently this value it's not been used when kodi use, now, the inputstream.adaptive....... it's possible to add somethigs similar ?? I thinks this it's be usefull for example for live stream to create a little buffer before to start the reproduction and this helped expecilly who does not have a super fast connection...

CaptainTK:
I second to that. Ideally, please use the values specified in [advancedsettings.xml](http://kodi.wiki/view/HOW-TO%3AModify_the_video_cache).

DjDiabolik:
yeah.... i have that but apparently it's useless also to set this value to 500MB :) I thinks the only problems it's kodi 17.x.....

peak3d:
inputstream.adaptive does not use those internal kodi settings.
Currently 8 sec. internal kodi buffer is used. This is fixed and cannot be changed from outside.

Buffering is one of the next things which will be done.

peterpan192:
Is there an approximate time when it will be implemented?

mirkorobocop:
@peak3d Is this already implemented?

glennguy:
Duplicate of #8

PR #732, opened 2021-07-10
Buffering and adaptive bitrate switching

### LINKED ISSUE #726: Halfway decent buffering concept for livestreams (e.g. HLS/Dash)
I might be missing something here and maybe everything has been entirely thought through and things are the way they are supposed to, but somehow the silly implementation of virtually any player is staggering me when it comes to proper buffering.

For instance, out of concrete practical needs these days during the soccer EM 2020/2021 - Kodi in conjunction with the ORF tvthek addon, which seems to make use of that very inputstream adaptive plugin, it is just awful.

Given the premise of a suboptimal, non-constant data throughout (otherwise also terrible implementations won't look too bad, which is part of my critic), be it due to the user's own internet access, interconnection between ISPs, a VPN/proxy/other tunnel congestion or the streaming provider itself, I would expect the following:

The player won't play anything at first but keep filling it's buffer which may be defined by the user in the best case without any xml/file editing acrobatics. This maximum amount naturally is limited by the streaming provider as at some point, certain segments will simply expire and not be available anymore so data loss by principle is unavoidable. Taken that aside for a moment, I would further assume that the player will then start playing the content while keep on requesting new parts as fast as possible to refill the buffer. Extending that to HLS with separate segments, the parallel request of those would be a clever thing as well, given the observation that a download of only one often turns out to be rather slow thanks to the TCP window fuss.

Instead I encounter the extremely frustrating and nutty behaviour as follows:

1. Any fiddling in that advancedsettings.xml for Kodi doesn't seem to have any effect for HLS/Dash streaming stuff whatsoever. So much for option 1 (buffer ANY source). 

2. While Kodi seems to halfway decently buffer the input stream at the start, the total amount looks very little like those 8 second window or so which have been mentioned here and there around ... Once that already way too stingy buffer runs empty (which especially on mobile connections like LTE occurs at a high probability), the player doesn't wait until at least that is full again, but no, it attempts to resume playback as soon as only a part of that small buffer is regained, unsurprisingly almost asking for the next buffer-underrun and stutter of course.

The oddity becomes apparent when one manually pauses the playback in such buffer-underrun event and starts only when at least the "stingy buffer" is entirely full again. Et voila, the playback is at least not pausing right away again but instead keeps on playing quite a while even if the total input rate is slightly below the normally demanded one. So while in such cases flawless playback would only be possible by really using a huge buffer and delay, even when assuming that repeated buffer-underruns are unavoidable due to the too slow throughput, it is beyond my comprehension why almost every player implements the buffering "strategy" in such a lousy and annoying way.

I think you as a developer at some point replied that a real decent buffering strategy would be very complex to implement, like dealing with different steam qualities, which segments to request, how to switch, etc. but how difficult can it be to at least improve the buffering method with a given stream comparable to what anyone apparently can manually due by pausing.

I would already be satisfied if Kodi or your plugin (whichever is effectively doing the buffering) would allow to max out the buffer up to the streaming operator's server limit and wait until that buffer is completely full again before resuming playback for one given stream.

glennguy:
## Description
This continues on the great work started last year by @spyderboy92 and @peak3d to bring this feature to this add-on.

## Motivation and context
The complaint of no real buffering feature comes up constantly in this repo and on forums. Users are usually unaware of the limitation and end up frustrated when changing buffering settings in advancedsettings.xml doesn't improve the performance. The GSOC 2020 project of implementing a solution to this gave some hope however due to personal commitments @peak3d wasn't able finalise and bring the changes in.
Adaptive bitrate switching is also a much requested feature, I know in my home country where internet connections are mediocre at best and many still on ADSL2 this would be a much welcomed improvement. Add-on authors have incorporated features like stream selection dialog and logic such as in @matthuisman's add-ons to help the user get the correct stream for their needs however a solution where the stream will adapt over playback to the conditions is what's really needed.

From the starting point in #506 multiple improvements and fixes have been made. I've documented briefly in each commit for myself and others to aid future contributions.

## How has this been tested?
The bulk of change in the code is in the AdaptiveStream class, and so the tests have been updated to run correctly.
There is an outstanding issue with how the thread locks are structured and acquired. So far in my testing it behaves fine at runtime but under automated test it caused issues. There is a workaround in place of sleeping the testing thread for a brief amount of time to avoid deadlocking, I hope that in the short term future myself or another can rewrite the affected area to acquire and release locks in a consistent order.

Users should have a pleasant experience hopefully devoid of any buffering pauses, as well as seeing the best quality that's possible on their connection.

## Types of change

## Checklist:

- [x] My change requires a change to the documentation, either Doxygen or wiki
- [x] I have added tests to cover my change

Note:
The codebase as a whole is full of mixed conventions and doesn't conform to the same unified standard as required in xbmc. The plan is to follow this PR with a big cleanup involving renaming variables to be consistent and fixing formatting issues. 

Fixes #726, #392, #141, #14, #8


glennguy:
TODO/issues:
* ~~HLS live multiperiods still behaves strange~~ behaviour seems good now
* ~~If we switch to a new representation on the last segment of a period, the init segment won't be downloaded.~~ Solved
* ~~Stream switching won't work if we're right at the live edge. The closest you can get is second segment from the edge i.e for a 1 minute timeshiftbuffer we're always viewing between 00:48 and 00:54, and in this scenario we can't choose a new representation, the code just doesn't seem structured to allow for it. Suggestion: prevent seeking to this point.~~ Solve separate to this PR
* ~~At some point I had working not choosing the same starting rep/quality for each new period, needs to be fixed.~~ fixed
* ~~Not tested with live DASH streams.~~ tested with a couple, seems ok
* ~~Not tested to see if it switches to a lower quality when bandwidth drops.~~ Seems to work quite well
* ~~AdaptiveStream tests can have unpredictable results regarding how many segments are loaded. This causes issues when relying on results like `downloadedUrls.back()`~~ AdaptiveStream tests have been reworked.

popy2k14:
Hey guys, thanks a lot for this rework. I hope this solves most of the buffering issues.

A question, was also Live Dash streaming like Zattoo was taken into consideration?

glennguy:
@popy2k14 No idea about 'Zattoo' but streams that work at the moment should work in this. Nothing has changed with regards to parsing particular types of manifests, just how downloading segments and storing them in a buffer along with switching quality to suit the current bandwidth.

popy2k14:
ok, thx, is there an android TV (arm7) testbuild for kodi 18.9? (yes my family dont want me to upgrade :-( )
I can test it beause i have issues in kodi with zattoo streams and the native ATV Zattoo app has no issues at the same time.

glennguy:
I'm sorry to say the code is only for Kodi 19 at the moment. Also, this will be released for Kodi 20 first, and *possibly* be backported to Kodi 19 afterwards. Currently 19 and 20 binary addons are compatible but that is likely to change sometime in the near future.
Kodi 18 however is pretty much finalized. If there's a major bug found it *may* get backported to 18 but definitely no features I'm sorry to say.

popy2k14:
ok, thanks for clearing things up.
Is there an kodi 19 android TV (arm7) testbuild?

Thank you

glennguy:
Sure thing, current builds are here: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-732/15/artifacts

phunkyfish:
Should all commits be reviewed or just yours @glennguy?

Also, there are a number of merge commits. In a PR they should not be required., or the PR should be cleared up to not depend on them. If that's just way too much work and we should just get this in please let me know.

glennguy:
Just my commits, I can't speak to the ones by peak and spyderboy.

I'll see if I can rebase and get rid of the merge commits to clear it up.

kontell:
I've installed this on a Shield Pro and have been using it with @matthuisman's SkySportNow addon and Youtube, here's by 2 cents.

There doesn't seem to be a way to disable the adaptive switching. The stream selection option "manually select all streams" doesn't actually disable the switching. It just allows you to manually select a stream which may be overwritten by the adaptive function shortly afterwards.

The screen flashes black whenever the stream is changed, can be annoying as the streams change very frequently some times. The screen sometimes remains black for a few seconds and then the video resumes and fast-forwards to catchup with the audio, which usually doesn't cut out.

Overall it seems very good, thanks for the great work.

bibitocarlos:
Installed on Archlinux, Matrix. Used your personal branch (https://github.com/glennguy/inputstream.adaptive/tree/rework-new)
Nothing else to say, it works, will report back if i find a bug.
One thing : it seems that the buffer is not displayed on the video OSD screen (not sure, i have put a 3 min buffer, and i cant see the gray bar ahead of the black bar, will try with a 15 min buffer).

glennguy:
@kontell Thanks for the feedback. I'll note the issue with not being able to force a manual selection. This PR is for the minimum viable product i.e it works, and most users will want to just leave it on default. WRT the screen blanking on Android - good to know, it works well on Windows with only a tiny amount of resyncing so maybe there's some platform specific issues with stream switching. The issue may lie on Kodi's side of the border

@bibitocarlos thanks as well for trying it out. The gray bar on the OSD is reporting Kodi's internal 8 second buffer for streams and is unaware of the buffer internal to inputstream.adaptive. Maybe in would be a feature that could be added to the inputstream API to query buffer size and fullness. Perhaps something useful that could be added on this side would be some debug logging of buffer health.

phunkyfish:
> @bibitocarlos thanks as well for trying it out. The gray bar on the OSD is reporting Kodi's internal 8 second buffer for streams and is unaware of the buffer internal to inputstream.adaptive. Maybe in would be a feature that could be added to the inputstream API to query buffer size and fullness. Perhaps something useful that could be added on this side would be some debug logging of buffer health.

Adaptive does not support the times interface for inputstreams yet. I.e. ` bool GetTimes(kodi::addon::InputstreamTimes& times)` and related methods. I had made some initial attempts quite a while back: https://github.com/xbmc/inputstream.adaptive/pull/349 and https://github.com/xbmc/inputstream.adaptive/pull/348. Would be worthwhile revisiting this after this PR is merged and it really would cleanup the OSD for users if implemented correctly.

phunkyfish (approved):
I won't pick on any nits in this review as you plan to refactor the whole codebase after this is merged.

Nice work, and great to see this going in ;)

Vikassm73:
Does local buffering is really implemented in Inputstream

Without Inputstrem

ffmpeg

With Inputstream
Inputstream

CastagnaIT:
kodi core currently dont have an implementation to show internal buffer of binary addons to the GUI
so you cant view the ISA buffering in any place of to the GUI

Vikassm73:
It does not show in progress bar also like "without Inputstream"? Reason is, sometimes, I face buffering problem in  case of Inputstream. Is there any specific settings to implement it?

Progress

CastagnaIT:
no that buffering show in your picture is the Kodi VP buffer and not the ISA buffer
as said ISA buffer is already implemented, but you cant show it in kodi GUI then its hidden,
to allow show ISA buffer in to kodi GUI is required Kodi core changes, and atm these changes are not planned

(assuming that its not a streaming provider problem) if depends from your network that is not capable to fullfill the VP+ISA buffer then you can force limit bitrate from ISA settings and/or set a stream selection type "ask" in to ISA settings to select manually a lower bitrate

if you have buffering problems with a specific manifest type, please open a new Issue by filling the template to provide all required info

Vikassm73:
In normal case, while you are facing problem from ISP, if you pause for some 10-15 minutes and play. It start working smoothly but in case of Inputstream, it play for 1-2 second and then again start buffering no matter how long you pause the video. Mostly using MPD link.

phunkyfish:
I believe that adaptive could implement the iTimes interface. I tried to do it a couple of years back. I’m sure the PR is still about somewhere.

CastagnaIT:
>  if you pause for some 10-15 minutes and play. It start working smoothly but in case of Inputstream, it play for 1-2 second and then again start buffering no matter how long you pause the video

i dont have encoutered this behaviour, there are many types of manifests then so without knowing you can't get any help,
if you think a problem you should open a new Issue, provide not only the log+manifest, but also if possible a sample stream or a way to replicate the problem

please stop use PR's pages like this to talk about problems, these are development threads only, Issue threads are done for them

PR #734, opened 2021-07-13
[Android] Use multiple WV sessions

### LINKED ISSUE #693: Audio causes video stuttering on L1 Android device
@Testato 
Did you open an issue for the Android stuttering?
I wonder if it does it if you go into Inputstream Adaptive Settings and set Media to "Video"
You won't get Audio - but interested to see if that fixes it

I have seen with HBO Max, EC3 (dolby digital+) audio causes the video to stutter etc
which made me think, maybe there is an issue decoding some audio streams with secure decoder

_Originally posted by @matthuisman in https://github.com/xbmc/inputstream.adaptive/issues/647#issuecomment-843703642_

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/693

as per peaks comment here:
https://forum.kodi.tv/showthread.php?tid=350092&pid=2908844#pid2908844

This fixes pixalation issues on some streams mainly on Sony android tvs.
Seems the issue appears when both audio and video are encrypted

Test build (Android AARCH64)
https://jenkins.kodi.tv/job/xbmc/job/inputstream.adaptive/job/PR-734/5/artifact/cmake/addons/build/zips/inputstream.adaptive+android-aarch64/inputstream.adaptive-2.6.22.zip

Test build (Android ARMV7)
https://jenkins.kodi.tv/job/xbmc/job/inputstream.adaptive/job/PR-734/5/artifact/cmake/addons/build/zips/inputstream.adaptive+android-armv7/inputstream.adaptive-2.6.22.zip

### DIFF
--- a/wvdecrypter/wvdecrypter_android_jni.cpp
+++ b/wvdecrypter/wvdecrypter_android_jni.cpp
@@ -487,9 +487,9 @@ std::vector<char> WV_CencSingleSampleDecrypter::GetChallengeData()
 
 bool WV_CencSingleSampleDecrypter::HasLicenseKey(const uint8_t *keyid)
 {
-  // We work with one session for all streams.
-  // All license keys must be given in this key request
-  return true;
+  // true = one session for all streams, false = one sessions per stream
+  // false fixes pixaltion issues on some devices when manifest has multiple encrypted streams
+  return false;
 }
 
 void WV_CencSingleSampleDecrypter::GetCapabilities(const uint8_t *keyid, uint32_t media, SSD_DECRYPTER::SSD_CAPS &caps)


matthuisman:
initial results look promising:

1. https://forums.whirlpool.net.au/thread/3wwnqx63?p=69#r1363
2. https://github.com/xbmc/inputstream.adaptive/issues/693#issuecomment-879855841
3. https://github.com/matthuisman/slyguy.addons/issues/28
4. Another user via email confirmed this fixed pixelation on 7plus widevine vod content on their sony tv

So i think this solves the problem of having both encrypted audio and video streams on android causing pixalation / buffering.
It doesn't seem to solve the issue of some audio streams (usually 5.1) causing video to buffer
https://github.com/xbmc/inputstream.adaptive/issues/693#issuecomment-878962095

matthuisman:
@CastagnaIT 
be interested to know if / how this affects your DRM init code?
as with this on Android, we can have multiple DRM sessions for each stream (audio / video)

CastagnaIT:
do not think "preinit" opens only one session without any check

but in your commit should be better if you add a comment somewhere for this particularity,
there is a lot of code uncommented make difficult to understand the things

matthuisman:
I'd love to add a comment if I knew exactly what it was doing. I simply going off peaks suggestion in the linked forum. 

Maybe I'll just put "return false to allow multiple sessions for different streams" 

Be handy if you could test with Netflix with the init stuff. Just to make sure it doesn't affect that

CastagnaIT:
yes i will do

CastagnaIT:
Tried on my smartphone played 3 videos without problems the data there are

glennguy:
I'm not sure if I know the right answer to this, but should we have a setting for this behaviour? Peak3d did mention that the tradeoff is streams may be slower to start up due to the need to create multiple DRM sessions in Android which would affect all users, not just ones affected by the issue this solves.

matthuisman:
I havn't noticed any delayed start up
Pretty sure more reliable streams should be more important than 0.5s delay?

Adding a setting just means more users reporting the issue and then needing
to figure out that's how they fix it. So would need to be default to true
at least. Then what's the point?

I'm sure we could make up the slight delay (if any) by improving other code
:)

CastagnaIT:
In my test i did not notice any startup slowdowns noticeable
and my phone is a Honor 8 is a bit grandfather..
but is always possible ask to others do others tests

matthuisman:
I don't mind adding a setting. But I actually think we need less settings, not more. As IA is used by other add-ons, it's not really that obvious to end users that it has its own settings.

Therefore, as less settings and best defaults is always best I reckon.

We are doing a good job if users never need to adjust the settings :)

When adaptive comes in, then that should remove need for users to muck around with bandwidth settings as well .

glennguy:
@matthuisman all good points, totally agree :)

glennguy (approved):

matthuisman:
:)
I still need to add the comment. I'll do this 2moro

UPDATE:
Comment added
Ready to go in

phunkyfish (approved):
👍

PR #737, opened 2021-07-15
[DEPS] Bento4 update

### LINKED ISSUE #730: Memory leak in AP4_Movie::GetPsshAtoms
Hi there. I've found out a possible memory leak at the line

https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/lib/libbento4/Core/Ap4Movie.cpp#L115

The definition of the property is at the line 
https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/lib/libbento4/Core/Ap4Movie.h#L79

This will actually cause a memory leak since the destructor of `AP4_Array` class will only free the allocated `AP4_PsshAtom*` pointers and not the actual `AP4_PsshAtom`.

A really simple fix could be to change the property definition

```C
AP4_Array<AP4_PsshAtom> m_PsshAtoms;

// and the get method
AP4_Array<AP4_PsshAtom>& GetPsshAtoms()   { return m_PsshAtoms; }
```
remove the new statement

```C
m_PsshAtoms.Append(AP4_PsshAtom(*pssh_item->GetData())); 
// or
m_PsshAtoms.Append(*pssh_item->GetData()); 
```

change where the method is used
https://github.com/xbmc/inputstream.adaptive/blob/a675541a5c10e0ff4e0ec2fa7b4c9d8306bcffb8/src/main.cpp#L2322-L2334

```C
AP4_Array<AP4_PsshAtom>& pssh = movie->GetPsshAtoms();

// and all pssh[i]-> to ppsh[i].
```

What do you think? Are there anything that I'm missing? If the memory leak is legit, should I make a PR?

Thank you.

CastagnaIT:
~This PR is intended as a test~

Commit: Updated Bento4 sources to 8a2bf30
switch the current Bento4 sources to the original (last) version from https://github.com/axiomatic-systems/Bento4

This also include the leak fix #730

i have tried on Windows and LibreElec with ntfx add-on and seem ok
but i have not done extensive testing


CastagnaIT:
perhaps i have found the problem for android builds
is needed to set the `ANDROID` define to the CMakelists.txt ? (seem related to the dynamic cast error)
https://github.com/CastagnaIT/inputstream.adaptive/blob/bento4_update/lib/libbento4/Core/Ap4Config.h#L122

Maven85:
I have fixed the compiler errors on Android through this:
https://github.com/Maven85/inputstream.adaptive/commit/836cd7776cfd300225cca0217f46585de561b008

But I have no idea whether that is correct.  Whatever.  I've been using this version for a few days, so far with no problems.

CastagnaIT:
Thanks! i think it is correct these are referred to new files added from sources
maybe they (bento dev) forgot? to add these defines
i try push it

CastagnaIT:
@Maven85 well done🎉
what addon have you used for tests?

Maven85:
I tested the following addons:
Amazon
Netflix
DAZN

CastagnaIT:
Sounds good, i will wait @glennguy tests to have green light

future steps:
-splitting patches as originals one
-add leak fix #730

integrate external fetch & patch system
i don't think I know enough about how makefile works to to it,
but for now i can prepare these commits to be converted into patches files

CastagnaIT:
i have splitted commits, where possible i have kept the original commits headings and authors

@babyraging can you check if your leak fix is correctly applied?

Maven85:
@CastagnaIT can you please rebase this PR

CastagnaIT:
@Maven85 done

matthuisman:
we have now branched to Nexus so should be able to get some more of these feature PRs in.
I'll try test this over next few days

CastagnaIT:
I should update the basic bento4 commit again some bugfixes have been added
I'm asking them what they think of our patches, let's see if we can upstream some of these so we have fewer patches to include

CastagnaIT:
Good news seem they have accepted to include in upstream some of our changes,
this can give us advantages in the future to find a more suitable solution to embed this component avoiding to keep whole code in our repository

matthuisman:
i found everything works as it should with this code with my tests
Should we get it in @CastagnaIT @glennguy ?
then we can all start using the latest master code builds and hopefully start catching any bugs etc from all the prs

CastagnaIT:
I don't know if they have started to take our patches upstream, may still take several days
however, at least here we have the list of all our patches cleaned up
It's ok to me if you want to merge it

matthuisman:
I'm just thinking even if they do upstream, the result in this PR will be the same right?

CastagnaIT:
if they include our upstream patches
all those accepted by them will be removed from my current list of this PR
but at the end should result in same code

lib/libbento4/Core/Ap4Results.h
```diff
@@ -55,8 +55,6 @@ const int AP4_ERROR_INVALID_TRACK_TYPE              = -19;
 const int AP4_ERROR_INVALID_RTP_PACKET_EXTRA_DATA   = -20;
 const int AP4_ERROR_BUFFER_TOO_SMALL                = -21;
 const int AP4_ERROR_NOT_ENOUGH_DATA                 = -22;
-const int AP4_ERROR_NOT_ENOUGH_SPACE                = -23;
-const int AP4_ERROR_STREAM_CHANGED					= -24;
```
Maven85, line 59: @CastagnaIT 
You need to add "AP4_ERROR_STREAM_CHANGED" again as this is what is used here: https://github.com/xbmc/inputstream.adaptive/blob/26f21455a9427b5aa829297c5c7a5b9eb095af53/src/main.cpp#L226

CastagnaIT: I think it's a bad idea to add an unknown constant to the upstream code,
Looking in the upstream code it seems that currently that there is no a specific error check in the ReadPartial method
with the exception of `AP4_ERROR_EOS`,
this suggests that something should be handled differently in our code, or as alternative use `AP4_ERROR_READ_FAILED` for both cases

CastagnaIT: @glennguy take a look

phunkyfish:
If we are going to update bento can we please use the binary addon depends system for it? It was designed to allow adding libraries based on known source versions and supports patches. Directly including source files is the wrong way to do it IMO.

As you already have the work done, it should be easy enough to change it. @AlwinEsch can also help with any guidance.

CastagnaIT:
i do not think to have skills to do it, i have tried to unserstand cmake/deps but it is beyond my capabilities...
if someone else has the skills, they can take over this PR or create a new one

phunkyfish:
If we still need patches are we using the kodi depends strcytre

> i do not think to have skills to do it, i have tried to unserstand cmake/deps but it is beyond my capabilities...
> if someone else has the skills, they can take over this PR or create a new one

Do you have the patches separate to the original source and the original source tarball? If so we could move to merge this and migrate it to depends afterwards?

CastagnaIT:
i have prepared these patches based on our repo only,
but if required i can prepare/adapt our patches to be used with original repo https://github.com/axiomatic-systems/Bento4

phunkyfish:
> i have prepared these patches based on our repo only, but if required i can prepare/adapt our patches to be used with original repo https://github.com/axiomatic-systems/Bento4

Ok, depends works taking a known release, i.e. from here: https://github.com/axiomatic-systems/Bento4/releases (e.g. the latest one: https://github.com/axiomatic-systems/Bento4/releases/tag/v1.6.0-639) and then it applies a set of patches against that known version. 99% of the work is preparing the patch set. For the cmake parts I'm sure @AlwinEsch can help out if I can't manage it.

Did you use master of bento or a known release version?

CastagnaIT:
When i reworked patches the new version had not yet been released then i have used the master branch,
but now it is possible to use the last tag release version mentioned without problem, there are no relevant changes on the code

CastagnaIT:
i have created the git patches files [bento4_patches.zip](https://github.com/xbmc/inputstream.adaptive/files/7250268/bento4_patches.zip)
from: https://github.com/CastagnaIT/Bento4/commits/kodi (sources upstream with tag release v1.6.0-639)

these patches not include:
* commit "Renamed wrong filenames": seems not more needed
* commit "Fix include paths": i do not know if it is needed with the new dep building system, in case it is the only one that needs to be added.
* commit "Updates for new Bento4": OFC need to be applied separately on ISA branch

phunkyfish:
Great, thanks!

lib/libbento4/Core/Ap4Results.h
glennguy: @CastagnaIT agreed, the constants for the most part aren't checked specifically, only whether a result is `AP4_SUCCESS` or `AP4_FAILED` (not `AP4_SUCCESS`). Maybe alter to  `AP4_FAILURE` or as you suggested?

CastagnaIT: i think keep `AP4_ERROR_READ_FAILED` is appropriate others similar implementations of ReadPartial return this specific error when fails

glennguy: Sounds good

matthuisman:
@phunkyfish Any updates on this?

I guess switching to the pulling upstream source and applying patches could be done in the future? Wouldn't matter if this PR went in the meantime?

phunkyfish:
> @phunkyfish Any updates on this?
> 
> I guess switching to the pulling upstream source and applying patches could be done in the future? Wouldn't matter if this PR went in the meantime?

Correct, up to @glennguy when it goes in. The end goal will be the same.

glennguy (approved):
@CastagnaIT Thanks for your hard work on this and for the patch files!

@matthuisman @phunkyfish yes agree, we'll get this in now and rework it as a dependency later

CastagnaIT:
i have rebased to Nexus

lib/libbento4/Core/Ap4Results.h
CastagnaIT: add the change

phunkyfish (approved):

PR #748, opened 2021-07-17
Remove version from README file as it makes automation across binary addons awkward

phunkyfish:
I propose removing the version from the README for a simple reason. No other binary add-on follows this convention. I would really prefer to not customise the automation created for Weblate to include this. Added to that it is very easy to see the latest release at the root of the repo under releases.

Thoughts?

I did not spot this till after performing two merges for Weblate language updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# inputstream.adaptive (2.6.18)
+# inputstream.adaptive
 
 This is an adaptive file addon for kodi's new InputStream Interface.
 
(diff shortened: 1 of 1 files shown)


michaelarnauts:
I don't think the version should be in the README.

Otherwise, you could use a GitHub badge that shows the version of the latest release.

Like we did here: https://github.com/add-ons/plugin.video.vtm.go/blob/master/README.md

phunkyfish:
Nice, let me update the PR to add the badge.

phunkyfish:
Do you know if the latest release badge can be done per branch @michaelarnauts?

Probably better to not show the Matrix version on Leia...

Maybe I will leave this PR as it is and we can find a better solution later.

glennguy (approved):
Sounds very sane to me

PR #761, opened 2021-07-29
[Dash] Add suggestedPresentationDelay

### LINKED ISSUE #647: Add support for suggestedPresentationDelay

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/647

This PR adds in a new live_delay_ manifest tree variable.
This tracks the delay / offset we want to keep from the live head.
I also updated the seeking code to use this value to stop seeking ahead of this point

With this PR, we will use suggestedPresentationDelay from a Dash manifest for this value.
Also we make sure this value is above ~~12s~~ 16s as anything lower breaks playback.

I've added a sanity check to the seek time.
If live_delay_ is greater than the total time of the stream, it stops us seeking a negative time.
start stream doesnt need such a check as that code already just grabs segment 0 if pos is < 0

removed the uint64_t overallsecs line as not actually used...

~~Oh, I found removing the preceeding = true in the seek code 
fixed a bug when seeking sometimes causing the stream to start pixelating.
Not sure why it is there. changing preceeding seems like a bad idea.
We are simply adjusting our seek amount.~~

TEST BUILDS:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-761/16/artifacts

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -284,10 +284,9 @@ bool AdaptiveStream::start_stream()
           if (!pos)
             pos = 1;
         }
-        //go at least 12 secs back
         uint64_t duration(current_rep_->get_segment(pos)->startPTS_ -
                           current_rep_->get_segment(pos - 1)->startPTS_);
-        pos -= static_cast<uint32_t>((12 * current_rep_->timescale_) / duration) + 1;
+        pos -= static_cast<uint32_t>((tree_.live_delay_ * current_rep_->timescale_) / duration);
         current_rep_->current_segment_ = current_rep_->get_segment(pos < 0 ? 0 : pos);
       }
       else // switching streams, align new stream segment no.
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -59,6 +59,7 @@ namespace adaptive
     , stream_start_(0)
     , available_time_(0)
     , base_time_(0)
+    , live_delay_(0)
     , minPresentationOffset(0)
     , has_timeshift_buffer_(false)
     , has_overall_seconds_(false)
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -466,7 +466,7 @@ public:
   uint32_t currentNode_;
   uint32_t segcount_;
   uint32_t initial_sequence_ = ~0UL;
-  uint64_t overallSeconds_, stream_start_, available_time_, base_time_;
+  uint64_t overallSeconds_, stream_start_, available_time_, base_time_, live_delay_;
   uint64_t minPresentationOffset;
   bool has_timeshift_buffer_, has_overall_seconds_;
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2295,6 +2295,10 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
 
   drmConfig_ = config;
 
+  // Always need at least 16s delay from live
+  if (adaptiveTree_->live_delay_ < 16)
+    adaptiveTree_->live_delay_ = 16;
+
   return InitializePeriod();
 }
 
@@ -3051,13 +3055,19 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
   // don't try to seek past the end of the stream, leave a sensible amount so we can buffer properly
   if (adaptiveTree_->has_timeshift_buffer_)
   {
+    double maxSeek(0);
     uint64_t curTime, maxTime(0);
     for (std::vector<STREAM*>::const_iterator b(streams_.begin()), e(streams_.end()); b != e; ++b)
       if ((*b)->enabled && (curTime = (*b)->stream_.getMaxTimeMs()) && curTime > maxTime)
         maxTime = curTime;
-    if (seekTime > (static_cast<double>(maxTime) / 1000) - 12)
+
+    maxSeek = (static_cast<double>(maxTime) / 1000) - adaptiveTree_->live_delay_;
+    if (maxSeek < 0)
+      maxSeek = 0;
+
+    if (seekTime > maxSeek)
     {
-      seekTime = (static_cast<double>(maxTime) / 1000) - 12;
+      seekTime = maxSeek;
       preceeding = true;
     }
   }
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1068,6 +1068,8 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       }
       else if (strcmp((const char*)*attr, "availabilityStartTime") == 0)
         dash->available_time_ = getTime((const char*)*(attr + 1));
+      else if (strcmp((const char*)*attr, "suggestedPresentationDelay") == 0)
+        AddDuration((const char*)*(attr + 1), dash->live_delay_, 1);
       else if (strcmp((const char*)*attr, "minimumUpdatePeriod") == 0)
       {
         uint64_t dur(0);
--- a/src/test/TestDASHTree.cpp
+++ b/src/test/TestDASHTree.cpp
@@ -527,3 +527,9 @@ TEST_F(DASHTreeTest, AdaptionSetSwitching)
 
   EXPECT_EQ(tree->periods_[0]->adaptationSets_[4]->representations_[0]->id, "8");
 }
+
+TEST_F(DASHTreeTest, SuggestedPresentationDelay)
+{
+  OpenTestFile("mpd/segtpl_spd.mpd", "https://foo.bar/segtpl_spd.mpd", "");
+  EXPECT_EQ(tree->live_delay_, 32);
+}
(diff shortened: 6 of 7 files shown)


matthuisman:
oh, Ill add some tests

matthuisman:
Eventually, we will use this new variable in combination with the cache ahead support.
example: User sets cache ahead to 2minutes
We would set live_delay_ to 120s.
This then gives the caching thread enough segments ahead to cache.

When we do that, I'd suggest a new function in main.cpp called "CalculateLiveDelay"
In this, we would add in the cache ahead seconds etc.
And make sure it's over 12s.
Then call this function from where the current 12s code is

eg.
```
void Session::CalculateLiveDelay()
{
  // Always need at least 12s delay from live
  if (adaptiveTree_->live_delay_ < 12)
    adaptiveTree_->live_delay_ = 12;

  adaptiveTree_->live_delay_ += kodi::GetSettingInt("CACHESECONDS");
}
```

matthuisman:
dont think tests are possible with this due to what we need to test is set in main.cpp

matthuisman:
@glennguy 
ready for review.

Can only do a basic test to make sure live_delay gets set from a dash manifest.
I wanted to check the fallback to 12s works, but that code is in main.cpp in function our tests dont call

Only thing slightly concerned about is removing the preceeding = true
maybe you can try to work out why it was there in the 1st place.
seems to cause issues

matthuisman:
also, I noticed when I was writing tests, mediaPresentationDuration value would be used instead of timeShiftBuffer.

mediaPresentationDuration
"If present, it SHALL accurately match the duration between the zero point on the MPD timeline and the end of the last period, "
https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#timing-and-presentation-types

No way a live stream can know the end of its last period..
It only makes sense with VOD, but with VOD we calculate the total time from adding up the segments.

matthuisman:
when i was writing tests I noticed an issue here:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/test/manifests/mpd/segtimeline_live_pd.mpd

IA was setting the overalseconds (total available time to timeshift) to 9000seconds!
However, it's timeShiftBufferDepth is only 18seconds.

The issue is IA is parsing the mediaPresentationDuration and using that instead of timeShiftBufferDepth 
https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#timing-and-presentation-types

We should always use timeShiftBufferDepth for live content.
For VOD content, overalseconds is calculated by adding up the segments anyway.

Therefore, the 2nd commit removes mediaPresentationDuration

src/common/AdaptiveStream.cpp
```diff
@@ -187,10 +187,10 @@ bool AdaptiveStream::start_stream(const uint32_t seg_offset,
       if (!pos)
         pos = 1;
     }
-    //go at least 12 secs back
     uint64_t duration(current_rep_->get_segment(pos)->startPTS_ -
                       current_rep_->get_segment(pos - 1)->startPTS_);
-    pos -= static_cast<uint32_t>((12 * current_rep_->timescale_) / duration) + 1;
+
+    pos -= static_cast<uint32_t>((tree_.live_delay_ * current_rep_->timescale_) / duration);
```
glennguy: Probably still the the `+ 1` here. Previously it has been that 12 seconds back gives us the segment which *finishes* 12 seconds back from the edge. Without the +1 we're starting the stream 1 segment forward from where it would have been before.

glennguy (changes requested):

src/common/AdaptiveStream.cpp
matthuisman: oops, your right. i totally missed that!

glennguy:
Can't test main.cpp yet but you could get a test in there to check `AdaptiveStream::start_stream` is setting current_segment_ to what it should.

src/common/AdaptiveStream.cpp
matthuisman: thinking about this, the seek code only limits to the tree_.live_delay_
(originally the 12 was hard-coded here)
https://github.com/xbmc/inputstream.adaptive/pull/761/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2L3048

so the seeking is letting us get an extra segment forward
I think we want this to be consistent.
Also, if an add-on starts with a resume time of a large number (to get latest live period), it will also do the same

the seek code only deals with the seek time as well (not segments)
so not as easy to -1 segment from that

once the cache ahead is in, live_delay_ will be larger to begin with as well which should at least get that 1 segment back

so I think we could either

1. Increase the 12s to maybe 16s to hopefully get that extra segment for most streams..
2. Add a "stand-in" cache ahead value like i said in a CalculateLiveDelay function that adds Xs that again will get us back at least that segment.

but i think it's important that the seeking and start playback both have the same offset
which currently they actually don't

glennguy: > but i think it's important that the seeking and start playback both have the same offset
which currently they actually don't

I can't get a stream to play closer than 12 seconds from the edge. Try Kayo for example as the 4s segments divide neatly.

Looking here, we get our '-1':
https://github.com/xbmc/inputstream.adaptive/blob/bce1270c2544aaa30a33545c4e24f5228b9dd590/src/common/AdaptiveStream.cpp#L544-L545

matthuisman: not sure if thats the same thing.
I think its just correcting it if the calculated segment start time is ahead of the seek time.
we would want a specific -1 for live streams only in this code somewhere.

I'm trying with latest stock IA.
Redbull TV live stream from IPTV AU.
We want to make sure where we start playback is same offset as when we ffwd.
Ignores 12s. just look at the timeline and what the maximum seconds get too.
When I start playback, it will start around 0:41 - 0:42
and the highest it goes is 0:47
Now if you mash ffwd a few times, you can get it to being playback further ahead than this
eg. starting at 0:47 and playing upto 0:52 or even further
which seems to me to be a segment ahead than initial playback
Its segments at 6s
Also, if you keep trying fwd enough on windows - you may see the bug I mentioned where the stream becomes distorted at the top. 

Now, with this version, it will start playback around 0:48 and get to around 0:52
makes sense as no longer the -1.
But when i fwd, I pretty much always get the same spot or slightly back.

Actually, I know a possible solution!
We know the segment length in start stream.
So, we could simply add 1x segment length to live_delay_
That will give you the same -1 but it will be consistent
Oh wait - start stream is called multiple times, so it will keep increasing it...
We would need to track an actual .live_delay_ and then a new adjusted_delay_

eg.

tree_.adjusted_delay_ = tree_.live_delay_ + (current_rep_->timescale_ / duration)
pos -= static_cast<uint32_t>((tree_.adjusted_delay_ * current_rep_->timescale_) / duration);

then use adjusted_delay_ in the seek code

glennguy: Ok, I'm seeing it as well. I can get to the 2nd segment from the edge with 6s segments.

Just thinking out loud here but when we start a stream we're always going to be consistent - we've just refreshed the playlist, and don't refresh again until a segment has elapsed. However when we seek, we're not necessarily performing it on the exact boundary of a segment.
Eg. stream with 10s segments. Stream is started/playlist downloaded just as a new segment has been added to the playlist. 8 seconds in we seek, playlist is refreshed and returns the same content/segments (which is now valid for another 10s), and we see the window in the timeline that's been played 'slide' to the right. So with this theory of course we can get 'closer' to the live edge...

Anyway, regarding the += seg length - we don't always have consistent segment length. HLS can vary as can some DASH configurations. 

I'm thinking maybe that changing the default from 12 is best left for Nexus/rework since we're changing the default behavior? We would be best to make it the greater of SPD and 12 seconds. In rework we can also factor in a 'preferred buffer' setting.

matthuisman:
@glennguy 
what needs to be done to get this in (its my last open PR yah!)

glennguy:
Let's increase the 12 second default. Since on start_stream we're no longer having the +1 segment maybe 16 seconds would be a good target? What do you think?

Also regarding `preceeding`, it fixes some problems but creates others (in my testing). It would be better to find out more about what this variable actually means, and then correcting the logic in adaptivestream.cpp to deal with the result in a way that makes all seek operations work correctly.

matthuisman:
sounds good & fully agree :)
Changes have been done now

So, with the new buffer, will our delay be 16s + buffer length?
Or will the buffer get segments from within that 16s?

I've opened a new issue to discuss the proceeding issue
https://github.com/xbmc/inputstream.adaptive/issues/817

glennguy:
> Or will the buffer get segments from within that 16s?

Yes this. Segments will buffer all the way to the live edge but we can't seek into this area because it will prevent stream switching (design does not allow for streams to switch when only 1 segment in the buffer) and users won't realise that it removes any benefit to be had from buffering. Of course if they're viewing 16s from the live edge then the buffer can only ever be that big regardless of the settings. That's something we'll need to think about - should the minimum buffer setting increase this 16s? And what if the timeshift buffer window is only say 20s long? Or 40s long etc...

glennguy (approved):

glennguy:
Regarding the changes with preceeding - I see your point, if you want to include the change here and can't find issues with ts or fmp4 streams then then I'm good with that.

matthuisman:
think maybe best i just do a separate PR to remove if after this is merged?
then easier to test with and without etc.

phunkyfish (approved):

PR #762, opened 2021-07-30
[Dash] Always use timeShiftBufferDepth for live

matthuisman:
currently if a live manifest has both timeShiftBufferDepth  and mediaPresentationDuration, IA will set the overalseconds to mediaPresentationDuration.

For live, we should always use timeShiftBufferDepth  as mediaPresentationDuration is only telling the total time not the time available to timeshift

also removed the overallsecs line as thats not actually used

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1065,12 +1065,14 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
     if (!mpt)
       mpt = tsbd;
 
-    AddDuration(mpt, dash->overallSeconds_, 1);
+    // Always use timeShiftBufferDepth for live manifest
+    if (dash->has_timeshift_buffer_)
+      AddDuration(tsbd, dash->overallSeconds_, 1);
+    else
+      AddDuration(mpt, dash->overallSeconds_, 1);
+
     dash->has_overall_seconds_ = dash->overallSeconds_ > 0;
-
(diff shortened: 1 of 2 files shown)


glennguy:
I think this code might cause problems as there's these point from DASH docos:
> MPD@timeShiftBufferDepth MAY have an arbitrarily large value, including a value greater than the total duration of periods in the presentation.

and

> MPD@timeShiftBufferDepth: optional, but recommended

Is there any issues that come from the current code?

matthuisman:
well maybe then we take the lowest value of either as the overalseconds?
it's this value we using to tell kodi how far back we can rewind.

glennguy:
Yes `overallSeconds_` gives Kodi the duration and this is currently an effective timeshift buffer, but `timeShiftBufferDepth` isn't necessarily going to be the same value as the duration of the segments in the mpd.

> it's this value we using to tell kodi how far back we can rewind.

It's actually going to tell us how far forward we can go from the first segment. As in #710 if you have a TSB that is shorter than the duration of segments, it won't be the start of the stream that you can't seek to, it will be the end/live edge.

We will end up with this:
image

What we really want is to see the proper duration of the mpd in Kodi, but in cases where TSB differs from the total duration that we can't seek *backwards* past the TSB:
image

Similar to you PR #761 we should be able to stop the user from seeking into this area.

matthuisman:
Going to close this. Your right, I'm trying to solve an issue we haven't actually seen causing issues yet. Something to keep in mind if we hear of any issues about too large or short Kodi timeline

PR #770, opened 2021-08-01
[Decrypter] undefine LOCLICENSE so .challenge, .init etc files are not written to disk

matthuisman:
not sure the need for these 4 files that are written.
probably for debugging back in the day.

IMO we shouldn't be writing to disc anymore than needed.
Devices like RPI SD cards have certain amount of writes blah.

It may also increase playback speed of WV content very slightly
More so on devices with slow write speeds (mechanical drive etc)

The "widevine" folder that is created I think is still needed for when persistent storage is enabled

### DIFF
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -33,7 +33,7 @@
 #error  "WIDEVINECDMFILENAME must be set"
 #endif
 
-#define LOCLICENSE 1
+//#define LOCLICENSE
 
 using namespace SSD;
 
--- a/wvdecrypter/wvdecrypter_android.cpp
+++ b/wvdecrypter/wvdecrypter_android.cpp
@@ -36,7 +36,7 @@ using namespace SSD;
 
 SSD_HOST *host = 0;
 
-#define LOCLICENSE 1
+//#define LOCLICENSE
 
 void Log(unsigned int loglevel, const char *format, ...)
 {


glennguy:
These files can be useful in debugging issues with users, like with #812 
Maybe it might be worth thinking about instead putting this behind a IA setting?

matthuisman:
be good if we could detect kodis log level and only enable for debug level

Also, i think thats the 1st time i've seen these requested...
Pretty rare to need them.
Even when we get sent them, i really dont see them helping :(

glennguy:
I don't think we can detect the log level...
I've seen peak3d ask for them once before as part of debugging.
How about we disable for now and if needed in the future we can put behind an expert setting?

wvdecrypter/wvdecrypter.cpp
```diff
@@ -33,7 +33,7 @@
 #error  "WIDEVINECDMFILENAME must be set"
 #endif
 
-#define LOCLICENSE 1
+//#define LOCLICENSE
```
glennguy, line 36: Leave both uncommented but set to 0

wvdecrypter/wvdecrypter_android_jni.cpp
```diff
@@ -36,7 +36,7 @@ using namespace SSD;
 
 SSD_HOST *host = 0;
 
-#define LOCLICENSE 1
+//#define LOCLICENSE 0
```
glennguy: as above

glennguy (changes requested):

wvdecrypter/wvdecrypter.cpp
matthuisman: sorry, im a c++ noob.
the checks later are
#ifdef LOCLICENSE
To me that just says if it's defined - not check it value?
So, if I set to 0, its still defined isn't it?

think i'd need to change the checks to #if
which checks if its equal to true

glennguy: Yep you're right, sorry!
Just to make it consistent then have them both 1 and commented out. i.e changes are just to add `//`

matthuisman: i just changed it to use if's instead, haha.
ill switch it back :)
and ill remove the 1 as thats lead to our confusion above and not needed

matthuisman:
ready

glennguy (approved):

phunkyfish (approved):

spacys:
> Maybe it might be worth thinking about instead putting this behind a IA setting?

Good idea, I would prefer this option.

PR #772, opened 2021-08-10
[Subtitles][WebVTT] Implemented format type check

CastagnaIT:
Implemented missing signature validating check with/without optional BOM
ref.: https://www.w3.org/TR/webvtt1/

### DIFF
--- a/src/parser/WebVTT.cpp
+++ b/src/parser/WebVTT.cpp
@@ -18,7 +18,26 @@
 
 #include "WebVTT.h"
 #include "../helpers.h"
+#include "../log.h"
 #include <cstring>
+
+constexpr const char* signatureCharsBOM = "\xEF\xBB\xBF\x57\x45\x42\x56\x54\x54";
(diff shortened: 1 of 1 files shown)


glennguy (approved):
Looks good to me, thanks @CastagnaIT

src/parser/WebVTT.cpp
```diff
@@ -18,8 +18,27 @@
...
+constexpr char signatureLastChars[] = {'\x0A', '\x0D', '\x20', '\x09'};
+
+static bool ValidateSignature(const char* data, const char* signature)
+{
+  if (strlen(data) > strlen(signature))
+  {
+    if (strncmp(data, signature, strlen(signature)) == 0)
+    {
+      // Check if last char is valid
+      if (strchr(signatureLastChars, data[strlen(signature)]) != NULL)
```
phunkyfish: All of the functions should be called C++ way, so fully qualified `std::strlen` etc.

phunkyfish (changes requested):

phunkyfish (approved):
👍

Issue #798, opened 2021-09-13
Kodi 19 Matrix / Kodi 20 Nexus - Version change inform


AlwinEsch:
**As this is the most used addon from Kodi, I am creating this issue to inform you about the following changes.**

Because the addon is not yet a branch for Kodi Nexus, it must be created.
Therefore, the default branch will also be changed and all pull requests will be swapped to it, so if a change to the Kodi Matrix version is to be made, a backport request must also be created again.

In addition, from now on the major versions of the addons will be created to match the associated Kodi version.
As a result, there will be a 19.0.0-Matrix release and a 20.0.0-Nexus release for this.

Unfortunately, these changes are absolutely necessary in order to have it correctly usable as an addon on the current Kodi Nexus on all platforms!

After the branches have been created, I leave this issue open for a short time to understand the background.

AlwinEsch:
The new releases with 19.0.0-Matrix and 20.0.0-Nexus are now there, all requests (expept 1) which was set before to Kodi Matrix are swapped now to Kodi Nexus.

The one where not https://github.com/xbmc/inputstream.adaptive/pull/791, there it can maybe closed as a version increase about not match now.

CastagnaIT:
i think now can be closed

PR #800, opened 2021-09-13
[Nexus] initial release change for new Kodi 20 Nexus version

AlwinEsch:
This change the builds to Kodi Nexus.

Further is the version to 20.0.0 increased to have equal to Kodi and to see on which Version this addon works.

See also issue #798.


glennguy (approved):
LGTM, thanks @AlwinEsch!

phunkyfish (approved):

Issue #773, opened 2021-08-10
[WebVTT][Dev-Question] Moving WebVTT parser to Kodi sources


CastagnaIT:
When the rendering transition to LibASS will be merged in kodi (currently opened Kodi WIP PR)
I would like to move the WebVTT parser to Kodi source parsers

this would allow some improvements like
- implementing support to text positioning
- implementing styling metadata support
- add the support to kodi to load external ".vtt" files

there are two problems that i require info, maybe someone hame more info than me:

1) **How to allow Kodi FFMPEG to identify the ISA subtitles as WebVTT**

currently ISA always send `SRT` string to kodi FFMPEG
https://github.com/xbmc/inputstream.adaptive/blob/2.6.23-Matrix/src/TSReader.cpp#L81-L83
https://github.com/xbmc/inputstream.adaptive/blob/2.6.23-Matrix/src/TSReader.cpp#L135
Here kodi ffmpeg translate the SRT codec name string to the ffmpeg codec ID:
https://github.com/xbmc/xbmc/blob/19.1-Matrix/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp#L387-L397
then kodi the subtitles always as "subrip" subtitles

for this reason we should be add `webvtt` to the enum (first link^^).
I have tried track the code to try understand where the enum value for the SRT enum is picked,
and i found that come from `/lib/mpegts/tsDemuxer.cpp` i think from this method:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/lib/mpegts/tsDemuxer.cpp#L917-L987
but i dont understand `desc_tag` byte code, where does it come from?
could a new hex code be added to identify a new format?

2) **After resolved step 1, pass the stream file to kodi, maybe for initial test can be reused current implementation with some change to send the header metadata before start sending the subtitles lines**

time permitting, I am planning this change for the future
but for now, I am only gathering information on how this could be done

glennguy:
> currently ISA always send SRT string to kodi FFMPEG

The 2 links below only relate to `ts` files which are only in HLS. IA can support `INCLUDEDSTREAM`s for the case where HLS has `ts` files that include video *and* audio, but subtitles are not implemented.

For fmp4, it will get set here: https://github.com/xbmc/inputstream.adaptive/blob/851d1c49c038704d106ffd8889ffb2c95e39524f/src/main.cpp#L2815-L2817
We set as `srt` for any case it seems.

[...]

PR #816, opened 2021-10-05
Fixed missing no-data check on DemuxRead

CastagnaIT:
While working on a new WebVTT implementation (#773) i experienced of a ISA crash,
this happens when there is no data to copy to the demuxer,
and does not happen in the Matrix branch

Comparing the Matrix branch this check has been removed in Nexus:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/main.cpp#L3892-L3901

@glennguy found the problem try look

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3852,7 +3852,7 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
     const AP4_UI08* pData(sr->GetSampleData());
     DEMUX_PACKET* p;
 
-    if (iSize && pData && sr->IsEncrypted())
+    if (sr->IsEncrypted() && iSize > 0 && pData)
     {
       unsigned int numSubSamples(*((unsigned int*)pData));
       pData += sizeof(numSubSamples);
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -3852,7 +3852,7 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
     const AP4_UI08* pData(sr->GetSampleData());
     DEMUX_PACKET* p;
 
-    if (iSize && pData && sr->IsEncrypted())
+    if (sr->IsEncrypted() && iSize > 0 && pData)
```
glennguy, line 3855: no change is needed here I think - `iSize > 0` check is redundant as it's the same as checking `iSize`

src/main.cpp
```diff
@@ -3871,13 +3871,16 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
       p = AllocateDemuxPacket(iSize);
 
-    p->dts = static_cast<double>(sr->DTS() + m_session->GetChapterStartTime());
-    p->pts = static_cast<double>(sr->PTS() + m_session->GetChapterStartTime());
-    p->duration = static_cast<double>(sr->GetDuration());
-    p->iStreamId = sr->GetStreamId();
-    p->iGroupId = 0;
-    p->iSize = iSize;
-    memcpy(p->pData, pData, iSize);
+    if (iSize > 0 && pData)
```
glennguy, line 3874: Previously the condition was just `if (iSize)` - is the extra condition check on `pData` necessary? Not trying to nitpick, were you getting crashes without it?

src/main.cpp
CastagnaIT: numerical variables should be compared in this way,
i was specifically requested to use this rule in the kodi source code
there are different rules between components?

glennguy: No that's fine, I'm not up to speed yet on all the style guidelines. Disregard my comment :)

glennguy (approved):

glennguy:
@phunkyfish can I get your tick when you've got a moment?

matthuisman (approved):

PR #820, opened 2021-10-10
[WebVTT] Moved WebVTT management to Kodi core

CastagnaIT:
The goal is to remove the dummy parser from ISA, and thus provide full subtitle support from Kodi core.

This is part of Kodi PR: https://github.com/xbmc/xbmc/pull/20244
and must be merged only when the mentioned PR will be merged in Kodi master.

~ATM the above mentioned PR is not ready and not updated with my local sources, to continue i am forced to wait for the merge of a another third PR, so i leave as a draft in case of early reviews.~

This change has been tested in two ways:

* For WebVTT text format type case with:
Net***x addon (data as file)
Disn€y+ addon (segmented data)

* For WebVTT "MP4" (ISO/IEC 14496-30:2014) format type case with:
http://media.axprod.net/ExoPlayer/Captions2/Manifest.mpd

Ref. discussion issue #773


matthuisman:
pretty cool!
Any idea if TTML will be able to do the same?

CastagnaIT:
certainly possible, the hardest part is creating the parser in Kodi from scratch as i have done with this

CastagnaIT:
i have make the PR as ready for review
but in any case we have to wait the merge of PR https://github.com/xbmc/xbmc/pull/20244
in the meantime, now you are able to test this change, because i have finally updated PR https://github.com/xbmc/xbmc/pull/20244

CastagnaIT:
Now this could be merged, the PR of webvtt parser in Kodi has been merged
but i suggest to find and fix the problem of #836 before merge this,
because there is the possibiliti that this complicate things to verify the cause of the current ISA bug due to the different management of the parser

CastagnaIT:
@Essam3152 please open an Issue on Kodi repo, i do not want to spam this PR for not related things
Yes the previous workaround works but not follow the W3C specs, then this require a better investigation on how to proceed

Essam3152:
Sorry, I will delete my comments in this PR.

glennguy (approved):
Thanks for this @CastagnaIT !

src/main.cpp
```diff
@@ -1621,9 +1640,14 @@ class ATTRIBUTE_HIDDEN SubtitleSampleReader : public SampleReader
   bool GetInformation(kodi::addon::InputstreamInfo& info) override { return false; };
   bool TimeSeek(uint64_t pts, bool preceeding) override
   {
-    if (m_codecHandler->TimeSeek(pts / 1000))
-      return AP4_SUCCEEDED(ReadSample());
-    return false;
+    if (dynamic_cast<WebVTTCodecHandler*>(m_codecHandler))
```
phunkyfish, line 1642: I would suggest brackets for the if as well as the else for readability.

phunkyfish (approved):

Issue #687, opened 2021-05-16
[HLS] TS Byterange fails to play


matthuisman:
Main manifest: https://pastebin.ubuntu.com/p/npZ8KT95NZ/
Audio / video manifest: https://pastebin.ubuntu.com/p/W6vdwb3PZV/

test strm: https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_ts_byterange/ia.strm

debug log: 
[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6488904/kodi.log.txt)

sometimes even crashes kodi

matthuisman:
interesting, I've found if the byte range starts at 0, then IA plays it OK

Plays ok:
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_ts_byterange_from_zero/ia.strm
manifest: https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_ts_byterange_from_zero/sub.m3u8

doesnt play:
https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_ts_byterange/ia.strm
manifest: https://raw.githubusercontent.com/matthuisman/ia_tests/master/hls_ts_byterange/sub.m3u8

matthuisman:
ok, think I found the issue
https://github.com/xbmc/inputstream.adaptive/blob/master/src/common/AdaptiveStream.cpp#L63
if I set that to 0 - it'll play fine.

matthuisman:
ok, i can make it work if I comment out

https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/common/AdaptiveStream.cpp#L59
and https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/common/AdaptiveStream.cpp#L213

It seems that check is thinking range_begin_ is referring to a timestamp?
as per here: https://github.com/xbmc/inputstream.adaptive/blob/Matrix/src/common/AdaptiveTree.h#L138

So, it's actually just a coincidence that EXT-X-BYTERANGE starting at 0 are working.
As that is setting absolute_position_ to 0 which is what is required

i think both those statements need to be updated to also skip TS or byterange streams..

glennguy:
Yes I think you're completely right. The implementation was the first implementation of fmp4 for hls (separate segment fmp4 didn't happen until disney+) and the init segment is always at 0 anyway:
[...]

PR #831, opened 2021-10-20
Set read position for byteranged TS streams to 0

glennguy:
Fixes byteranged TS streams.

As documented in https://github.com/xbmc/inputstream.adaptive/issues/687

AFAIK there's no real difference between a TS file split into chunks or one that is retrieved with ranged requests. Each chunk has it's own init data bundled with it, this change will ensure that byteranged files are read the same as the chunks are.

Tested on several streams, no issue or regression found.


phunkyfish (approved):

matthuisman (approved):
tested fine :)
thank you @glennguy

Issue #832, opened 2021-10-20
[Dash] Live stream 'craziness' due to audio tracks(s)


matthuisman:
Same provider, ESPN and ESPN2.
(slyguy proxy off or on same issue)

ESPN2 - no issues

ESPN will either
start at negative time with audio but no video
or
start at a time above end of stream with audio but no video

With either, if you seek fwd or back - the image will come back but no audio.
Also, if you try to ffwd to live edge - the stream will stop like it think it's the end.

If i set IA to Video only - it correctly starts at live head with no timeline weirdness
If I set IA to Audio only - it correctly starts at live head with no timeline weirdness 

So only issue when using both Video & Audio
Both ESPN and ESPN2 manifests attached as well as screenshots.
they are very similar so hopefully that can help us find the different and narrow down issue

I also had some Hulu live channels that would play at live head OK, but no audio.
The "fix" for those was to make Hulu only return a single audio track.
I suspect that could be related to this.
This issue could be related to https://github.com/xbmc/inputstream.adaptive/issues/821

[espn-2-GOOD.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/7385513/espn-2-GOOD.mpd.txt)
[espn-channel-BAD.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/7385514/espn-channel-BAD.mpd.txt)
[espn-2-GOOD-less-logging.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/7393206/espn-2-GOOD-less-logging.mpd.txt)
[espn-channel-BAD-less-logging.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/7393208/espn-channel-BAD-less-logging.mpd.txt)
**(both the above were generated about 2 seconds apart)**

[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/7384853/kodi.log.txt)

![screenshot00002](https://user-images.githubusercontent.com/6225961/138183880-4872052c-bdb3-4819-9567-926b6b36e7a1.png)
![screenshot00003](https://user-images.githubusercontent.com/6225961/138183876-d7ce9913-5f18-425b-8a46-f00d647d6778.png)

matthuisman:
Possibly this is fixed with pr
https://github.com/xbmc/inputstream.adaptive/pull/913

CastagnaIT:
it doesnt happen anymore?

matthuisman:
[...]

PR #834, opened 2021-10-25
Fix wrong condition in PTSToElapsed

CastagnaIT:
From what i know unsigned int cannot handle negative values
if that is the purpose of this condition it is wrong

these kinds of bugs are annoying because they are hidden and cause only bad works with conditions,
i have not checked all the code if unsigned int variables are handled correctly i hope there are no others

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2942,12 +2942,14 @@ uint64_t Session::PTSToElapsed(uint64_t pts)
 {
   if (timing_stream_)
   {
-    uint64_t manifest_time = (pts - timing_stream_->reader_->GetPTSDiff() > 0)
-                                 ? pts - timing_stream_->reader_->GetPTSDiff()
-                                 : 0;
-    return (manifest_time > timing_stream_->stream_.GetAbsolutePTSOffset())
-               ? manifest_time - timing_stream_->stream_.GetAbsolutePTSOffset()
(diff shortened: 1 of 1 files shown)


glennguy:
@CastagnaIT sounds sane to me, thanks for catching this.
@matthuisman could this be behind #832?

glennguy (approved):

src/main.cpp
```diff
@@ -2942,7 +2942,7 @@ uint64_t Session::PTSToElapsed(uint64_t pts)
 {
   if (timing_stream_)
   {
-    uint64_t manifest_time = (pts - timing_stream_->reader_->GetPTSDiff() > 0)
+    uint64_t manifest_time = (static_cast<int64_t>(pts) - timing_stream_->reader_->GetPTSDiff() > 0)
```
phunkyfish: Precedence rules apply here, and this is correct. However adding an extra set of brackets to be explicit to the reader that the subtraction occurs before the comparison is a good idea.

Alternatively calculating the value for the comparison on the line before is also an option and may be preferred.

phunkyfish (approved):

matthuisman:
@glennguy 
doesnt fix https://github.com/xbmc/inputstream.adaptive/issues/832 unfortunately

PR #841, opened 2021-11-21
Fix sub streams being unable to re-start

### LINKED ISSUE #836: [Kodi 20][Subtitles] Disn€y+ subtitles broken
### Describe the bug

I am doing tests with "Ribelle - the brave" movie but i think also other movies are same

When you start (or resume) the video the subtitles not works at first start,
if you try switching to another track apparently is ok, for the first time,
if you switch another time subtitles are full broken

I have tried investigate but i am not able to find where the real problem come from

To note that if you force install the ISA binary of Kodi 19, to Kodi 20, the subtitles works as expected

## Expected Behavior

### To Reproduce

1. Open "Ribelle - the brave"
2. (only once) Play it a first time and choose the CC subtitles in order to save it as preferred, then stop it
4. Play it
5. No subtitles shown

### Debuglog

https://paste.kodi.tv/ocokocupum.kodi

### MPD/M3U8s/ISM

### Screenshots 

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Operating system version/name: Windows 10
 - Kodi version: Kodi 20 nightly
 - inputstream.adaptive version: Last from repository

glennguy:
## Description

When we switch bandwidths for the video stream (send `DEMUX_SPECIALID_STREAMCHANGE` packet to Kodi) the video and audio streams are simply opened again, however with subtitles the current stream is closed/disabled first - see [here](https://github.com/xbmc/xbmc/blob/7478f983b67dc1e3b09e061acf2eaa675b0f0170/xbmc/cores/VideoPlayer/VideoPlayer.cpp#L930).
In Matrix whenever streams are opened the segment number passed in always directs the AdaptiveStream to start from the beginning (first segment), Kodi then directs IA to seek the stream to the current location (calls `PosTime`).

For subtitles however PosTime isn't called, and we have the 'leftovers' from the last time the stream was running - `segment_read_pos_` and `current_segment_` in particular. This change will restore the Matrix behaviour by resetting these values when `stop()` is called on the AdaptiveStream (which only happens on `DEMUX_SPECIALID_STREAMCHANGE` for subtitle streams as per above)

I don't know if this is the correct solution, I think ideally VideoPlayer would ask the subtitle streams to seek as well as audio/video, but this may affect things outside of inputstream add-ons.... @CastagnaIT since you're much more familiar with subtitles within Kodi you may have an idea or opinion on this.

## Motivation and context
Fixes #836 

## How has this been tested?
Tested on https://playertest.longtailvideo.com/adaptive/elephants_dream_v4/index.m3u8

Will (have not tested yet) fix D+ subtitles and other HLS/webvtt segmented subtitles 

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -870,9 +870,13 @@ void AdaptiveStream::info(std::ostream& s)
 
 void AdaptiveStream::stop()
 {
-  if (current_rep_)
+  if (current_rep_) 
+  {
     const_cast<adaptive::AdaptiveTree::Representation*>(current_rep_)->flags_ &=
         ~adaptive::AdaptiveTree::Representation::ENABLED;
+    current_rep_->current_segment_ = 0;
+  }
+
   if (thread_data_)
   {
     StopWorker(STOPPED);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1967,6 +1967,7 @@ void Session::STREAM::disable()
   if (enabled)
   {
     stream_.stop();
+    stream_.Reset();
     reset();
     enabled = encrypted = false;
   }


CastagnaIT:
> I don't know if this is the correct solution, I think ideally VideoPlayer would ask the subtitle streams to seek as well as audio/video

i think the subtitle "seek" could be needed, although not for all type of subtitles:
with fmp4 subtitle data seem handled in different way not sure that is needed,

formats like webvtt/ttml when they are segmented may be need the seek, but currently all segments are downloaded in full before starting the video and not during the playback, why i do not know... at least on how i see with D+ that use segmented webvtt, i suspect there is poorly implemented, need to be investigated

instead formats provided like "single file" (again with webvtt/ttml) not need the seek, (in my webvtt PR in fact i have changed the seek call as no-op)
EDIT: OFC currently for TTML is needed, until it will be moved to Kodi

but if we change/fix behaviour of some of these, we will need to adapt parsers in kodi
or we'll have the risk of rendering duplicate subtitles (due to missing libass data flush)
(this when ISA parsers will be transferred to Kodi)

glennguy:
I've been thinking more about this.
If Kodi was to call PosTime when enabling/disabling sub streams then video and audio would seek as well, making a pretty bad user experience which explains why that doesn't happen at the moment.
Therefore in my mind this is the best way forward for now. Until there is a way in the inputstream interface to tell a subtitle stream where to seek to then we will have the current behavior of downloading all segments in a stream (not sure about Kodi 19 behavior on this and there's only 4 segments in the test stream so of course it will buffer them all with v20). It would be a big effort to save a (usually) small amount of potentially unnecessary data transfer happening.

Thoughts? If you're happy then accept the changes and I'll get phunkyfish in for his thoughts, and we can also tick off your PR.

CastagnaIT:
just a note in my previous comment, i have not mentioned TS subtitles, at moment i do not know how are handled.

for me ok we can go on
for my PR you can read my last messages in the issue #836 but if you want to proceed with the merge of my PR there is no problem for me, we know that is needed a solution/alternative in Kodi for the DEMUX_SPECIALID_STREAMCHANGE that currently is cause of side effects

CastagnaIT (approved):

src/common/AdaptiveStream.cpp
```diff
@@ -870,9 +870,13 @@ void AdaptiveStream::info(std::ostream& s)
 
 void AdaptiveStream::stop()
 {
-  if (current_rep_)
+  if (current_rep_) 
+  {
     const_cast<adaptive::AdaptiveTree::Representation*>(current_rep_)->flags_ &=
-        ~adaptive::AdaptiveTree::Representation::ENABLED;
+      ~adaptive::AdaptiveTree::Representation::ENABLED;
```
phunkyfish: 4 space indent for a continuation

glennguy: Fixed

phunkyfish (approved):

Issue #785, opened 2021-08-31
Android (esp. Nvidia Shield) - Seeking causes video freeze


popy2k14:
### Describe the bug
Here is a clear and concise description of what the problem is:

When trying a seek during video playback of the pvr zattoo addon, the video freezes in-place and stays stuck. Sometimes it will skip to correct place, but can be up to a 20+ second delay. Other times the skip doesn't complete and Kodi must be force-closed. Most of the time, audio playback is running but picture freezes.

There is a similar issue here: xbmc/xbmc#19713
and here: https://github.com/rbuehlma/pvr.zattoo/issues/126
(where we already talked about this issue)

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

Picture should not freeze.

## Actual Behavior

Picture freezes on seeking.

## Possible Fix

no

### To Reproduce

What i have done is:
1. Factory reset shield tv pro 2019
2. Install kodi 19.1
3. Install zattoo pvr addon and set it up
4. enable debug log and exit
5. restart kodi
6. started ORF1 and seeked 1 minute backwards -> picture freezed, audio was running
7. switched to Pro7 and seeked 1 minute backwards -> picture freezed, audio was running

### Debuglog

The debuglog 2.6.15 can be found here: https://paste.kodi.tv/xezuwinewi.kodi
Sadly kodi has'nt updated IA when i made the log.
After that i realised that. Can provide an log with 2.6.23, if needed.
Just let me know.

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:

### Screenshots 
Here are some links or screenshots to help explain the problem:

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

### Your Environment
Used Operating system:

 - [X ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Android TV 9
 - Kodi version: 19.1
 - inputstream.adaptive version: 2.6.15 & newest 2.6.23

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

matthuisman:
please get a new log with latest Inputstream Adaptive 2.6.23

Also, could you see if either of these IA builds fix the issue
They are suitable for Kodi 19 on Nvidia shield

[...]

PR #848, opened 2021-12-07
[Matrix] Fix seek accuracy / Remove duplicate call to SeekTime

glennguy:
## Description

* In `PosTime` the call to `Session::SeekTime` is unnecessarily made twice, this change addresses that.
* Seeking made more accurate for audio/sub streams

## Motivation and context

As can be observed in #785 seeking a stream in Kodi has the AdaptiveStreams and stream readers seek twice in a row - first to the requested location, then again 'on the spot'. ~~Undesired behaviour seems to stem from this on some platforms (Android in particular) possibly due to differences in how the data from the stream is consumed (suspect that inbetween calls some data is read and on second call one of the streams positions is then 'rounded up', causing misalignment).~~

ISA has a designated 'timing stream' (timing master) - this is normally the video stream if there is one and there will sometimes be a difference in pts between eg. audio and video streams. 
The current behavior when seeking is (has been documented in code comments previously and in this PR as best I can) that for each of the enables AdaptiveStreams, the AdaptiveStream will seek to the requested segment, then followed by stream's 'reader' to seek within the segment. The issues seen in #785 have been from this process running and arriving at different segments for audio/video. ~~While Windows seems to handle this fine, Android does not.~~ - not the cause of seeking issues

In part of this mechanism of seeking in `Session::SeekTime` here:
https://github.com/xbmc/inputstream.adaptive/blob/1d8877960ccf33f423b6d18ee6f9456fc9301fc8/src/main.cpp#L3099-L3104
the `seekTimeCorrected` variable is being changed to where the video stream actually ended up seeking to (i.e the start of a segment, rather than an arbitrary time that was selected on the timeline). What has been happening though on the next iterations of this function is the pts offset of the AdaptiveStream being iterated is used to calculate the next segment seek here:
https://github.com/xbmc/inputstream.adaptive/blob/1d8877960ccf33f423b6d18ee6f9456fc9301fc8/src/main.cpp#L3081-L3085 ~~which causes issues because this offset differs from the video stream offset.~~

Since Kodi only has one timeline for playback and not multiple for each component (video/audio/sub) it then follows that we should be using the same pts offset for each of these calculations in order to end up in the same place. Having the audio stream for example a couple of milliseconds forward of the video stream will cause the `AdaptiveStream::seek_time` function to advance a whole segment ahead, causing the misalignment.

The change therefore is to always use the timing stream pts offset when available for the seek_time operation.

## How has this been tested?

Runtime tested on Win 10 using YouTube live videos. Previously it could be observed that different segment numbers were seeked to on some videos at some times (although with no observable issue in playback on Windows), now the segments always align.

Attempt to resolve issues of #785 - update: does not.

## Types of change

- [] **Bug fix** (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3079,10 +3079,11 @@ bool Session::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
       if (!(*b)->reader_->IsStarted())
         StartReader((*b), seekTimeCorrected, ptsDiff, preceeding, false);
       // advance adaptiveStream to the correct segment (triggers segment download)
-      if ((*b)->stream_.seek_time(
-              static_cast<double>(seekTimeCorrected - (*b)->reader_->GetPTSDiff()) /
-                  STREAM_TIME_BASE,
-              preceeding, bReset))
+      int64_t ptsDiff = timing_stream_ ? timing_stream_->reader_->GetPTSDiff() :
+          (*b)->reader_->GetPTSDiff();
+      if ((*b)->stream_.seek_time(static_cast<double>(seekTimeCorrected - ptsDiff) /
+                                      STREAM_TIME_BASE,
+                                  preceeding, bReset))
       {
         if (bReset)
           (*b)->reader_->Reset(false);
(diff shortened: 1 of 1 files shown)


glennguy:
Closing this to split the changes between 19 and 20 (duplicate call 19, seek accuracy 20)

Issue #853, opened 2021-12-11
Kodi 20: IA crashing with HGO streams on Android


enen92:
Inputstream adaptive/Kodi is crashing in Kodi 20 with all streams from HGo (HBO) in Android (works on my other test machines which have linux). They play well under kodi 19 so its definitely some change in IA for nexus.
Can't tell much about the crash, the backtrace in logcat shows this:

```
12-11 19:19:16.481  8492  8492 D NowPlayingManager: onMediaDataUpdated: A notification for Now Playing Card has been posted with NowPlayingData [title=100 Foot Wave - 1 SEASON, EPISODE 2, artist=Unknown]
12-11 19:19:16.516 11679 11679 I chatty  : uid=10076(org.xbmc.kodi) expire 5 lines
12-11 19:19:16.594  3107  3107 I /system/bin/tombstoned: received crash request for pid 11644
12-11 19:19:16.649 11679 11679 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
12-11 19:19:16.650 11679 11679 F DEBUG   : Build fingerprint: 'Xiaomi/oneday/oneday:9/PI/3595:user/release-keys'
12-11 19:19:16.650 11679 11679 F DEBUG   : Revision: '0'
12-11 19:19:16.650 11679 11679 F DEBUG   : ABI: 'arm'
12-11 19:19:16.650 11679 11679 F DEBUG   : pid: 11495, tid: 11644, name: Thread-14  >>> org.xbmc.kodi <<<
12-11 19:19:16.650 11679 11679 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
12-11 19:19:16.650 11679 11679 F DEBUG   : Cause: null pointer dereference
12-11 19:19:16.650 11679 11679 F DEBUG   :     r0  00000000  r1  fffffff9  r2  ccd24510  r3  00000000
12-11 19:19:16.650 11679 11679 F DEBUG   :     r4  ccd24510  r5  00000000  r6  ced68ca4  r7  ced68c6c
12-11 19:19:16.650 11679 11679 F DEBUG   :     r8  f1bf9d8c  r9  75756964  r10 c9785d00  r11 ccd244b8
12-11 19:19:16.651 11679 11679 F DEBUG   :     ip  c55ea7d4  sp  ccd244b0  lr  c5554844  pc  c5555a50
[...]

PR #854, opened 2021-12-13
Test CastagnaIT fix for segfault (ism)

matthuisman:


matthuisman:
didnt work unfortunately https://github.com/xbmc/inputstream.adaptive/issues/853#issuecomment-992418654

PR #862, opened 2021-12-15
Internal bento

wsnipex:
Adds an option to build bento4 internally, analog to INTERNAL_FMT/SPDLOG... in kodi core.
This allows debian packaging to supply the bento source zip.

We still need a way to have jenkins download the bento source zip file to fully fix the PPA builds. Any ideas?


glennguy (approved):
Looks sane to me, thank you for the effort in resolving this.

phunkyfish (approved):
Amazing thank you. So when we do something similar for the mpegts library is there a specific change we need to make other than adding the Tarball to the Jenkins file?

wsnipex:
in theory adding the dependency in the Jenkinsfile and debian/rules should be enough

phunkyfish:
> in theory adding the dependency in the Jenkinsfile and debian/rules should be enough

We'll add you as a reviewer once we have mpegts ready. Should allow us to skip any broken builds for linux next time.

PR #864, opened 2021-12-23
[Nexus] API related update

AlwinEsch:
Related to xbmc/xbmc#20705 and xbmc/xbmc#20710

Runtime test with Netflix was OK.


Acidzero2020:
@AlwinEsch 

https://github.com/xbmc/inputstream.adaptive/blob/0d1e529a7d8227315b2d89cef10608f895280a1f/src/main.cpp#L66
Is it possible that you forgot to update AddonLog => ADDON_LOG

AlwinEsch:
Is updated and also included commit to fix C++17 build by Windows UWP.

phunkyfish (approved):

PR #873, opened 2022-01-11
On pre-init DRM still use the session opened

CastagnaIT:
This change keep opened the drm session opened from the pre-init, and goes through the manifest request,
this will be applied to non-Android systems. Android mediadrm not allow to add KeyId to the session, therefore the session will continue to be closed and reopened a new one.

-Why
New changes to ntfx website reveals more info and new problems
now i can see that in the response data obtained from the "Licensed manifest" request, contains the manifest data, plus, the license data (all in one),
this is not a problem because i can split the response as two data and provide them to ISA without any problems via proxy

but the problem is that when i provide the license data to ISA, and the drm session is different (because previously closed), ISA fails to process license data, then playback of video is not possible

I initially suspected this, but I wasn't sure because it worked without problems
this because we used to make a separate request for the license, but now everything is included in the manifest.

I think, however, that website is still in the transition phase between old and new methods because there is a bit of a mix.

This needs a backport to Kodi 19

Currently tested on:
- Windows 10 64 bit with latest Kodi 20 nightly
- CoreElec Kodi 20 nightly (ARM)

### DIFF
--- a/src/SSD_dll.h
+++ b/src/SSD_dll.h
@@ -42,7 +42,7 @@ namespace SSD
     {
       PROPERTY_HEADER
   };
-    static const uint32_t version = 15;
+    static const uint32_t version = 16;
 #if defined(ANDROID)
     virtual void* GetJNIEnv() = 0;
     virtual int GetSDKVersion() = 0;
--- a/src/common/AdaptiveDecrypter.h
+++ b/src/common/AdaptiveDecrypter.h
@@ -8,6 +8,9 @@
 
 #include <bento4/Ap4.h>
 
+#include <stdexcept>
+#include <string_view>
+
 enum class ENCRYPTION_SCHEME
 {
   NONE,
@@ -26,6 +29,22 @@ public:
     m_SkipBlocks = static_cast<uint32_t>(skipBlocks);
   };
   virtual void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme){};
+
+  /*! \brief Add a Key ID to the current session
+   *  \param keyId The KID
+   */
+  virtual void AddKeyId(std::string_view keyId)
+  {
+    throw std::logic_error("AddKeyId method not implemented.");
+  };
+
+  /*! \brief Set a Key ID as default
+   *  \param keyId The KID
+   */
+  virtual void SetDefaultKeyId(std::string_view keyId)
+  {
+    throw std::logic_error("SetDefaultKeyId method not implemented.");
+  };
 
 protected:
   uint32_t m_CryptBlocks{0};
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1838,6 +1838,7 @@ bool Session::Initialize(const std::uint8_t config, uint32_t max_user_bandwidth)
 
   // Preinitialize the DRM, if pre-initialisation data are provided
   std::map<std::string, std::string> additionalHeaders = std::map<std::string, std::string>();
+  bool isSessionOpened{false};
 
   if (!m_kodiProps.m_drmPreInitData.empty())
   {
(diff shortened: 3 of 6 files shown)


glennguy (approved):
Looks good!

glennguy:
let me know when you're happy with this and I'll bring it in, NF is the only thing using this right so low risk

CastagnaIT:
seems that other changes are needed
yesterday i have tested better and i have see that after 2 minuts of playback the image freeze and the decrypter fails
https://paste.kodi.tv/sevuvagizi.kodi

glennguy:
Is this worth looking at again now?
I can see from the log that we're sending a dummy KID for pre-init, and just an idea but maybe this causes the issue?? - any way to see how NF web does it with EME logger or similar?

CastagnaIT:
yes dummy kid/pssh i think is causing the problem
~but how update the kid/pssh in current session with the one provided from manifest?~
I am not sure of my affirmation either...from the last log the key used is not the dummy

i have tried to see the eme log but i am not able to understand nothing
the only thing highlighted when the playback start is a keystatuseschange event where seem that the keys are changed
[EMELogFile.txt](https://github.com/xbmc/inputstream.adaptive/files/8262040/EMELogFile.txt)

glennguy:
Seems to me that decryption doesn't even happen until 2 minutes in, the first 2 minutes may be in the clear? EME log doesn't show keys arriving until 2 minutes has passed, compared to almost straight away with bitmovin test stream (using chrome/eme logger).

I can see from your EME log the exact same dummy KID is being used so that shouldn't be the issue.

What we are doing though is keeping that dummy KID in our WV_CencSingleSampleDecrypter instance. What I think you should  try is creating a method to be able to update m_defaultKeyId from main.cpp (would have to use AdaptiveDecrypter.h I think) and then in InitializeDRM if we know that we've preinitialized, update cdmsessions[1].ssd with the correct default KID. From there the SendMesssage calls should hopefully bring us back a working key, worth a shot at least....

CastagnaIT:
i have tried roughly your proposed changes to add the kid manually and works!
played 6 minutes without problems!
i will try prepare the changes better way (also i haven't looked at the android side yet) i will need to do more tests

CastagnaIT:
Ok updated commit with a bit of cleanups,
from what i understood on android is not possible add kid's to the session, the only way is close and open a new session, thinking of using mediadrm `restorekeys` seem not appropriate since this is for offline content.

However at least for now i could use pre-init only on non-android systems, and on android use the old way, since the Ntfx add-on has been always a bit hackered on android. If in the future stop working on android i will consider how to act when the time comes...

if you have other suggestions are welcome, meantime i will need to do other tests before remove WIP flag

CastagnaIT:
ok i have tested for some days and i dont find any problem with playback i will prepare also the backport

PS. sorry for the confusion in the changes, but i have fixed all wrong spaces in the code

glennguy:
Do they do the whole fake KID thing on Android NF app to get the manifest as well? And does NF android app keep the same SID for playback? Maybe it doesn't matter as much to them for Android as the platform is more secure...

I'll bring this in, looks good and as you say we can deal with Android NF issues later on if they come up. Great work :)

CastagnaIT:
N€tflix is a really obsessive with protections...it is not easy to understand how it works on android
the native app uses different endpoints/calls on android than the website, for now using preinit or not on Android makes no difference

i should install an android VM to make proxy works on android to itercept the https calls to check what data has the manifest request and try doing again some reverse eng. , but long time ago i had to reset my smartphone lost the things and now i have also forgotten how to do it

for now i don't want to waste my time on this again... we'll see when it breaks
thanks for merge i will do the backport next days

Issue #870, opened 2022-01-06
Some HLS WebVTT subtitles doesn't work


Dis90:
### Describe the bug
Here is a clear and concise description of what the problem is:

WebVTT subtitles in some HLS streams doesn't work.

(Case 1) On Kodi 19 and Kodi 20 these doesn't work. Kodi 19 doesn't show subtitles and on Kodi 20 video playback doesn't start.
[prog_index.txt](https://github.com/xbmc/inputstream.adaptive/files/7824316/prog_index.txt)
[caption_0.txt](https://github.com/xbmc/inputstream.adaptive/files/7824395/caption_0.txt)
[caption_1.txt](https://github.com/xbmc/inputstream.adaptive/files/7824318/caption_1.txt)
[master.txt](https://github.com/xbmc/inputstream.adaptive/files/7824319/master.txt)
Log Kodi 19: 
[kodi19.log](https://github.com/xbmc/inputstream.adaptive/files/7824329/kodi19.log)
Log Kodi 20:
[kodi20.log](https://github.com/xbmc/inputstream.adaptive/files/7824410/kodi20.log)

(Case 2) These does work on Kodi 19 but doesn't work on Kodi 20 (subtitles is listed in subtitles menu)
[subs.txt](https://github.com/xbmc/inputstream.adaptive/files/7824363/subs.txt)
[caption_0.txt](https://github.com/xbmc/inputstream.adaptive/files/7824365/caption_0.txt)
[master_new.txt](https://github.com/xbmc/inputstream.adaptive/files/7824366/master_new.txt)
Log Kodi 19:
[kodi19_new.log](https://github.com/xbmc/inputstream.adaptive/files/7824372/kodi19_new.log)
Log Kodi 20:
[kodi20_new.log](https://github.com/xbmc/inputstream.adaptive/files/7824428/kodi20_new.log)

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

Subtitles to work.

### To Reproduce

1. 
2.
3.

### Debuglog

The debuglog can be found here:

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:

### Screenshots 
Here are some links or screenshots to help explain the problem:

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [x] OSX
[...]

PR #875, opened 2022-01-15
[WebVTT] Fix sending segments packages

CastagnaIT:
We are talking about segmented WebVTT on HLS streams,
until now all the packages (of each segment) were being sent in the kodi demuxer just before the playback starts,
in general this could be not a big problem perhaps that is why it has never been fixed,
but the problem will be raised when a stream have a lot of segments

A case is on issue #870 (see case 1) where the stream have more than 1000 segments,
this means download and processing more than 1000 files before the playback with a huge delay some minutes before the video can start, a clearly useless effort...

The purpose of this fix is set the right PTS on each segment in order to download and send to the demuxer the packets during playback and not before start the playback

This also revert the changes introduced with PR #841 no longer needed

This PR needs also a change on Kodi parser to handle better the flush of data PR https://github.com/xbmc/xbmc/pull/20858

#### HLS test streams used (the first link for the PR, the others to verify no regressions)
WebVTT Segmented (30min - 60 segments): https://s3.amazonaws.com/_bc_dml/example-content/bipbop-advanced/bipbop_16x9_variant.m3u8
WebVTT "as single file": http://dash.edgesuite.net/akamai/test/caption_test/ElephantsDream/elephants_dream_480p_heaac5_1.mpd
WebVTT fMP4 ISO/IEC: http://media.axprod.net/ExoPlayer/Captions2/Manifest.mpd

Sidenote in the segmented test stream, the subtitles are not displayed immediately this is due to X-TIMESTAMP-MAP i think that need some other change, that not concern this PR

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -696,26 +696,24 @@ bool AdaptiveStream::seek(uint64_t const pos)
   return false;
 }
 
-bool AdaptiveStream::getSize(unsigned long long& sz)
+bool AdaptiveStream::retrieveCurrentSegmentBufferSize(size_t& size)
 {
   if (state_ == STOPPED)
     return false;
 
   std::unique_lock<std::mutex> lckrw(thread_data_->mutex_rw_);
 
-  if (ensureSegment())
+  while (true)
   {
-    while (true)
+    if (worker_processing_)
     {
-      if (worker_processing_)
-      {
-        thread_data_->signal_rw_.wait(lckrw);
-        continue;
-      }
-      sz = segment_buffers_[0].buffer.size();
-      return true;
+      thread_data_->signal_rw_.wait(lckrw);
+      continue;
     }
+    size = segment_buffers_[0].buffer.size();
+    return true;
   }
+
   return false;
 }
 
@@ -877,7 +875,6 @@ void AdaptiveStream::stop()
   {
     const_cast<adaptive::AdaptiveTree::Representation*>(current_rep_)->flags_ &=
         ~adaptive::AdaptiveTree::Representation::ENABLED;
-    current_rep_->current_segment_ = 0;
   }
 
   if (thread_data_)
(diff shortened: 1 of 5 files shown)


glennguy (approved):

src/common/AdaptiveStream.cpp
```diff
@@ -696,26 +696,24 @@ bool AdaptiveStream::seek(uint64_t const pos)
   return false;
 }
 
-bool AdaptiveStream::getSize(unsigned long long& sz)
+bool AdaptiveStream::getSegmentBufferSize(size_t& size)
```
phunkyfish: I always find names like this strange. For me a getter should return what it gets. a function name like `segmentBufferSizeRetrieveable(size_t& size)` works better when used for flow control. But maybe it's just me.

phunkyfish (approved):
Just a comment really on function names when used for flow control. But nothing blocking for this PR. Nice work.

src/common/AdaptiveStream.cpp
CastagnaIT: maybe i prefer better as `getCurrentSegmentBufferSize` since should get data from the actual segment
@glennguy opinion?

glennguy: I think what phunkyfish is saying is that having the name start with get makes it seem like it's following convention of naming getter/setter functions - implies that the function should return some value or object, not a success/fail. His suggestion goes along with naming the function based on return type but I also see your point that it's assigning a value to the input parameter by 'getting' it from the segment....

I'm not too fussed either way and won't complain if it goes in as-is and we refactor more later on. Perhaps an idea could be to return the size from the function instead of bool, and where the code would return false it now returns -1 and we act accordingly?

CastagnaIT: ah ok i got it from the wrong angle...
no here we have unsigned value it is bad return negative value, then in this case i prefer rename it as
`retrieveCurrentSegmentBufferSize`

phunkyfish: Sorry, this comment was in no way supposed to hold up this PR. Just something for discussion 😉

CastagnaIT: no problem this will make it easier for me to keep it in mind

Issue #249, opened 2019-04-03
WARNING: ActiveAE - large audio sync error on Netflix Plugin


Vitocorl:
I opened this at CastagnaIT/plugin.video.netflix, but since then i found that peak3d has already dealt with similar errors in Kodi and inputstream, so i closed the issue at CastagnaIT/plugin.video.netflix and i am continuing here:
The issue is that audio gets out of sync and video starts stuttering on Xiaomi Mibox S, Android TV 8.1
Addon version used plugin.video.netflix_0.14.0beta23_20190331.
I also installed the latest Kodi nightly(kodi-20190402-f8100282-master-armeabi-v7a), as i realized that peak3d has provided a fix for a similar error happening after skipping forward a video file in Kodi, but it does not fix it for me.

The problem occurs at the same videos and at the same time of the video each time, for example, at Luce Cage TV Show, episode "Manifest" after 30 seconds of play. If i scroll backwards or forward the video then it resumes playing normally. I suspect it is some kind of buffering problem, but it only occurs on the Kodi Netflix plugin and not in the Android app. I am not sure if it is a Netflix plugin error or an inputstream adaptive error.
I cleaned cache of Netflix plugin, but afterwards the same happens.
I also used an optimised advancedsettings.xml, but i do not think Netflix plugin is using it.

My Kodi debug log is too big (2,2Mb) and hastebin does not accept it, so i am attching it as a zip here...
[kodi_debug_log.zip](https://github.com/peak3d/inputstream.adaptive/files/3037989/kodi_debug_log.zip)

Vitocorl:
@peak3d 
Is there a way to adjust the buffer memory size in inputstream? i think my advancesettings.xml has no effect whatsoever. Moreover, i still cannot tell if this problem is due to Kodi or inputstream, the issues have started since about ten days ago...

Vitocorl:
@peak3d
Any idea about this problem, it keeps happening, i have updated Kodi to the latest nightly with no effect...

Vitocorl:
@peak3d
I have just checked and when the problem happens i can see the buffer going down to 0%, as long as the buffer is over 0% (even if it is at 5%) the video plays smoothly, it is a buffering problem, but i do not know how to adjust it in inputstream....

peak3d:
From log there is a A/V diff of ~ 8ms, what is not nice but should not be visible much.
First I would try to switch screen to 59.94Hz and enable "Sync Playback To Display"

[...]

PR #878, opened 2022-01-18
Fix for huge media sequence numbers

glennguy:
In the case of a playlist that looks like this:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-ALLOW-CACHE:NO
#EXT-X-TARGETDURATION:6
#EXT-X-MEDIA-SEQUENCE:20220118074255
#EXTINF:6,
#EXT-X-PROGRAM-DATE-TIME:2022-01-18T11:27:55.000+0100
...
```
The `atol` function is returning the max value for 32 bit integer when fed large values like this which prevents the correct media sequence from being used to align segments between old and new playlists. Net effect is that segments get skipped, creating discontinuities and eventually the AdaptiveStream is always looking for a segment that is off into the future.

This PR fixes by using `atoll`, and all references to the value once stored are now using 64 bit uint variables. Tested on tennix.it stream as referenced in this https://github.com/xbmc/inputstream.adaptive/issues/249#issuecomment-1014926605 (live event now ended) and other HLS and DASH live streams tested for regression.

Will backport to Matrix

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -294,10 +294,10 @@ bool AdaptiveStream::start_stream()
       }
       else // switching streams, align new stream segment no.
       {
-        std::int32_t segmentId = segment_buffers_[0].segment_number;
+        std::int64_t segmentId = segment_buffers_[0].segment_number;
         if (segmentId >= current_rep_->startNumber_ + current_rep_->segments_.size())
           segmentId = current_rep_->startNumber_ + current_rep_->segments_.size() - 1;
-        current_rep_->current_segment_ = current_rep_->get_segment(segmentId - current_rep_->startNumber_);
+        current_rep_->current_segment_ = current_rep_->get_segment(static_cast<uint32_t>(segmentId - current_rep_->startNumber_));
       }
     }
     else
@@ -331,7 +331,7 @@ bool AdaptiveStream::start_stream()
 
     segment_buffers_[0].segment.Copy(loadingSeg);
     segment_buffers_[0].rep = current_rep_;
-    segment_buffers_[0].segment_number = ~0U;
+    segment_buffers_[0].segment_number = ~0ULL;
     segment_buffers_[0].buffer.clear();
     segment_read_pos_ = 0;
 
@@ -392,7 +392,7 @@ bool AdaptiveStream::prepareNextDownload()
   const AdaptiveTree::Representation* rep = segment_buffers_[valid_segment_buffers_].rep;
   const AdaptiveTree::Segment* seg = &segment_buffers_[valid_segment_buffers_].segment;
   // segNum == ~0U is initialization segment!
-  unsigned int segNum = segment_buffers_[valid_segment_buffers_].segment_number;
+  uint64_t segNum = segment_buffers_[valid_segment_buffers_].segment_number;
   segment_buffers_[valid_segment_buffers_].buffer.clear();
   ++valid_segment_buffers_;
 
@@ -401,7 +401,7 @@ bool AdaptiveStream::prepareNextDownload()
 
 bool AdaptiveStream::prepareDownload(const AdaptiveTree::Representation* rep,
                                      const AdaptiveTree::Segment* seg,
-                                     unsigned int segNum)
+                                     uint64_t segNum)
 {
   if (!seg)
     return false;
@@ -520,8 +520,8 @@ bool AdaptiveStream::ensureSegment()
     }
     if (valid_segment_buffers_)
       nextSegment = ~segment_buffers_[0].segment_number
-                        ? current_rep_->get_segment(segment_buffers_[0].segment_number -
-                                                    current_rep_->startNumber_)
+                        ? current_rep_->get_segment(static_cast<uint32_t>(
+                              segment_buffers_[0].segment_number - current_rep_->startNumber_))
                         : nullptr;
     else
       nextSegment = current_rep_->get_next_segment(current_rep_->current_segment_);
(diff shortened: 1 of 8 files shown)


src/common/AdaptiveStream.cpp
```diff
@@ -549,13 +549,13 @@ bool AdaptiveStream::ensureSegment()
...
       uint32_t nextsegmentPosold = current_rep_->get_segment_pos(nextSegment);
-      uint32_t nextsegno = current_rep_->getSegmentNumber(nextSegment);
+      uint64_t nextsegno = current_rep_->getSegmentNumber(nextSegment);
       AdaptiveTree::Representation* newRep;
       bool lastSeg =
           (current_period_ != tree_.periods_.back() &&
            nextsegmentPosold + available_segment_buffers_ == current_rep_->segments_.size() - 1);
       
-      if (segment_buffers_[0].segment_number == ~0L || valid_segment_buffers_ == 0 ||
+      if (segment_buffers_[0].segment_number == ~0LL || valid_segment_buffers_ == 0 ||
```
CastagnaIT: should be `if (segment_buffers_[0].segment_number == ~0ULL || valid_segment_buffers_ == 0 ||`

src/common/AdaptiveTree.h
```diff
@@ -251,14 +251,14 @@ class ATTR_DLL_LOCAL AdaptiveTree
       return get_segment_pos(current_segment_);
     };
 
-    uint32_t getCurrentSegmentNumber() const
+    uint64_t getCurrentSegmentNumber() const
     {
-      return current_segment_ ? get_segment_pos(current_segment_) + startNumber_ : ~0U;
+      return current_segment_ ? get_segment_pos(current_segment_) + startNumber_ : ~0ULL;
```
CastagnaIT: missing cast
`return current_segment_ ? static_cast<uint64_t>(get_segment_pos(current_segment_)) + startNumber_ : ~0ULL;`

src/common/AdaptiveTree.h
```diff
@@ -251,14 +251,14 @@ class ATTR_DLL_LOCAL AdaptiveTree
...
     };
 
-    uint32_t getSegmentNumber(const Segment *segment) const
+    uint64_t getSegmentNumber(const Segment *segment) const
     {
-      return segment ? get_segment_pos(segment) + startNumber_ : ~0U;
+      return segment ? get_segment_pos(segment) + startNumber_ : ~0ULL;
```
CastagnaIT: also here missing cast

src/parser/HLSTree.cpp
```diff
@@ -390,9 +390,9 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
   if (!rep->source_url_.empty())
   {
     SPINCACHE<Segment> newSegments;
-    unsigned int newStartNumber;
+    uint64_t newStartNumber;
     Segment newInitialization;
-    uint32_t segmentId(rep->getCurrentSegmentNumber());
+    uint64_t segmentId(rep->getCurrentSegmentNumber());
```
CastagnaIT, line 395: this change require more attention on line 751:
`rep->current_segment_ = rep->get_segment(segmentId - rep->startNumber_);`
`get_segment` method parameter accept uint32 not uint64
have you check for possible problems due to uint32?
PS. better re-check all calls to `get_segment`

CastagnaIT:
i think that also this line need to be changed to uint64:
https://github.com/xbmc/inputstream.adaptive/blob/fc91f5f2ae17ac084b7d8eaf0728d0c3fe06dd82/src/common/AdaptiveTree.h#L296

just for note `atol` and `atoll` not return unsigned values then need a cast

glennguy:
Thanks for all the feedback. Just realised as well that somehow I undid the change from `atol` to `atoll` in HLSTree for media sequence, which is the one thing that fixes the issue 🤡

src/parser/HLSTree.cpp
glennguy: I think get_segment should still accept uint32 as there's never going to be billions of segments, in the playlist example in the OP the expression would evaluate as `20220118074272 - 20220118074255` (== 17)  
How do feel about then casting the input for these cases i.e `rep->current_segment_ = rep->get_segment(static_cast<uint32_t>(segmentId - rep->startNumber_));` ? I don't think we'd ever suffer any loss of data but I know it's not good how it currently is.

CastagnaIT: mmh ok then add only a cast to uint32_t in the passed get_segment argument

IMO in another cleanup PR i think `SPINCACHE` type `pos` variable/method should be changed to size_t, since all this struct works with size_t and vector

glennguy:
@CastagnaIT ready for another review

src/common/AdaptiveTree.h
```diff
@@ -429,7 +429,8 @@ class ATTR_DLL_LOCAL AdaptiveTree
 
     std::vector<AdaptationSet*> adaptationSets_;
     std::string base_url_, id_;
-    uint32_t timescale_ = 1000, startNumber_ = 1, sequence_ = 0;
+    uint32_t timescale_ = 1000, sequence_ = 0;
```
CastagnaIT: since you changing existing code variables must be placed on each line, not in-line

src/parser/HLSTree.cpp
```diff
@@ -527,7 +527,7 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
         }
         else if (line.compare(0, 22, "#EXT-X-MEDIA-SEQUENCE:") == 0)
         {
-          newStartNumber = atol(line.c_str() + 22);
+          newStartNumber = atoll(line.c_str() + 22);
```
CastagnaIT:
```suggestion
          newStartNumber = static_cast<uint64_t>(std::atoll(line.c_str() + 22));
```
atoll is not unsigned

glennguy: I didn't add the cast as negative numbers in the HLS playlist would be illegal but you're right, better to show completeness here and remove any uncertainty from others reading.

Shame there is no atoull to make things nice and easy.

Edit:
How does this look now?

CastagnaIT: is not a problem of what value you parse but the type that it is different
unsigned type has a different range from signed, therefore this is not a good code
you assume that there will be no negative values
but if for any reason a negative value occurs, it will result in an unexpected positive value

CastagnaIT: new change is good one!

CastagnaIT (approved):

glennguy:
Thanks for your comments @CastagnaIT !

I'll do up a backport for Matrix tomorrow

Issue #880, opened 2022-01-20
Sling MPD Live Issue


eracknaphobia:
While trying to debug an [issue](https://github.com/d21spike/plugin.video.sling/issues/40) in the Sling TV Addon it was [suggested](https://github.com/d21spike/plugin.video.sling/issues/40#issuecomment-1017708480) to create an issue here.

1. First segment played `https://p-cdn4-e-cg14-linear-cbd46b77.movetv.com/15803/live/ESPNHD/4a8a1a5479cd11ecbc620025b5472112/video/vid06/0000493d.m4s`

2. MPD file
[spanning_ads-real-pto.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/7908021/spanning_ads-real-pto.mpd.txt)

3. Epoch  when fired `1642706751`

```
wall clock time = 2022-01-20T19:25:51Z (1642706751)
availabilityStartTime = 2022-01-20T19:00:00Z (1642705200)
duration = 2048
timescale = 1000
startNumber = 18015

((wall clock time - availabilityStartTime ) / (duration / timescale )) + startNumber

((1642706751 - 1642705200) / (2048 / 1000)) + 18015 
(1511 / 2.048) + 18015 
737 + 18015 = 18752
which is only 3 segments off of the first segment played in the browser (493d = 18749)
```

Hope this is helpful...

matthuisman:
@eracknaphobia IA is doing a few things wrong
1) https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/parser/DASHTree.cpp#L1285
 tpl.timescale / tpl.duration should be  tpl.duration/ tpl.timescale 
2) overallSeconds shouldnt be remove here. 
So the formula should be:
`seg.range_end_ += (static_cast<int64_t>(dash->stream_start_ - dash->available_time_ - sample_time)) * tpl.duration / tpl.timescale + 1;`
sample_time = the start time of current period = correct.
eg. if we are on 2nd period and it started 10s after the 1st period - we need to remove 10s

3) We are "double dipping" i think
  Here: https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/parser/DASHTree.cpp#L1242
  we are calculating the total number of segments based off oversalseconds (which will be based off the period duration = wrong in our case as its live and the period isnt finished) 
    This number of segements gets added here: https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/parser/DASHTree.cpp#L1300
[...]

Issue #731, opened 2021-07-09
 [Dash] Manifest without mediaPresentationDuration & timeShiftBufferDepth won't play


matthuisman:
Below won't play 
```
<?xml version="1.0" encoding="UTF-8"?>
<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:mpeg:DASH:schema:MPD:2011 DASH-MPD.xsd" profiles="urn:mpeg:dash:profile:isoff-live:2011" type="dynamic" availabilityStartTime="2021-07-08T22:35:15Z" publishTime="2021-07-08T22:35:00Z" minimumUpdatePeriod="PT24H" minBufferTime="PT2.048S" suggestedPresentationDelay="PT4.096S" maxSegmentDuration="PT2.048S">
  <ProgramInformation/>
  <Period id="1" start="PT0S">
    <BaseURL serviceLocation="15805-Akamai">https://p-cdn1-805-cg14-linear-cbd46b77.movetv.com/15805/live/COMEDY/0b804a18dfd111ebbc620025b5472112/</BaseURL>
    <AdaptationSet contentType="video" mimeType="video/mp4" segmentAlignment="true" startWithSAP="1">
      <ContentProtection schemeIdUri="urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b"/>
      <Accessibility schemeIdUri="urn:scte:dash:cc:cea-608:2015" value="CC1=eng"/>
      <Accessibility schemeIdUri="urn:scte:dash:cc:cea-708:2015" value="1=lang:eng"/>
      <Role schemeIdUri="urn:mpeg:dash:role:2011" value="main"/>
      <SegmentTemplate initialization="video/$RepresentationID$/init.mp4" media="video/$RepresentationID$/$Number%08x$.m4s" startNumber="22588" timescale="1000" duration="2048" presentationTimeOffset="46255543"/>
      <Representation id="vid06" bandwidth="5632000" width="1920" height="1080" codecs="avc1.640028" sar="1:1" frameRate="29970/1000" qualityRanking="0"/>
      <Representation id="vid05" bandwidth="4300800" width="1280" height="720" codecs="avc1.640028" sar="1:1" frameRate="29970/1000" qualityRanking="1"/>
[...]

PR #881, opened 2022-01-21
WIP

matthuisman:
fixes: https://github.com/xbmc/inputstream.adaptive/issues/880
fixes: https://github.com/xbmc/inputstream.adaptive/issues/731


src/test/TestDASHTree.cpp
```diff
@@ -187,30 +187,30 @@ TEST_F(DASHTreeTest, CalculateCorrectSegmentNumbersFromSegmentTimeline)
...
 
   OpenTestFile("mpd/segtpl_pto.mpd", "", "");
 
   adaptive::SPINCACHE<adaptive::AdaptiveTree::Segment> segments =
       tree->periods_[0]->adaptationSets_[0]->representations_[0]->segments_;
 
-  EXPECT_EQ(segments.size(), 451);
-  EXPECT_EQ(segments[0]->range_end_, 404305525);
-  EXPECT_EQ(segments[450]->range_end_, 404305975);
+  EXPECT_EQ(segments.size(), 25); //25 * 4 = 100 seconds
```
glennguy, line 197: I'm not understanding the 100 second result here for both manifests - TSBD for this is 1800 and 120 for the other, the old values for segment size seem correct.

matthuisman:
100s was just a random value "in the future" for then checking if we get
the right number of segments that add up to it.

PR #883, opened 2022-01-22
Begin refactor of wvdecrypter, sample-aes-cbc support

glennguy:
This is the result of looking into adding support for CBC encrypted streams. Not very common at the moment but I imagine it would be more popular with 'apple native' services - Fairplay uses this scheme. CBC uses a 'pattern' based encryption, where there is a pattern of clear vs encrypted samples. Typically (and in the case of the test video) there is 1 encrypted sample followed by 9 unencrypted or 'skipped' samples. An advantage of this method can be lower cpu utilization.

main.cpp doesn't want to know the concrete decrypter class, only the generic ap4 cencsinglesampledecrypter. This will provide the flexibility needed in future if there's a Playready or Fairplay decrypter module developed. Therefore for this AES-CBC feature in order to prevent yet more patches to bento4 or to cast into the concrete wv cenc ssd class we have now the 'Adaptive_CencSingleSampleDecrypter' base class which has this extended functionality.

If CBC encryption is found upon stream initialisation we set the decrypter up appropriately.

NOTE: This doesn't include support for the CBCS decoding feature on Android, changes and possibly API bump will be required in Kodi. Follow-up PR to come very soon.

Tested with
```
#KODIPROP:inputstream.adaptive.license_key=https://proxy.uat.widevine.com/proxy?provider=widevine_test||R{SSM}|
https://storage.googleapis.com/wvmedia/cbcs/h264/tears/tears_aes_cbcs.mpd
```
for CBC, and for regular CTR tested with:
```
#KODIPROP:inputstream.adaptive.license_key=https://cwip-shaka-proxy.appspot.com/no_auth|Content-Type=|R{SSM}|
https://bitmovin-a.akamaihd.net/content/art-of-motion_drm/mpds/11331.mpd
```

on Windows 10.

edit: some additional background

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ set(ADP_HEADERS
 	src/codechandler/VP9CodecHandler.h
 	src/codechandler/WebVTTCodecHandler.h
 	src/codechandler/ttml/TTML.h
+    src/common/AdaptiveDecrypter.h
 	src/common/AdaptiveStream.h
 	src/common/AdaptiveTree.h
 	src/common/RepresentationChooser.h
--- a/src/common/AdaptiveDecrypter.h
+++ b/src/common/AdaptiveDecrypter.h
@@ -1,0 +1,33 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include <bento4/Ap4.h>
+
+enum class ENCRYPTION_SCHEME
+{
+  NONE,
+  CENC,
+  CBCS
+};
+
+class Adaptive_CencSingleSampleDecrypter : public AP4_CencSingleSampleDecrypter
+{
+public:
+  Adaptive_CencSingleSampleDecrypter() : AP4_CencSingleSampleDecrypter(0){};
+
+  void SetCrypto(AP4_UI08 cryptBlocks, AP4_UI08 skipBlocks)
+  {
+    m_CryptBlocks = static_cast<uint32_t>(cryptBlocks);
+    m_SkipBlocks = static_cast<uint32_t>(skipBlocks);
+  };
+  virtual void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme){};
+
+protected:
+  uint32_t m_CryptBlocks{0};
+  uint32_t m_SkipBlocks{0};
+};
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -577,7 +577,7 @@ public:
                          AP4_Movie* movie,
                          AP4_Track* track,
                          AP4_UI32 streamId,
-                         AP4_CencSingleSampleDecrypter* ssd,
+                         Adaptive_CencSingleSampleDecrypter* ssd,
                          const SSD::SSD_DECRYPTER::SSD_CAPS& dcaps)
     : AP4_LinearReader(*movie, input),
       m_track(track),
@@ -965,6 +965,19 @@ protected:
                            AP4_CencSampleDecrypter::Create(sample_table, algorithm_id, 0, 0, 0, reset_iv,
                                                            m_singleSampleDecryptor, m_decrypter)))
           return result;
+
+        // Inform decrypter of pattern decryption (CBCS)
+        if (m_protectedDesc->GetSchemeType() == AP4_PROTECTION_SCHEME_TYPE_CENC)
+        {
+          m_singleSampleDecryptor->SetEncryptionScheme(ENCRYPTION_SCHEME::CENC);
+          m_singleSampleDecryptor->SetCrypto(0, 0);
+        }
+        else if (m_protectedDesc->GetSchemeType() == AP4_PROTECTION_SCHEME_TYPE_CBCS)
+        {
+          m_singleSampleDecryptor->SetEncryptionScheme(ENCRYPTION_SCHEME::CBCS);
+          m_singleSampleDecryptor->SetCrypto(sample_table->GetCryptByteBlock(),
+                                             sample_table->GetSkipByteBlock());
+        }
       }
     }
   SUCCESS:
@@ -1048,7 +1061,7 @@ private:
   const AP4_UI08* m_defaultKey;
 
   AP4_ProtectedSampleDescription* m_protectedDesc;
-  AP4_CencSingleSampleDecrypter* m_singleSampleDecryptor;
+  Adaptive_CencSingleSampleDecrypter* m_singleSampleDecryptor;
   AP4_CencSampleDecrypter* m_decrypter;
   uint64_t m_nextDuration, m_nextTimestamp;
 };
@@ -2002,21 +2015,43 @@ bool Session::InitializeDRM()
 
     unsigned char key_system[16];
     AP4_ParseHex(strkey.c_str(), key_system, 16);
+    uint32_t currentSessionTypes = 0;
 
     for (size_t ses(1); ses < cdm_sessions_.size(); ++ses)
     {
       AP4_DataBuffer init_data;
       const char* optionalKeyParameter(nullptr);
+      adaptive::AdaptiveTree::Period::PSSH sessionPsshset =
+        adaptiveTree_->current_period_->psshSets_[ses];
+      uint32_t sessionType = 0;
 
-      if (adaptiveTree_->current_period_->psshSets_[ses].pssh_ == "FILE")
+      if (sessionPsshset.media_ > 0)
+      {
+        sessionType = sessionPsshset.media_;
+      }
+      else
+      {
+        switch (sessionPsshset.adaptation_set_->type_)
+        {
+          case adaptive::AdaptiveTree::VIDEO:
+            sessionType = adaptive::AdaptiveTree::Period::PSSH::MEDIA_VIDEO;
+            break;
+          case adaptive::AdaptiveTree::AUDIO:
+            sessionType = adaptive::AdaptiveTree::Period::PSSH::MEDIA_AUDIO;
+            break;
+          default:
+            break;
+        }
+      }
+
+      if (sessionPsshset.pssh_ == "FILE")
       {
         LOG::Log(LOGDEBUG, "Searching PSSH data in FILE");
 
         if (m_kodiProps.m_licenseData.empty())
         {
-          Session::STREAM stream(
-              *adaptiveTree_, adaptiveTree_->current_period_->psshSets_[ses].adaptation_set_,
-              media_headers_, representationChooser_, m_kodiProps.m_playTimeshiftBuffer, 0, false);
+          Session::STREAM stream(*adaptiveTree_, sessionPsshset.adaptation_set_, media_headers_,
+                                 representationChooser_, m_kodiProps.m_playTimeshiftBuffer, 0, false);
 
           stream.enabled = true;
           stream.m_kodiAdStream.start_stream();
@@ -2037,10 +2072,10 @@ bool Session::InitializeDRM()
             if (memcmp(pssh[i].GetSystemId(), key_system, 16) == 0)
             {
               init_data.AppendData(pssh[i].GetData().GetData(), pssh[i].GetData().GetDataSize());
-              if (adaptiveTree_->current_period_->psshSets_[ses].defaultKID_.empty())
+              if (sessionPsshset.defaultKID_.empty())
               {
                 if (pssh[i].GetKid(0))
-                  adaptiveTree_->current_period_->psshSets_[ses].defaultKID_ =
+                  sessionPsshset.defaultKID_ =
                       std::string((const char*)pssh[i].GetKid(0), 16);
                 else if (AP4_Track* track = movie->GetTrack(TIDC[stream.m_kodiAdStream.get_type()]))
                 {
@@ -2053,16 +2088,20 @@ bool Session::InitializeDRM()
                     AP4_TencAtom* tenc(
                         AP4_DYNAMIC_CAST(AP4_TencAtom, schi->GetChild(AP4_ATOM_TYPE_TENC, 0)));
                     if (tenc)
-                      adaptiveTree_->current_period_->psshSets_[ses].defaultKID_ =
-                          std::string((const char*)tenc->GetDefaultKid(), 16);
+                    {
+                      sessionPsshset.defaultKID_ =
+                          std::string(reinterpret_cast<const char*>(tenc->GetDefaultKid()), 16);
+                    }
                     else
                     {
                       AP4_PiffTrackEncryptionAtom* piff(
                           AP4_DYNAMIC_CAST(AP4_PiffTrackEncryptionAtom,
                                            schi->GetChild(AP4_UUID_PIFF_TRACK_ENCRYPTION_ATOM, 0)));
                       if (piff)
-                        adaptiveTree_->current_period_->psshSets_[ses].defaultKID_ =
-                            std::string((const char*)piff->GetDefaultKid(), 16);
+                      {
+                        sessionPsshset.defaultKID_ =
+                            std::string(reinterpret_cast<const char*>(piff->GetDefaultKid()), 16);
+                      }
                     }
                   }
                 }
@@ -2079,10 +2118,9 @@ bool Session::InitializeDRM()
           }
           stream.disable();
         }
-        else if (!adaptiveTree_->current_period_->psshSets_[ses].defaultKID_.empty())
+        else if (!sessionPsshset.defaultKID_.empty())
         {
-          init_data.SetData(
-              (AP4_Byte*)adaptiveTree_->current_period_->psshSets_[ses].defaultKID_.data(), 16);
+          init_data.SetData(reinterpret_cast<AP4_Byte*>(sessionPsshset.defaultKID_.data()), 16);
 
           std::string decLicenseData{BASE64::Decode(m_kodiProps.m_licenseData)};
           uint8_t* decLicDataUInt = reinterpret_cast<uint8_t*>(decLicenseData.data());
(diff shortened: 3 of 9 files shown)


src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,20 @@
+#include <bento4/Ap4.h>
+
+enum class ENCRYPTION_SCHEME
+{
+  ENCRYPTION_SCHEME_NONE,
```
CastagnaIT: on `enum class` the class name prefix on each entries is not needed

src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,20 @@
+#include <bento4/Ap4.h>
+
+enum class ENCRYPTION_SCHEME
+{
+  ENCRYPTION_SCHEME_NONE,
+  ENCRYPTION_SCHEME_CENC,
+  ENCRYPTION_SCHEME_CBCS
+};
+
+class Adaptive_CencSingleSampleDecrypter : public AP4_CencSingleSampleDecrypter
```
CastagnaIT, line 18: since now we handle differents types of encryption
I wonder if it would be better to make the class name more generic, something like
`Adaptive_SingleSampleDecrypter` ?

CastagnaIT (commented):
nice addition! only some code style requests

src/common/AdaptiveDecrypter.h
glennguy: Happy to change, was just emulating style from here: https://github.com/xbmc/inputstream.adaptive/blob/d05dba529df20a5a243b2ff0ae98b61cfff61e29/src/common/AdaptiveTree.h#L107-L131

but I can see now that it's very redundant

src/common/AdaptiveDecrypter.h
glennguy: The issue here is that there's CENC and cenc.

CENC is the common encryption standard, and within the different encryption/decryption modes are supported. CTR mode is (annoyingly) referred to as 'cenc' and CBC as 'cbcs' - defined in content_decryption_module.h.

So I believe that AP4_CencSingleSampleDecrypter has been named after CENC and not cenc.

Therefore... I think we stick with keeping the same naming convention for the derived classes. Hope that makes sense!

CastagnaIT: I had not realised this distinction, what a mess...OFC not your fault

src/common/AdaptiveDecrypter.h
CastagnaIT: note that these are regular enum's are similar but different
in this case you havent the class name then without prefix you could create confusion/conflicts between entries with others enum's

glennguy: Oh I see now, thanks for the explanation

src/main.cpp
```diff
@@ -1492,6 +1492,20 @@ class ATTR_DLL_LOCAL FragmentedSampleReader : public SampleReader, public AP4_Li
                            AP4_CencSampleDecrypter::Create(sample_table, algorithm_id, 0, 0, 0, reset_iv,
                                                            m_singleSampleDecryptor, m_decrypter)))
           return result;
+
+        // Inform decrypter of pattern decryption (CBCS)
+        AP4_UI08 cryptBlocks = sample_table->GetCryptByteBlock();
+        AP4_UI08 skipBlocks = sample_table->GetSkipByteBlock();
+
+        if (cryptBlocks > 0 && skipBlocks > 0)
```
CastagnaIT: Just a question, i see that here you give priority to these variables to set CBCS,
is because `m_protectedDesc->GetSchemeType()` not provide the right scheme?

src/main.cpp
```diff
@@ -2700,14 +2714,14 @@ bool Session::InitializeDRM()
         AP4_FormatHex(reinterpret_cast<const AP4_UI08*>(defkid), 16, hexkid), hexkid[32] = 0;
         kodi::Log(ADDON_LOG_DEBUG, "Initializing stream with KID: %s", hexkid);
 
-        for (unsigned int i(1); i < ses; ++i)
+        /*for (unsigned int i(1); i < ses; ++i)
```
CastagnaIT: its oversight or why has been commented out?

src/main.h
```diff
@@ -164,9 +165,9 @@ class ATTR_DLL_LOCAL Session : public adaptive::AdaptiveStreamObserver
   unsigned int GetStreamCount() const { return streams_.size(); };
   const char *GetCDMSession(int nSet) { return cdm_sessions_[nSet].cdm_session_str_; };;
   uint8_t GetMediaTypeMask() const { return media_type_mask_; };
-  AP4_CencSingleSampleDecrypter * GetSingleSampleDecryptor(unsigned int nIndex)const{ return cdm_sessions_[nIndex].single_sample_decryptor_; };
+  Adaptive_CencSingleSampleDecrypter * GetSingleSampleDecryptor(unsigned int nIndex)const{ return cdm_sessions_[nIndex].single_sample_decryptor_; };
```
CastagnaIT: please apply clang formatting, here and all others modified below

CastagnaIT (commented):
i havent tested the feature, just check code style

src/main.cpp
glennguy: I totally forgot about this... I need to do my own review before submitting!

This needs some thought actually, I will try to find a good solution. But in any case, CBC can't be done with a single CDM session as there are separate patterns required between audio and video.

src/main.cpp
glennguy: Audio will come back with 0,0 for these values (which is 100% of samples encrypted) and video will usually be 1,9. 

I could rearrange it to a 
```
if (CBCS) 
{
  if (blocks >0)
  {
   //send pattern and cbcs
  }
else
  {
  // just send cbcs
  }
}
else
{
// send cenc
}
```
if you wanted with the same result?

src/main.cpp
glennguy: The main problem being if we have an mpd with say 10 video representations, each with their own key then license server get contacted 10 times == slow to start stream. This would already be affecting Android with cenc streams btw

src/main.cpp
CastagnaIT: if it is the same yes seem more correct

glennguy: Ok I think it's much better now

src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,24 @@
+#include <bento4/Ap4.h>
```
CastagnaIT, line 9: here is missing the copyright, you can copy from the main.cpp and change the years with 2022 only,
when i have time i think i can convert them all with SPDX standard

src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,24 @@
...
+{
+  NONE,
+  CENC,
+  CBCS
+};
+
+class Adaptive_CencSingleSampleDecrypter : public AP4_CencSingleSampleDecrypter
+{
+public:
+
```
CastagnaIT: uneeded space

src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,24 @@
...
+  NONE,
+  CENC,
+  CBCS
+};
+
+class Adaptive_CencSingleSampleDecrypter : public AP4_CencSingleSampleDecrypter
+{
+public:
+  
+  Adaptive_CencSingleSampleDecrypter() : AP4_CencSingleSampleDecrypter(0) {};
```
CastagnaIT:
```suggestion
  Adaptive_CencSingleSampleDecrypter() : AP4_CencSingleSampleDecrypter(0){};
```
clang formatting

src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,24 @@
...
+
+  void SetCrypto(AP4_UI08 cryptBlocks, AP4_UI08 skipBlocks)
+  {
+    m_CryptBlocks = cryptBlocks;
+    m_SkipBlocks = skipBlocks;
+  };
+  virtual void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme){};
+  AP4_UI08 m_CryptBlocks = 0;
```
CastagnaIT: please always separate variables and methods in headers with a empty line

CastagnaIT: i think that these variable should be `protected` and not `public`

CastagnaIT: on new variables please use this standard:
`AP4_UI08 m_CryptBlocks{0};`

src/common/AdaptiveDecrypter.h
```diff
@@ -0,0 +1,24 @@
...
+public:
+  
+  Adaptive_CencSingleSampleDecrypter() : AP4_CencSingleSampleDecrypter(0) {};
+
+  void SetCrypto(AP4_UI08 cryptBlocks, AP4_UI08 skipBlocks)
+  {
+    m_CryptBlocks = cryptBlocks;
+    m_SkipBlocks = skipBlocks;
+  };
+  virtual void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme){};
```
CastagnaIT, line 28:
```suggestion
  virtual void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme) = 0;
```
this is pure virtual

src/main.cpp
```diff
@@ -2558,21 +2571,43 @@ bool Session::InitializeDRM()
...
     AP4_ParseHex(strkey.c_str(), key_system, 16);
+    uint32_t currentSessionTypes = 0;
 
     for (size_t ses(1); ses < cdm_sessions_.size(); ++ses)
     {
       AP4_DataBuffer init_data;
       const char* optionalKeyParameter(nullptr);
+      adaptive::AdaptiveTree::Period::PSSH sessionPsshset =
+        adaptiveTree_->current_period_->psshSets_[ses];
+      uint32_t sessionType = 0;
```
CastagnaIT, line 2026: IMO this is not very clear, here we should make it easier understand that we handle flag operations,
at least we should change all these integer variables to handle a enum, something like this:
```c++
enum MediaTypeFlags
{
MEDIA_TYPE_UNDEFINED = 0,
MEDIA_TYPE_VIDEO,
MEDIA_TYPE_AUDIO
...
```
if it is too much work for this PR, keep it for later cleanup

wvdecrypter/wvdecrypter.cpp
```diff
@@ -214,7 +215,9 @@ class WV_CencSingleSampleDecrypter : public AP4_CencSingleSampleDecrypter
 
   bool OpenVideoDecoder(const SSD_VIDEOINITDATA *initData);
   SSD_DECODE_RETVAL DecodeVideo(void* hostInstance, SSD_SAMPLE *sample, SSD_PICTURE *picture);
-  void ResetVideo();
+  void ResetVideo();
+
+  virtual void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme) override;
```
CastagnaIT:
```suggestion
  void SetEncryptionScheme(ENCRYPTION_SCHEME encryptionScheme) override;
```
no need virtual

CastagnaIT (commented):
review second round done,
i have tested it on Windows 10 and no problem,
but when i have tested it on CoreElec (last Kodi 20 nightly) every time i try to play multiple times the first test stream Kodi always crashes (usually after 2 or 3 times), this not happens with the second test stream or others test streams

Crash logs:
[kodi_crashlog_20220128145622.log.txt](https://github.com/xbmc/inputstream.adaptive/files/7960550/kodi_crashlog_20220128145622.log.txt)
[kodi_crashlog_20220128153636.log.txt](https://github.com/xbmc/inputstream.adaptive/files/7960551/kodi_crashlog_20220128153636.log.txt)

src/common/AdaptiveDecrypter.h
glennguy: The intention here was to have a default implementation in the case of a future module that doesn't need this info, therefore not having to override just to put `{}` in the derived class.

glennguy: In the interest of expedience though I'll change it :)
Edit: I forgot about the case of wvdecrypter_android. Now it makes more sense to leave as-is - the Android implementation needs to do nothing here as the info will come through the demux packets. IMHO it's cleaner to not have to implement functions that are no-op. What do you think?

src/main.cpp
glennguy: Thanks for pointing out however let's keep for a later cleanup

src/common/AdaptiveDecrypter.h
CastagnaIT: in this case i agree keep it as is

glennguy:
Hmmm very interested in the crash on CoreELEC. Must be some sort of race condition, it seems very similar to other reports we've had from crashes on those platforms. But very good that it seems fairly easy to reproduce.

First crash shows that ISA isn't even running still which makes me think it's to do with the wv timer renewal introduced in https://github.com/xbmc/inputstream.adaptive/pull/589 and modified again in https://github.com/xbmc/inputstream.adaptive/pull/729

One hint for us to look at is that there's actually a callback that is invoked by the CDM that we're not currently using which may be the key to doing this right.

I know that the changes here are seeming to cause crashes on your setup but I feel that the changes merely expose an issue that's already there. One thing I didn't mention in the OP is this introduces using separate CDM sessions for audio and video - you'll notice in the log that the license is retrieved twice. Same behaviour as Android is since https://github.com/xbmc/inputstream.adaptive/pull/734

Or, alternatively something inside ISA is causing a failure or exit but it's not waiting properly for CDM session to finish closing, therefore crash of all Kodi when it tries to call back.... will have to see.

I'll have a dig very soon.

glennguy:
@CastagnaIT just a thought: would switching to separate audio and video widevine sessions cause issues with the pre-init for NF? Is the platform that you model the plugin from use separate audio/video sessions for playback or just 1?

CastagnaIT:
i have some questions

if the audio track is not DRM protected, is the licence still required/requested?

> you'll notice in the log that the license is retrieved twice. Same behaviour as Android is since 

because i have tried this PR on ntfx but i am not able to see in the log the two license requests,
can you explain where is in log?
https://paste.kodi.tv/azoliyifuv.kodi

If i use the licese provided with the licensedManifest request, this give me only one license data, not two licenses,
ATM I can request a license separately then is not a big problem,
but i don't know if request two licenses to the server is the right way
at least in the netfx website service this has never happened over the years

For the sake of completeness, after the implementation of thelicensedManifest  in website, there is an additional licese request, but this happens only after the video playback is started, which might be the reason why playback fails in my WIP PR after some minuts

glennguy:
> if the audio track is not DRM protected, is the licence still required/requested?

Sorry I should have clarified, my question should have been this condition first. No licence request for audio if not protected. In this case we are good and it makes sense.

I haven't had much time this weekend to look further but I'll have a go in the evenings this coming week. My gut feeling is this has to do with the CDM timer functions. I logged the timer getting started and finishing, it seems to run every second. Maybe some sort of race condition....

edit:
Just spotted this https://github.com/xbmc/inputstream.adaptive/blob/dc500f59873d515fd66d318ec86428d2e7f9da1e/wvdecrypter/cdm/media/cdm/cdm_adapter.cc#L524
I don't think this is right. If the cdm wants to set a timer it should be set without condition... it's possible that SetTimer is being called by CDM before wvdecrypter has a chance to set the flag.

glennguy:
@CastagnaIT I've rebased and cleaned up, and retested on LE11 - I couldn't get it to crash on wv 4.10.2252.5
As discussed previously, I fully believe that 4.10.2391.0 is broken and the cause of crashing previously. I tried to run on Raspberry Pi OS and test in a browser - performance rendered it unwatchable.

If you get a chance could you try again with the 'good' CDM?

glennguy:
Also - I reverted a change made to this line https://github.com/xbmc/inputstream.adaptive/blob/781d59ef2431d047753891a73aa0b055eb840f24/wvdecrypter/wvdecrypter.cpp#L483 as the variable was global to the cdm adapter and not the session, when setting up the second session the `session_active` flag would already be true.

CastagnaIT:
i will have to check i do not remember anymore the situation on my ARM CoreELEC device
i will let you know

CastagnaIT:
I have done tests

WV 4.10.2391.0 (last):
as you said kodi crash still present, performance audio/video always good,
and after about 1:30 the video become black, kodi try buffering after resume only audio for 2 secs, after that still black screen and no audio until you stop manually the video.
https://paste.kodi.tv/uguwogajon.kodi

WV 4.10.2252.0 (13904):
i confirm no kodi crashes, performance audio/video always good,
but and after about 1:30 the video become black, kodi try buffering after resume only audio for 2 secs, after that still black screen and no audio until you stop manually the video.
https://paste.kodi.tv/norilasima.kodi

WV 4.10.2252.0 (13816) and WV 4.10.2252.5 (14150):
same behaviour of 4.10.2252.0 (13904) but kodi crashed

after these tests on multiple versions i don't know if we can call it as library bug, crash always occurs at playback startup,
i would not want it to be a side effect of the problem we already have, namely that the WV library is not unloaded from memory, maybe it is possible that some process is like zombie

maybe wait for a new WV version to be released and see if crash still occurs
or merge it but we have to add in Wiki a big warning for unstable feature

glennguy:
Thanks for the detailed testing!

I haven't been able to find evidence that it's not unloaded from memory. What I can say is there's the long standing issue in CE/LE that seems to do with their filesystem/caching that usually requires a restart after updating the add-on, maybe that has to do with not 'unloading'?

These different versions (13816, 13904, 14150) are coming from different device images right? Are you able do checksums on the two different 4.10.2252.0 versions? Maybe one of these is being optimised in a certain way that causes issues for us, especially when combined with the TLS relocation 'fixes' that are now required in LE/CE?

Regarding the black screen, can you check that the changes from https://github.com/xbmc/inputstream.adaptive/pull/886/commits/d6b3c5291f3c6ad570b4de398599270d4c981897 are in your source? It looks to me like the incorrect init segment is downloaded on stream change (480p instead of 1080p).

CastagnaIT:
i dont know i have done exactly what you need
i have used `md5sum` linux command:
WV 4.10.2252.0 (13816): 6f6a10bb4a0cfc492815cab996addc4f
WV 4.10.2252.0 (13904): 6f6a10bb4a0cfc492815cab996addc4f
WV 4.10.2252.5 (14150): 9dfe48c60265444f08548e4769e00d8f

> Regarding the black screen, can you check that the changes from https://github.com/xbmc/inputstream.adaptive/commit/d6b3c5291f3c6ad570b4de398599270d4c981897 are in your source? It looks to me like the incorrect init segment is downloaded on stream change (480p instead of 1080p).

i have builded from your branch, your branch include that commit

> These different versions (13816, 13904, 14150) are coming from different device images right?

yes ISHelper download from different devices images
i think could be also possible that some devices are optimized in different way

glennguy:
Thanks for md5sums, that was what I was looking for. Both 2252.0 are the same file, interesting that there are crashes on one and not the other. It's making me think about the apparent file caching issue.

Ok, so I think the black screen issue is to do with this `2022-03-14 16:40:51.040 T:4113    ERROR <general>: CDVDVideoCodecAmlogic::Open - InstanceGuard locked` in the log, not sure what it means and haven't researched yet. No problems on RPi3 with stream changes though.

> i have builded from your branch, your branch include that commit

my bad, I read the log wrong, all is working as it should in this area.

Overall I'd like to just get this in. The feature is unlikely to be used at the moment, and at least it's prepared for the future. If there are issues later on then we can revisit then. When I move my other devices in the house to Kodi 20 I can do some more work.

I found with testing that the SetTimer function was the point at which crashing occurred, usually on the 10th (sometimes 7th) call to `TimerExpired`. Easiest way to stop timer is to comment out here https://github.com/xbmc/inputstream.adaptive/blob/781d59ef2431d047753891a73aa0b055eb840f24/wvdecrypter/cdm/media/cdm/cdm_adapter.cc#L507
Even 4.10.2391.0 seemed to behave nice (although with stuttering/slowness) with this.
This observation makes me think we're not doing anything wrong on our side.

CastagnaIT:
`InstanceGuard locked` error not come from code of xbmc original repository,
seem to be a CoreElec implementation and now i realized that this happens after DEMUX_SPECIALID_STREAMCHANGE
i will try update the latest Coreelec nightly and see if it still happens
in case i will contact CoreElec devs

EDIT: same error on last CE nightly, i ask support to CE dev

CastagnaIT:
i asked to CE dev and seem there is a problem with the decoder that is forced limited to one instance in CE or something related to this, they will let us know if a possible solution could be done, they suspect/theorize a open/close stream overlapping in kodi not allowed in CE,
for now the only way to playback with CE is disable stream adaptive

CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ set(ADP_HEADERS
 	src/codechandler/VP9CodecHandler.h
 	src/codechandler/WebVTTCodecHandler.h
 	src/codechandler/ttml/TTML.h
+  src/common/AdaptiveDecrypter.h
```
CastagnaIT: just can you use same spaces of others

CastagnaIT (approved):

PR #887, opened 2022-01-28
Send mode and pattern with demux packets

glennguy:
Dependent on https://github.com/xbmc/xbmc/pull/20915

Additional information required by inputstream interface is sent with these changes:
* Encryption mode
* Sample encryption pattern

This allows for decryption of CBCS streams on platforms receiving encrypted packets (Android)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2193,7 +2193,11 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
       memcpy(p->cryptoInfo->kid, pData, 16);
       pData += 16;
       iSize -= (pData - sr->GetSampleData());
+      ReaderCryptoInfo cryptoInfo = sr->GetReaderCryptoInfo();
       p->cryptoInfo->flags = 0;
+      p->cryptoInfo->cryptBlocks = cryptoInfo.m_cryptBlocks;
+      p->cryptoInfo->skipBlocks = cryptoInfo.m_skipBlocks;
+      p->cryptoInfo->mode = static_cast<uint16_t>(cryptoInfo.m_mode);
     }
     else
       p = AllocateDemuxPacket(iSize);
--- a/src/samplereader/FragmentedSampleReader.cpp
+++ b/src/samplereader/FragmentedSampleReader.cpp
@@ -408,16 +408,20 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
       }
 
       // Inform decrypter of pattern decryption (CBCS)
-      if (m_protectedDesc->GetSchemeType() == AP4_PROTECTION_SCHEME_TYPE_CENC)
+      AP4_UI32 schemeType = m_protectedDesc->GetSchemeType();
+      if (schemeType == AP4_PROTECTION_SCHEME_TYPE_CENC || schemeType == AP4_PROTECTION_SCHEME_TYPE_CBCS)
       {
-        m_singleSampleDecryptor->SetEncryptionScheme(ENCRYPTION_SCHEME::CENC);
-        m_singleSampleDecryptor->SetCrypto(0, 0);
-      }
-      else if (m_protectedDesc->GetSchemeType() == AP4_PROTECTION_SCHEME_TYPE_CBCS)
-      {
-        m_singleSampleDecryptor->SetEncryptionScheme(ENCRYPTION_SCHEME::CBCS);
-        m_singleSampleDecryptor->SetCrypto(sample_table->GetCryptByteBlock(),
-                                           sample_table->GetSkipByteBlock());
+        m_readerCryptoInfo.m_cryptBlocks = sample_table->GetCryptByteBlock();
+        m_readerCryptoInfo.m_skipBlocks = sample_table->GetSkipByteBlock();
+
+        if (schemeType == AP4_PROTECTION_SCHEME_TYPE_CENC)
+          m_readerCryptoInfo.m_mode = CryptoMode::AES_CTR;
+        else
+          m_readerCryptoInfo.m_mode = CryptoMode::AES_CBC;
+        
+        m_singleSampleDecryptor->SetEncryptionScheme(m_readerCryptoInfo.m_mode);
+        m_singleSampleDecryptor->SetCrypto(m_readerCryptoInfo.m_cryptBlocks,
+                                           m_readerCryptoInfo.m_skipBlocks);
       }
     }
   }
--- a/src/samplereader/FragmentedSampleReader.h
+++ b/src/samplereader/FragmentedSampleReader.h
@@ -86,4 +86,5 @@ private:
   AP4_CencSampleDecrypter* m_decrypter{nullptr};
   uint64_t m_nextDuration{0};
   uint64_t m_nextTimestamp{0};
+  ReaderCryptoInfo m_readerCryptoInfo{ReaderCryptoInfo()};
 };
--- a/src/samplereader/SampleReader.h
+++ b/src/samplereader/SampleReader.h
@@ -14,6 +14,22 @@
 #include <kodi/AddonBase.h>
 #include <kodi/addon-instance/Inputstream.h>
 
+// These values must match their respective constant values
+// defined in the Android MediaCodec class 
+enum class CryptoMode : uint16_t
+{
+  NONE = 0,
+  AES_CTR = 1,
+  AES_CBC = 2
+};
+
+struct ReaderCryptoInfo
+{
+  uint8_t m_cryptBlocks{0};
+  uint8_t m_skipBlocks{0};
+  CryptoMode m_mode{CryptoMode::NONE};
+};
+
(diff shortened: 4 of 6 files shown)


src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -410,14 +410,19 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
       // Inform decrypter of pattern decryption (CBCS)
       if (m_protectedDesc->GetSchemeType() == AP4_PROTECTION_SCHEME_TYPE_CENC)
       {
+        m_cryptBlocks = 0;
```
glennguy: I think this whole section could be simplified but I thought to be a bit conservative - I'm not sure if GetCrypt/SkipByteBlock will always return 0 for regular old CTR, and if it's safe to just use the `m_cryptMode` values to send to `SetCrypto` since the values all line up with each other - would this be bad practice?

CastagnaIT: for me its ok as you have done by explicit set all vars

the only thing that i am wondering is,
its better take the example of the IS interface that group all crypto info vars in to a single struct var, like done in `cryptoInfo` with `DEMUX_CRYPTO_INFO` struct? instead to have to add new getters for each new crypto info var?

src/samplereader/FragmentedSampleReader.h
```diff
@@ -86,4 +86,7 @@ class ATTR_DLL_LOCAL CFragmentedSampleReader : public ISampleReader, public AP4_
   AP4_CencSampleDecrypter* m_decrypter{nullptr};
   uint64_t m_nextDuration{0};
   uint64_t m_nextTimestamp{0};
+  AP4_UI08 m_cryptBlocks{0};
+  AP4_UI08 m_skipBlocks{0};
+  AP4_UI32 m_cryptMode{0};
```
CastagnaIT: m_cryptMode should be initialized with AP4_CENC_CIPHER_NONE

glennguy:
@CastagnaIT I've reworked and done the crypto info struct as you suggested. I think it looks better now.

src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -408,16 +408,17 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
...
-      else if (m_protectedDesc->GetSchemeType() == AP4_PROTECTION_SCHEME_TYPE_CBCS)
-      {
-        m_singleSampleDecryptor->SetEncryptionScheme(ENCRYPTION_SCHEME::CBCS);
-        m_singleSampleDecryptor->SetCrypto(sample_table->GetCryptByteBlock(),
-                                           sample_table->GetSkipByteBlock());
+        m_readerCryptoInfo.m_cryptBlocks = sample_table->GetCryptByteBlock();
+        m_readerCryptoInfo.m_skipBlocks = sample_table->GetSkipByteBlock();
+        m_readerCryptoInfo.m_mode = schemeType == AP4_PROTECTION_SCHEME_TYPE_CENC
+                                        ? AP4_CENC_CIPHER_AES_128_CTR
+                                        : AP4_CENC_CIPHER_AES_128_CBC;
```
CastagnaIT: expand it its better readable
```suggestion
if (schemeType == AP4_PROTECTION_SCHEME_TYPE_CENC)
        m_readerCryptoInfo.m_mode = AP4_CENC_CIPHER_AES_128_CTR;
else
       m_readerCryptoInfo.m_mode = AP4_CENC_CIPHER_AES_128_CBC;
```

src/main.cpp
```diff
@@ -2193,7 +2193,11 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
       memcpy(p->cryptoInfo->kid, pData, 16);
       pData += 16;
       iSize -= (pData - sr->GetSampleData());
+      ReaderCryptoInfo info = sr->GetReaderCryptoInfo();
       p->cryptoInfo->flags = 0;
+      p->cryptoInfo->cryptBlocks = info.m_cryptBlocks;
+      p->cryptoInfo->skipBlocks = info.m_skipBlocks;
+      p->cryptoInfo->mode = info.m_mode;
```
CastagnaIT: here you are assuming that bento4 variables
AP4_CENC_CIPHER_AES_128_CTR/AP4_CENC_CIPHER_AES_128_CBC
are always equal to mediacodec android variables

are we sure that bento4 will always follow this pattern? i wondering what happens if bento4 change values?
do you think it would be better to define our own enum based on android values? or these are universally fixed values?

src/main.cpp
```diff
@@ -2193,7 +2193,11 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
       memcpy(p->cryptoInfo->kid, pData, 16);
       pData += 16;
       iSize -= (pData - sr->GetSampleData());
+      ReaderCryptoInfo info = sr->GetReaderCryptoInfo();
```
CastagnaIT:
```suggestion
      ReaderCryptoInfo cryptoInfo = sr->GetReaderCryptoInfo();
```

src/main.cpp
glennguy: yes we're assuming these values stay the same - equally we're relying on the android side that these values don't change to set the crypto mode either:
```
packet.cryptoInfo->mode == CJNIMediaCodec::CRYPTO_MODE_AES_CBC
                ? CJNIMediaCodec::CRYPTO_MODE_AES_CBC
                : CJNIMediaCodec::CRYPTO_MODE_AES_CTR);

```
So no I could never say universally fixed.... I'll define our own enum in samplereader.h

src/samplereader/SampleReader.h
```diff
@@ -14,6 +14,20 @@
 #include <kodi/AddonBase.h>
 #include <kodi/addon-instance/Inputstream.h>
 
+enum CryptoMode : uint16_t
+{
+  NONE,
+  AES_CTR,
+  AES_CBC
+};
```
CastagnaIT: is preferible to have a class enum over ungrouped old enum, and with fixed values like:

```suggestion
// The crypto modes values must match android MediaCodec values
enum class CryptoMode
{
  NONE=0,
  AES_CTR=1,
  AES_CBC=2
};
```
you can also partial restore the original implementation where SetEncryptionScheme use the enum instead an integer arg variable

the only cast needed to add is in main.cpp

p->cryptoInfo->mode = static_cast<uint16_t>(cryptoInfo.m_mode);

glennguy: Sure thing, done

src/samplereader/SampleReader.h
```diff
@@ -14,6 +14,20 @@
 #include <kodi/AddonBase.h>
 #include <kodi/addon-instance/Inputstream.h>
 
+enum class CryptoMode : uint16_t
```
CastagnaIT, line 19: please add a comment above this enum with something similar to this:
`// The crypto modes values must match android MediaCodec values`
We should start documenting the code better, this will make life easier for us and for future developers

glennguy:
@CastagnaIT If this looks good to you now could you please approve?

CastagnaIT (approved):

Acidzero2020:
it looks like this commit broke the build.

```
/<<PKGBUILDDIR>>/src/main.cpp: In member function ‘virtual DEMUX_PACKET* CInputStreamAdaptive::DemuxRead()’:
/<<PKGBUILDDIR>>/src/main.cpp:2146:22: error: ‘struct DEMUX_CRYPTO_INFO’ has no member named ‘cryptBlocks’
 2146 |       p->cryptoInfo->cryptBlocks = cryptoInfo.m_cryptBlocks;
      |                      ^~~~~~~~~~~
/<<PKGBUILDDIR>>/src/main.cpp:2147:22: error: ‘struct DEMUX_CRYPTO_INFO’ has no member named ‘skipBlocks’
 2147 |       p->cryptoInfo->skipBlocks = cryptoInfo.m_skipBlocks;
      |                      ^~~~~~~~~~
/<<PKGBUILDDIR>>/src/main.cpp:2148:22: error: ‘struct DEMUX_CRYPTO_INFO’ has no member named ‘mode’
 2148 |       p->cryptoInfo->mode = static_cast<uint16_t>(cryptoInfo.m_mode);
      |                      ^~~~
```

https://launchpadlibrarian.net/599274989/buildlog_ubuntu-jammy-amd64.kodi-inputstream-adaptive_6%3A20.2.0-1~jammy_BUILDING.txt.gz

CastagnaIT:
@Acidzero2020 you have to sync your kodi sources to upstream
and rebuild from scratch

Acidzero2020:
Thanks.
As info, the build log is from kodi ppa, it failed there too.

https://launchpad.net/~team-xbmc/+archive/ubuntu/xbmc-nightly/+packages?field.name_filter=inputstream&field.status_filter=published&field.series_filter=

CastagnaIT:
the change has been merged today
i think will be updated next days

PR #892, opened 2022-01-31
Codechandler refactor

glennguy:
Moving all of CodecHandler classes out of main.cpp and renaming member variables


glennguy:
@CastagnaIT I forgot about copyright headers - How do you feel about just copy/paste like the existing ones? You mentioned the other day about another format...

CastagnaIT:
@glennguy i answered on slack, so we can coordinate the works

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
+#include "AVCCodecHandler.h"
+
+AVCCodecHandler::AVCCodecHandler(AP4_SampleDescription* sd)
+  : CodecHandler(sd), m_countPictureSetIds(0), m_needSliceInfo(false)
```
CastagnaIT: on new code the variables should be always initialized with curly brackets
```
AVCCodecHandler::AVCCodecHandler(AP4_SampleDescription* sd)
  : CodecHandler{sd}, m_countPictureSetIds{0}, m_needSliceInfo{false}
```

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+{
+  unsigned int width(0), height(0);
```
CastagnaIT: to avoid in-line variable declaratios then also change with curly brackets

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+    width = video_sample_description->GetWidth();
+    height = video_sample_description->GetHeight();
+  }
+  if (AP4_AvcSampleDescription* avc =
+          AP4_DYNAMIC_CAST(AP4_AvcSampleDescription, m_sampleDescription))
+  {
+    m_extraData.SetData(avc->GetRawBytes().GetData(), avc->GetRawBytes().GetDataSize());
+    m_countPictureSetIds = avc->GetPictureParameters().ItemCount();
+    m_naluLengthSize = avc->GetNaluLengthSize();
+    m_needSliceInfo = (m_countPictureSetIds > 1 || !width || !height);
```
CastagnaIT: avoid handle integer variables as booleans,
maybe:
```suggestion
    m_needSliceInfo = (m_countPictureSetIds > 1 || width > 0 || height > 0);
```

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+bool AVCCodecHandler::ExtraDataToAnnexB()
+{
+  if (AP4_AvcSampleDescription* avc =
+          AP4_DYNAMIC_CAST(AP4_AvcSampleDescription, m_sampleDescription))
+  {
+    //calculate the size for annexb
+    size_t sz(0);
```
CastagnaIT:
```suggestion
    AP4_Size size{0};
```
better use the same type of the getter's/setter's

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+    AP4_Array<AP4_DataBuffer>& sps(avc->GetSequenceParameters());
+    for (unsigned int i(0); i < sps.ItemCount(); ++i)
+      sz += 4 + sps[i].GetDataSize();
+
+    m_extraData.SetDataSize(sz);
+    uint8_t* cursor(m_extraData.UseData());
+
+    for (unsigned int i(0); i < sps.ItemCount(); ++i)
+    {
+      cursor[0] = cursor[1] = cursor[2] = 0;
```
CastagnaIT: in-line assigning should be avoided (here and other cases)

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+    size_t sz(0);
+    AP4_Array<AP4_DataBuffer>& pps(avc->GetPictureParameters());
+    for (unsigned int i(0); i < pps.ItemCount(); ++i)
+      sz += 4 + pps[i].GetDataSize();
+    AP4_Array<AP4_DataBuffer>& sps(avc->GetSequenceParameters());
+    for (unsigned int i(0); i < sps.ItemCount(); ++i)
+      sz += 4 + sps[i].GetDataSize();
+
+    m_extraData.SetDataSize(sz);
+    uint8_t* cursor(m_extraData.UseData());
```
CastagnaIT: ~i suspect here is missing a reinterpreter cast~

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+  return false;
+}
+
+void AVCCodecHandler::UpdatePPSId(AP4_DataBuffer const& buffer)
+{
+  if (!m_needSliceInfo)
+    return;
+
+  //Search the Slice header NALU
+  const AP4_UI08* data(buffer.GetData());
```
CastagnaIT: also here seem missing a cast

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+  unsigned int data_size(buffer.GetDataSize());
```
CastagnaIT:
```suggestion
  AP4_Size data_size(buffer.GetDataSize());
```

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+  for (; data_size;)
+  {
+    // sanity check
+    if (data_size < m_naluLengthSize)
```
CastagnaIT: here compare different data types,
m_naluLengthSize should need a cast into data_size type

src/codechandler/AVCCodecHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+    // get the next NAL unit
+    AP4_UI32 nalu_size;
+    switch (m_naluLengthSize)
+    {
+      case 1:
+        nalu_size = *data++;
+        data_size--;
+        break;
+      case 2:
+        nalu_size = AP4_BytesToInt16BE(data);
```
CastagnaIT: look at method output (here and case below)
in this case `AP4_SI16` the `_SI` prefix means signed integer
but here we assign it on a unsigned variable, then miss a cast

src/codechandler/CodecHandler.cpp
```diff
@@ -0,0 +1,48 @@
...
+          (!info.GetSampleRate() && asd->GetSampleRate() != info.GetSampleRate()) ||
+          (!info.GetBitsPerSample() && asd->GetSampleSize() != info.GetBitsPerSample()))
+      {
+        if (!info.GetChannels())
+          info.SetChannels(asd->GetChannelCount());
+        if (!info.GetSampleRate())
+          info.SetSampleRate(asd->GetSampleRate());
+        if (!info.GetBitsPerSample())
+          info.SetBitsPerSample(asd->GetSampleSize());
+        return true;
```
CastagnaIT: to replace integer value's handled as booleans
e.g.:
`if ((!info.GetChannels() && asd->GetChannelCount() != info.GetChannels()) ||`
to
`if ((info.GetChannels() == 0 && asd->GetChannelCount() != info.GetChannels()) ||`

src/codechandler/CodecHandler.cpp
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+    else
+    {
+      //Netflix Framerate
+      AP4_Atom* atom;
+      AP4_UnknownUuidAtom* nxfr;
+      static const AP4_UI08 uuid[16] = {0x4e, 0x65, 0x74, 0x66, 0x6c, 0x69, 0x78, 0x46,
```
CastagnaIT: you are know to what is referred this fixed UUID? or from what come from?
i think that this is a good candidate to be moved as constexpr in a unnamed namespace on top of this file

src/codechandler/CodecHandler.cpp
```diff
@@ -0,0 +1,48 @@
...
+                                        0x72, 0x61, 0x6d, 0x65, 0x52, 0x61, 0x74, 0x65};
+
+      if ((atom =
+               m_sampleDescription->GetDetails().GetChild(static_cast<const AP4_UI08*>(uuid), 0)) &&
+          (nxfr = dynamic_cast<AP4_UnknownUuidAtom*>(atom)) && nxfr->GetData().GetDataSize() == 10)
+      {
+        AP4_UI16 fpsRate = nxfr->GetData().GetData()[7] | nxfr->GetData().GetData()[6] << 8;
+        AP4_UI16 fpsScale = nxfr->GetData().GetData()[9] | nxfr->GetData().GetData()[8] << 8;
```
CastagnaIT: since we handle fpsRate fpsScale as `unsigned int` these vars should be declared as `unsigned int`

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+bool HEVCCodecHandler::ExtraDataToAnnexB()
+{
+  if (AP4_HevcSampleDescription* hevc =
+          AP4_DYNAMIC_CAST(AP4_HevcSampleDescription, m_sampleDescription))
+  {
+    const AP4_Array<AP4_HvccAtom::Sequence>& sequences = hevc->GetSequences();
+
+    if (!sequences.ItemCount())
```
CastagnaIT:
```suggestion
    if (sequences.ItemCount() == 0)
```

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+    {
+      kodi::Log(ADDON_LOG_WARNING, "No available sequences for HEVC codec extra data");
+      return false;
+    }
+
+    //calculate the size for annexb
+    size_t sz(0);
```
CastagnaIT:
```suggestion
    AP4_Size size{0};
```

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+              m_extraData.GetDataSize());
+    return true;
+  }
+  kodi::Log(ADDON_LOG_WARNING, "No HevcSampleDescription - annexb extradata not available");
+  return false;
+}
+
+bool HEVCCodecHandler::GetInformation(kodi::addon::InputstreamInfo& info)
+{
+  if (!info.GetFpsRate())
```
CastagnaIT:
```suggestion
  if (info.GetFpsRate() == 0)
```
also on all other cases below

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+  {
+    if (AP4_HevcSampleDescription* hevc =
+            AP4_DYNAMIC_CAST(AP4_HevcSampleDescription, m_sampleDescription))
+    {
+      bool ret = false;
+      if (hevc->GetAverageFrameRate())
+      {
+        info.SetFpsRate(hevc->GetAverageFrameRate());
+        info.SetFpsScale(256);
```
CastagnaIT, line 80: you know why value 256? (set also below)
i think could be a candidate for a constexpr in a unnamed namespace

src/codechandler/WebVTTCodecHandler.h
```diff
@@ -0,0 +1,22 @@
...
+  bool TimeSeek(AP4_UI64 seekPos) override { return true; }
+  void Reset() override {}
+
+private:
+  AP4_UI64 m_ptsOffset;
+  AP4_DataBuffer m_data;
+  AP4_UI64 m_pts{0};
+  AP4_UI32 m_duration{0};
+  const AP4_Byte m_extraDataFILE[4] = {'f', 'i', 'l', 'e'};
+  const AP4_Byte m_extraDataFMP4[4] = {'f', 'm', 'p', '4'};
```
CastagnaIT: m_extraDataFILE m_extraDataFMP4 need to be moved as costexpr in an unnamed namespace (on top of the relative .cpp)

CastagnaIT (commented):
some first requests, I understand there are many,
but i think that with the refactor we need to improve the code a little

CMakeLists.txt
```diff
@@ -8,6 +8,13 @@ find_package(Kodi REQUIRED)
 
 set(ADP_SOURCES
 	src/main.cpp
+	src/codechandler/AVCCodecHandler.cpp
+	src/codechandler/CodecHandler.cpp
+	src/codechandler/HEVCCodecHandler.cpp
+	src/codechandler/MPEGCodecHandler.cpp
+	src/codechandler/TTMLCodecHandler.cpp
+	src/codechandler/VP9CodecHandler.cpp
+	src/codechandler/WebVTTCodecHandler.cpp
```
CastagnaIT, line 17: all these should be moved in an appropriate CMakeLists file in codechandler folder

src/codechandler/AVCCodecHandler.cpp
CastagnaIT: Let me explain better, it is not necessary to rewrite every line of code, only for new lines or variables rewritten from scratch

src/codechandler/AVCCodecHandler.cpp
CastagnaIT: i correct myself, for now keep it as is

src/codechandler/CodecHandler.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+#include <bento4/Ap4.h>
```
CastagnaIT, line 11: missing space between #pragma once and include

src/codechandler/AVCCodecHandler.h
```diff
@@ -0,0 +1,16 @@
+#include "CodecHandler.h"
```
CastagnaIT, line 9: in this and many other .h headers is missing
`#pragma once` on top
there is a reason behind?

src/codechandler/AVCCodecHandler.cpp
glennguy: I don't think so, this is just unsigned int compared to char - from what I read integral promotion will convert automatically.

src/codechandler/CodecHandler.cpp
glennguy: It's literally `NetflixFrameRate`
Maybe there's a nicer way via helper function to make this, if there's other examples that need it too

CMakeLists.txt
glennguy: Why?
My cmake is very basic but I think that then requires creating a new project, adding as dependency in main cmakelists... 
If the goal is to make it look less 'ugly' then we can do as other add-ons do and make an effective wildcard of all *.cpp/*.h files.

src/codechandler/AVCCodecHandler.h
glennguy: Yes.
It's not required unless the file is going be be #included more than once.

src/codechandler/CodecHandler.cpp
glennguy: how about:
```
AP4_Atom* atom;
      AP4_UnknownUuidAtom* nxfr;
      static const std::string uuid = "NetflixFrameRate";
      atom = m_sampleDescription->GetDetails().GetChild(
          reinterpret_cast<const AP4_UI08*>(uuid.c_str()));
      if (atom && (nxfr = dynamic_cast<AP4_UnknownUuidAtom*>(atom)) &&
          nxfr->GetData().GetDataSize() == 10)
```
Can then move uuid to unnamed namespace, easier to read

CMakeLists.txt
CastagnaIT: I was thinking if with the future refactor's we could have many new files and this list starts to grow
but no problem if appropriate could be considered in the future

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+    if (!sequences.ItemCount())
+    {
+      kodi::Log(ADDON_LOG_WARNING, "No available sequences for HEVC codec extra data");
+      return false;
+    }
+
+    //calculate the size for annexb
+    size_t sz(0);
+    for (const AP4_HvccAtom::Sequence *b(&sequences[0]), *e(&sequences[sequences.ItemCount()]);
```
glennguy: For here I'm thinking of rewriting a bit more succinctly:
 ```
for (unsigned int i{0}; i < sequences.ItemCount(); ++i)
      for (unsigned int j{0}; j < (&sequences[i])->m_Nalus.ItemCount(); ++i)
        sz += (4 + (&sequences[i])->m_Nalus[j].GetDataSize());
```
@CastagnaIT does this seem better to you?

File: CMakeLists.txt (hunk 19 above)
Comment (glennguy) [reply CMakeLists.txt:17]: Sure thing. For next refactor PR I'll change to the wildcard way, should then just be adding the folders.

File: src/codechandler/HEVCCodecHandler.cpp (hunk 17 above)
Comment (glennguy) [reply src/codechandler/HEVCCodecHandler.cpp:80]: I can only guess that HEVC doesn't have a rate and scale, just a rate.
@kszaq would you be able to help us here?

File: src/codechandler/AVCCodecHandler.cpp (hunk 7 above)
Comment (glennguy) [reply src/codechandler/AVCCodecHandler.cpp]: I think this one makes no difference - AP4_Byte == AP4_UI08 == unsigned char
Can change to AP4_Byte though to at least be more 'proper'

File: src/codechandler/HEVCCodecHandler.cpp (hunk 22 above)
Comment (CastagnaIT) [reply src/codechandler/HEVCCodecHandler.cpp]: good thing better readable
please on `for` loops always add brackets,
relying on spaces in this case is not a good thing
a little space oversights can cause hidden problems

on last line i think there is a non needed parenthesis:
`sz += 4 + (&sequences[i])->m_Nalus[j].GetDataSize();`
since we have space rename `sz` to `size`

File: src/codechandler/CodecHandler.cpp [hunk 23]
```diff
@@ -0,0 +1,63 @@
...
+        {
+          info.SetFpsScale(fpsScale);
+          info.SetFpsRate(fpsRate);
+          return true;
+        }
+      }
+    }
+  }
+  return false;
+};
```
CastagnaIT: missing empty end line

src/codechandler/CodecHandler.h
```diff
@@ -0,0 +1,40 @@
...
+  virtual void SetPTSOffset(AP4_UI64 offset){};
+  virtual bool TimeSeek(AP4_UI64 seekPos) { return true; };
+  virtual void Reset(){};
+
+  AP4_SampleDescription* m_sampleDescription;
+  AP4_DataBuffer m_extraData;
+  AP4_UI08 m_naluLengthSize;
+  AP4_UI08 m_pictureId;
+  AP4_UI08 m_pictureIdPrev;
+};
```
CastagnaIT: missing empty end line

src/codechandler/TTMLCodecHandler.h
```diff
@@ -0,0 +1,26 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "../parser/TTML.h"
```
CastagnaIT: do you think is possible replace all these "..." with a full path reference?

src/codechandler/CodecHandler.cpp
CastagnaIT: it was so simple I hadn't thought...
leave it as it is for now we avoid conversions but move the uuid to unnamed namespace
suggest only to add in the constexpr variable a little comment to explain that the value is "NetflixFrameRate" for posterity
if we will find others similar cases we will think to create an utils method

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+bool HEVCCodecHandler::GetInformation(kodi::addon::InputstreamInfo& info)
+{
+  if (info.GetFpsRate() == 0)
+  {
+    if (AP4_HevcSampleDescription* hevcSampleDescription =
+            AP4_DYNAMIC_CAST(AP4_HevcSampleDescription, m_sampleDescription))
+    {
+      bool ret = false;
+      if (hevcSampleDescription->GetAverageFrameRate())
```
CastagnaIT:
```suggestion
      if (hevcSampleDescription->GetAverageFrameRate() > 0)
```

src/codechandler/HEVCCodecHandler.cpp
```diff
@@ -0,0 +1,90 @@
...
+      {
+        info.SetFpsRate(hevcSampleDescription->GetAverageFrameRate());
+        info.SetFpsScale(256);
+        ret = true;
+      }
+      else if (hevcSampleDescription->GetConstantFrameRate())
```
CastagnaIT:
```suggestion
      else if (hevcSampleDescription->GetConstantFrameRate() > 0)
```

src/codechandler/AVCCodecHandler.cpp
CastagnaIT: forgot this

glennguy: naluSize is now AP4_SI32 so should be ok as it's a promotion?

src/codechandler/HEVCCodecHandler.cpp
kszaq: Unfortunately I don't know why 256. The constant was introduced in this commit: https://github.com/xbmc/inputstream.adaptive/commit/8bfd456a556ab43976c2023ffbd2eb2e48a965cf

It looks like HVCC atom does not provide fps scale, instead it gives only average or constant frame rate. The constant 256 scale value is probably defined in MPEG-4 Part 15.

src/codechandler/TTMLCodecHandler.h
glennguy: What I think is better here is to move TTML.h into the codechandler folder as it's only used by this file anyway.

src/codechandler/CodecHandler.cpp
glennguy: I've done my own test to see that the data is the same in both scenarios but if you feel it's safer I can revert and just move the variable as suggested?

src/codechandler/HEVCCodecHandler.cpp
glennguy: Ah ok, thanks!

src/codechandler/CodecHandler.cpp
CastagnaIT: not so good, in an unnamed namespace, `static` it is irrilevant, and you must use `constexpr` on variables,
since `"NetflixFrameRate"` is already a `const char*` there is no reason to convert it into a string object to be re-converted as `const char*`, then should be defined as:

`constexpr const char* NETFLIX_FRAMERATE_UUID = "NetflixFrameRate";`

and remove `.c_str()` on reinterpreter cast, suggest to use also full name ^^
edit: forgot to say that constexpr var must be uppercase

src/codechandler/TTMLCodecHandler.h
CastagnaIT: for me ok, but IMO also better under a `/codechandler/ttml/` folder,
which will have to be dropped in the future...,
i had started to think/create the TTML parser on Kodi core as done with WebVTT,
but it is really complex format with tons of features, which have to be compared with the limitations of libass renderer,
W3C define allow extensive customisation of implementation but i am still in difficulty on how to create a first parser version

src/codechandler/AVCCodecHandler.cpp
CastagnaIT: now that i look better, if take a look at bento4 appears that the called method are only wrappers,
then we can change the called methods as:
`AP4_BytesToInt16BE`  >to>  `AP4_BytesToUInt16BE`
`AP4_BytesToInt32BE` >to> `AP4_BytesToUInt32BE`

then you can keep naluSize type as in origin as `AP4_UI32` type
in this way we avoid unuseful casts

glennguy: Awesome, thanks for that!

src/codechandler/TTMLCodecHandler.h
glennguy: Sounds good. I've moved into ttml folder and fixed formatting. Should we do a cleanup on this or leave?

src/codechandler/CodecHandler.cpp
```diff
@@ -0,0 +1,63 @@
...
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "CodecHandler.h"
+
+namespace
+{
+constexpr const char* netflixFramerateUuid = "NetflixFrameRate";
```
CastagnaIT: sorry for the boredom but I edited my comment too late 
constexpr vars name must be uppercase like old "static const"
NETFLIX_FRAMERATE_UUID

src/codechandler/WebVTTCodecHandler.cpp
```diff
@@ -0,0 +1,58 @@
...
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "WebVTTCodecHandler.h"
+
+namespace
+{
+constexpr AP4_Byte extraDataFILE[4] = {'f', 'i', 'l', 'e'};
```
CastagnaIT: also here uppercase e.g.
EXTRADATA_FILE

src/codechandler/TTMLCodecHandler.h
CastagnaIT: not strictly necessary as they will have to be deleted
i could do it in a separate PR i would like to make a refactor of the helper class in to different files
(to add other helper methods to cleanup this)
i will think of something i will write to you on slack if not today next days

src/codechandler/CodecHandler.cpp
glennguy: No need for apologies, thanks for your patience with me!

CastagnaIT (approved):
For me it's now ready
if possible before merging this, merge SPDX PR

Issue #893, opened 2022-01-31
Dash: problem opening stream containing multiple $RepresentationID$ in URL


gizmocuz:
There are streams with multiple **$RepresentationID$** in the URL like:

https://dash.akamaized.net/akamai/bbb_30fps/bbb_30fps.mpd

Sollution is to change the following function

```
static void ReplacePlaceHolders(std::string& rep, const std::string& id, uint32_t bandwidth)
{
    while (1)
    {
        std::string::size_type repPos = rep.find("$RepresentationID$");
        if (repPos == std::string::npos)
            break;
        rep.replace(repPos, 18, id);
    }
    std::string::size_type repPos = rep.find("$Bandwidth$");
    if (repPos != std::string::npos)
    {
        char bw[32];
        sprintf(bw, "%u", bandwidth);
        rep.replace(repPos, 11, bw);
    }
}

```

gizmocuz:
I created 3 PR's ... Not sure if this is the right way to do this when a project has 3 branches

basilgello:
Right now, only master (Nexus) and Matrix are supported Kodi versiobs. Leia is no more supported, in case you made a PR targeting 18.x.

glennguy:
Fixed in #898

PR #894, opened 2022-01-31

gizmocuz:
See issue #893

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -236,17 +236,8 @@ bool ParseContentProtection(const char** attr, DASHTree* dash)
 
 static void ReplacePlaceHolders(std::string& rep, const std::string& id, uint32_t bandwidth)
 {
-  std::string::size_type repPos = rep.find("$RepresentationID$");
-  if (repPos != std::string::npos)
-    rep.replace(repPos, 18, id);
-
-  repPos = rep.find("$Bandwidth$");
-  if (repPos != std::string::npos)
-  {
-    char bw[32];
-    sprintf(bw, "%u", bandwidth);
-    rep.replace(repPos, 11, bw);
-  }
+  replaceAll(rep, "$RepresentationID$", id, false);
+  replaceAll(rep, "$Bandwidth$", std::to_string(bandwidth), false);
 }
 
 static void XMLCALL start(void* data, const char* el, const char** attr)


src/parser/DASHTree.cpp
```diff
@@ -236,11 +236,14 @@ bool ParseContentProtection(const char** attr, DASHTree* dash)
 
 static void ReplacePlaceHolders(std::string& rep, const std::string& id, uint32_t bandwidth)
 {
-  std::string::size_type repPos = rep.find("$RepresentationID$");
-  if (repPos != std::string::npos)
+  while (1)
```
CastagnaIT: on new changes on Nexus branch, is needed to clean a bit the code,
please replace existing method `replaceAll` in `helpers.cpp` with this:

```
int ReplaceAll(std::string &source, const std::string &oldStr, const std::string &newStr)
{
  if (oldStr.empty())
    return 0;

  int replacedChars = 0;
  size_t index = 0;

  while (index < source.size() && (index = source.find(oldStr, index)) != std::string::npos)
  {
    source.replace(index, oldStr.size(), newStr);
    index += newStr.size();
    replacedChars++;
  }

  return replacedChars;
}
```
then use `ReplaceAll` helper method to replace the "$RepresentationID$"
and use `replace` helper method to replace the first occurence of "$Bandwidth$"

CastagnaIT (changes requested):

src/parser/DASHTree.cpp
matthuisman: multiple $Bandwidth$ I guess would be possible as well, so maybe a good idea to use the ReplaceAll for both?

src/parser/DASHTree.cpp
```diff
@@ -236,11 +236,14 @@ bool ParseContentProtection(const char** attr, DASHTree* dash)
...
+    if (repPos == std::string::npos)
+        break;
     rep.replace(repPos, 18, id);
-
-  repPos = rep.find("$Bandwidth$");
+  }
+  std::string::size_type repPos = rep.find("$Bandwidth$");
   if (repPos != std::string::npos)
   {
     char bw[32];
```
CastagnaIT: since you have to change the replace methods,
replace also `sprintf` with `std::to_string` to convert the number to string

matthuisman:
Maybe also consider adding new tests for multiple same tags etc

There is already some tests for template tags, so you could use as a template for your new tests

src/parser/DASHTree.cpp
```diff
@@ -236,17 +236,8 @@ bool ParseContentProtection(const char** attr, DASHTree* dash)
...
-    rep.replace(repPos, 18, id);
-
-  repPos = rep.find("$Bandwidth$");
-  if (repPos != std::string::npos)
-  {
-    char bw[32];
-    sprintf(bw, "%u", bandwidth);
-    rep.replace(repPos, 11, bw);
-  }
+  replaceAll(rep, "$RepresentationID$", id, true);
```
CastagnaIT: if you set nextEmpty to true only the first case will be replaced all others will be deleted

src/parser/DASHTree.cpp
```diff
@@ -236,17 +236,8 @@ bool ParseContentProtection(const char** attr, DASHTree* dash)
...
-
-  repPos = rep.find("$Bandwidth$");
-  if (repPos != std::string::npos)
-  {
-    char bw[32];
-    sprintf(bw, "%u", bandwidth);
-    rep.replace(repPos, 11, bw);
-  }
+  replaceAll(rep, "$RepresentationID$", id, false);
+  replaceAll(rep, "$Bandwidth$", std::to_string(bandwidth).c_str(), false);
```
CastagnaIT: why you pass a const char .c_str() when the method accept string?

gizmocuz:
Sorry for all the extra commits... most of the code works with char* and some parts are already modernized...

Regarding the extra tests, I think it's totally covered and will always return true

Regarding the replaceAll function that sometimes does not replace all, maybe in the Nexus branch we can rewrite this function, skip the last argument (as it's only used in DashTree.cpp and it should also replace 'all') to

```
	while (1)
	{
		const int pos = s.find(from);
		if (pos == -1) break;
		s.replace(pos, from.size(), to);
	}

```

Is it possible to squash and merge the PR's ?

CastagnaIT:
no you have to squash your commits yourself before the merge

CastagnaIT:
> Regarding the replaceAll function that sometimes does not replace all, maybe in the Nexus branch we can rewrite this function, skip the last argument (as it's only used in DashTree.cpp and it should also replace 'all') to

@gizmocuz currently we are already working on a refactor on this branch, therefore it does not matter

gizmocuz:
Thanks, okey let me see how I can squash this without github

CastagnaIT:
to make changes in the project you should use shell `git`
`github` not allow full control of the versioning system

`git reset --soft HEAD^` to each commit
`git commit -a` to add the changes pending in a commit
`git push -f yourreponame/Nexus` to force update changes in your upstream github repo

gizmocuz:
Tried and failing... there should be a green button here to merge PR's, it's a dropdown, one option is to squash it

CastagnaIT:
> Tried and failing... there should be a green button here to merge PR's, it's a dropdown, one option is to squash it

this is not the way that Team Kodi work with projects,
reread my previous amended comment and try again

CastagnaIT:
@gizmocuz you were on the right track but you have resetted too much commits...
the easy way to fix this mess is force resync your branch with the original branch
and make your changes again

`git reset --hard origin/Nexus` to reset everything in your branch with original branch (OFC you lose every your changes)

i do not know if your origin repo is called "origin" if needed change this with right repo name

gizmocuz:
Okey now everything is broke... I don't understand why you can not do a squash in github ... 
Closing this, making a new one for both... taking to much time for things that can be done in in github directly

CastagnaIT:
look at changes now PR seem OK

gizmocuz:
I was expecting all these comments to be removed as well... But i assumed wrong....
But above command was not working, it asked 'more?' ... Must be a windows thing...

git reset --soft HEAD~1 works...

Still why not just use the green button here on GitHub to squash ?

`When you select the Squash and merge option on a pull request on GitHub.com, the pull request's commits are squashed into a single commit. Instead of seeing all of a contributor's individual commits from a topic branch, the commits are combined into one commit and merged into the default branch.`

That was ecactly the idea... lost a good 30 minutes on this and hope to never have to do that again **lol** ;-)

CastagnaIT:
> Still why not just use the green button here on GitHub to squash ?

Squash button not allow to create the merge commit, and your superfluous commits will remain in the verisioning history
so also for ease of consultation, we want to keep the branch clean

PR #897, opened 2022-01-31
SPDX license identifier and updated copyright

CastagnaIT:
Cleanup all copyrights and replaced with SPDX license identifier

I have followed the same cleanup has been done in xbmc repository and reused same policy/files

Operations done initially by check the licenses on each file by using https://github.com/nexB/scancode-toolkit
cmd: `scancode -l -n 16 ../kodi --html ../kodi_licenses.html`

a lot of false positives from the scan but not bad this helped me a lot, 
after that verified and changed each file, at first i thought it would take longer...

Current licenses under inputstream.adaptive:

| Identifier | Description |
| --------- | ---------------- | 
| BSD-2-Clause-Views | BSD 2-Clause with views sentence License |
| BSD-3-Clause | BSD 3-clause "New" or "Revised" License |
| google-patent-license-webm | Google Patent License for WebM |
| GPL-2.0-or-later | GNU General Public License v2.0 or later |
| RSA-MD5 | RSA Message-Digest License |

The only folder that i preferred not touch is: `lib/webm_parser`
that seems files are a copy of part of google repo


glennguy (approved):
All looks good!

PR #898, opened 2022-02-04

gizmocuz:
fix issue #893


CastagnaIT (approved):

CastagnaIT:
@matthuisman last word to you if the test case it is really required
since the "replace" method replace all anyway

glennguy (approved):
Thanks @gizmocuz

PR #899, opened 2022-02-04

gizmocuz:
See #899


CastagnaIT (approved):

matthuisman:
So this would require @CastagnaIT util refactor PR to also be backported to 19?

If that's not going to happen, then maybe 19 should use the old in-place fix you originally proposed so 19 at least works and then the replaceall is used in 20

Edit:
Sorry, ignore me. I see replaceall is an existing helper function. I thought it was a new function added by Casta. Nevermind :)

glennguy (approved):

PR #901, opened 2022-02-08
Refactor generic helpers to new utils

CastagnaIT:
Refactor of the generic helper file in more organized utils files
each file has its own namespace, in a similar way to the xbmc utils

Has been removed some string helper methods in favour of the use of `kodi::tools` string utils provided by Kodi interface

This refactor covers where possible only minor changes/adjustments to the code,
major code improvements on specified parts of code can be done with other PR's


wvdecrypter/wvdecrypter.cpp
```diff
@@ -1424,7 +1453,7 @@ class WVDecrypter : public SSD_DECRYPTER
     return cdmsession_->GetCdmAdapter() != nullptr;
   }
 
-  virtual AP4_CencSingleSampleDecrypter *CreateSingleSampleDecrypter(AP4_DataBuffer &pssh, const char *optionalKeyParameter, const uint8_t *defaultkeyid, bool skipSessionMessage) override
+  virtual AP4_CencSingleSampleDecrypter *CreateSingleSampleDecrypter(AP4_DataBuffer &pssh, const char *optionalKeyParameter, std::string_view defaultkeyid, bool skipSessionMessage) override
```
glennguy: Line width

wvdecrypter/wvdecrypter_android.cpp
```diff
@@ -1243,7 +1264,7 @@ class WVDecrypter : public SSD_DECRYPTER, public jni::CJNIMediaDrmOnEventListene
     return cdmsession_->GetMediaDrm();
   }
 
-  virtual AP4_CencSingleSampleDecrypter *CreateSingleSampleDecrypter(AP4_DataBuffer &pssh, const char *optionalKeyParameter, const uint8_t *defaultkeyid, bool skipSessionMessage) override
+  virtual AP4_CencSingleSampleDecrypter *CreateSingleSampleDecrypter(AP4_DataBuffer &pssh, const char *optionalKeyParameter, std::string_view defaultkeyid, bool skipSessionMessage) override
```
glennguy: line width

CastagnaIT:
I had to make some changes to how data are handled on `CreateISMlicense` method:
https://github.com/xbmc/inputstream.adaptive/pull/901/files#diff-324890c574efe975104680bc3111381b4e066c8546890d98966cfdfdd3254cedR222
this method IMO appear so ugly, perhaps it could be greatly simplified by handling everything as a string,
ATM i do not have sufficient knowledge of what data are processed to do it myself,
but for now it is sufficient to test a stream to check no regressions,
@glennguy if you know/can a test stream to test, i do not remember which stream it uses

wvdecrypter/wvdecrypter_android.cpp
```diff
@@ -266,7 +280,7 @@ class WV_CencSingleSampleDecrypter : public AP4_CencSingleSampleDecrypter
 {
 public:
   // methods
-  WV_CencSingleSampleDecrypter(WV_DRM &drm, AP4_DataBuffer &pssh, const char *optionalKeyParameter, const uint8_t* defaultKeyId);
+  WV_CencSingleSampleDecrypter(WV_DRM &drm, AP4_DataBuffer &pssh, const char *optionalKeyParameter, std::string_view defaultKeyId);
```
glennguy: Line width

glennguy (changes requested):

glennguy:
~~Here's an ISM to test:~~ nvm, no protection...
```
#KODIPROP:inputstream.adaptive.manifest_type=ism
http://playready.directtaps.net/smoothstreaming/SSWSS720H264/SuperSpeedway_720.ism/Manifest
```

Seems impossible to find the combination of widevine/ISM for testing.

HGo uses it in this issue https://github.com/xbmc/inputstream.adaptive/issues/853

glennguy:
One last thing, this should be bumped since we're modifying the class: https://github.com/CastagnaIT/inputstream.adaptive/blob/92309b6fff17e1e25ac6f941c1957b9a940ec686/src/SSD_dll.h#L45

I know in 2022 pretty much no one would be mixing and matching versions especially since Android now works without being bundled in the apk but it this case it would prevent random crashing and instead the module just doesn't load.

CastagnaIT:
@enen92 i need to test ISM to check no regressions,
have you the possibility to test this with HGo (HBO)? (artifact are on jenkins)

glennguy:
[Manifest.txt](https://github.com/xbmc/inputstream.adaptive/files/8047356/Manifest.txt)

@CastagnaIT alternatively you could quickly mock up a test using all or part of this manifest (taken from the issue log)
Not saying that a test is required for this PR, that can come later.

glennguy (approved):

enen92:
@CastagnaIT sorry for the delay, yeah I can test it. Do you have android builds I can test?

glennguy:
@enen92 yes builds are here https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-901/13/artifacts

enen92:
Sorry for the delay, just now I've found a bit of time to test this. Unfortunately it doesn't work with HGO eu, player starts but goes into a black screen. The version from the official repository works which means there's something wrong with the implementation.

Log from this PR:
https://paste.kodi.tv/ilalilebol

src/SSD_dll.h
```diff
@@ -190,7 +196,7 @@ namespace SSD
     // Return supported URN if type matches to capabilities, otherwise null
     virtual const char *SelectKeySytem(const char* keySystem) = 0;
     virtual bool OpenDRMSystem(const char *licenseURL, const AP4_DataBuffer &serverCertificate, const uint8_t config) = 0;
-    virtual AP4_CencSingleSampleDecrypter *CreateSingleSampleDecrypter(AP4_DataBuffer &pssh, const char *optionalKeyParameter, const uint8_t *defaultkeyid, bool skipSessionMessage) = 0;
+    virtual AP4_CencSingleSampleDecrypter *CreateSingleSampleDecrypter(AP4_DataBuffer &pssh, const char *optionalKeyParameter, std::string_view defaultkeyid, bool skipSessionMessage) = 0;
```
enen92, line 199: missing const modifier for string_view?

CastagnaIT: i have made a bit of tests with string_view, but since it already encapsulate a const readonly value the only effect to apply "const std::string_view" is that the string_view functions become unusable

CastagnaIT:
@glennguy i have tried implement the smooth test and compare the reworked `CreateISMlicense` medtod output with the original method, in this way:
https://github.com/CastagnaIT/inputstream.adaptive/commit/6243304e43e9a682bfe89951bbd22b39fb27ccb9
try take a look

but as is the test result say that the output of the two methods are the same,
see `SmoothTreeTest.ISMLoadLicenseTest` on `Run tests` page:
https://github.com/CastagnaIT/inputstream.adaptive/runs/5232026495?check_suite_focus=true

@enen92 can you try play video with the regular nightly build? from here https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/Nexus/59/artifacts
I would like to be sure that the problem does not come from the Nexus branch

enen92:
@CastagnaIT just tested with master (link you shared) and streams play fine :/ I can try to share my credentials later (temporarily if you're interested in runtime testing it)

CastagnaIT:
@enen92 i have recheck and i have found a little oversight in kid conversion, not sure if this is the cause but influence the `CreateISMlicense`, if you can do a last test:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-901/14/artifacts

if it still doesn't work if you can send me temporary credentials,
i will have to exclude all changes one by one...

enen92:
@CastagnaIT nice one, now it works :)

CastagnaIT:
finally works thanks!
i will make the final refinements and after the PR will be ready

CastagnaIT:
@glennguy the PR is fixed and ready

glennguy:
Fantastic work @CastagnaIT 👍

PR #907, opened 2022-02-18
Update some pointers to use smart pointers

CastagnaIT:
Remove the `SAFE_DELETE` macro in favour of `unique_ptr`,
this apply also a bit of cleanup on the name of members that they were too generic
and add missing code check for null pointers to avoid crashes, before unmanaged

sidenote for future cleanups:
`struct STREAM` in main.h is placed inside `Session` class
since this struct is handled as a regular class, i think should be changed as `class Stream` and placed as stand-alone outside `Session` class,
maybe all "session" things like these should be moved within a proper "SESSION" namespace

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -53,13 +53,6 @@
 
 #define STREAM_TIME_BASE 1000000
 
-#define SAFE_DELETE(p) \
-  do \
-  { \
-    delete (p); \
-    (p) = NULL; \
-  } while (0)
-
 using namespace UTILS;
 using namespace kodi::tools;
 
@@ -213,16 +206,17 @@ private:
 Bento4 Streams
 ********************************************************/
 
-class ATTR_DLL_LOCAL AP4_DASHStream : public AP4_ByteStream
+class ATTR_DLL_LOCAL AdaptiveByteStream : public AP4_ByteStream
 {
 public:
   // Constructor
-  AP4_DASHStream(adaptive::AdaptiveStream* stream) : stream_(stream){};
+  AdaptiveByteStream(adaptive::AdaptiveStream* adStream) : m_adStream(adStream){};
+  virtual ~AdaptiveByteStream() {};
 
   // AP4_ByteStream methods
   AP4_Result ReadPartial(void* buffer, AP4_Size bytesToRead, AP4_Size& bytesRead) override
   {
-    bytesRead = stream_->read(buffer, bytesToRead);
+    bytesRead = m_adStream->read(buffer, bytesToRead);
     return bytesRead > 0 ? AP4_SUCCESS : AP4_ERROR_READ_FAILED;
   };
   AP4_Result WritePartial(const void* buffer,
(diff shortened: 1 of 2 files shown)


garbear:
Just hopping in to say, make sure you don't make heavy use of smart pointers inside tight loops. I've had the overhead of STL shared pointers kill the performance of a decoding thread.

CastagnaIT:
is this a past experience, or have you tested this PR?

almost all smartptr are created when ISA is initialised before playback,
IMO maybe `std::vector<std::unique_ptr<STREAM>> m_streams;` could cause some overhead when ISA starts
on this will be created a smartptr for each stream/bitrate loaded
there could be as many as 2 or 10 or 30 streams etc... (depends how many video audio subs...)

instead when OpenStream() is fired this create 3 smartptr
this method is called 3 times (only) on playback start (audio-video-subs) and for bitrate change during playback but does not happen frequently

during playback what might happen most of the time is call the `get` for the pointer
i mean e.g. this call https://github.com/xbmc/inputstream.adaptive/pull/907/files#diff-910d89612d74e91ae70ed40289b3910b1c1a09b1f5a1bb0b15849f70760cbba2R135

the decoding is not managed here but separately (with wvdecrypter.cpp) which should not affect
surely @glennguy can explain better and/or amend me

if for the `std::vector<std::unique_ptr<STREAM>> m_streams;` generate too much smartpointers i can revert back for this case

CastagnaIT:
hovewer a little search the `std::unique_ptr` has very little overhead almost like raw pointers, the problem is on `std::shared_ptr` use
see: https://www.modernescpp.com/index.php/memory-and-performance-overhead-of-smart-pointer

this suggests that we can keep it that way

garbear:
I didn't look at the diff that closely. unique_ptr should have no overhead.

src/main.cpp
```diff
@@ -1229,8 +1232,8 @@ class ATTR_DLL_LOCAL SubtitleSampleReader : public SampleReader
 
   AP4_Sample m_sample;
   AP4_DataBuffer m_sampleData;
-  AP4_ByteStream* m_input{nullptr};
-  KodiAdaptiveStream* m_adaptiveStream{nullptr};
```
glennguy, line 1233: Just a thought, doesn't have to be for this PR but thought I'd mention:
I know this is renamed to match the derived class, my question is do we really need this derived class at all? Yes, in Nexus now this derived class holds the rep chooser and also having download functions defined here made creating the test classes a little easier for me but I can't see why it was made this way?

glennguy:
Looks good

glennguy (approved):

src/main.cpp
CastagnaIT: I have not check it in depth, but i agree that we could remove this derived class and merge it into the base class

glennguy: Cool. We'll bring this in for now as-is and look at that later

PR #912, opened 2022-03-05
[cleanup] Moved kodi properties parser to utils

CastagnaIT:
Cleanup to kodi properties parser and relative little cleanup to related code/contructors
to removes all incomprehensible variable names and make everything more uniform
to better understand where the data comes from,
the second thing is allow to get/propagate the properties easily with less variables

### DIFF
--- a/src/test/TestHelper.h
+++ b/src/test/TestHelper.h
@@ -13,6 +13,7 @@
 #include "../common/RepresentationChooser.h"
 #include "../parser/DASHTree.h"
 #include "../parser/HLSTree.h"
+#include "../utils/PropertiesUtils.h"
 
 std::string GetEnv(const std::string& var);
 void SetFileName(std::string& file, const std::string name);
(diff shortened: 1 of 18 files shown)


src/test/TestHelper.h
```diff
@@ -75,10 +76,13 @@ class AESDecrypter : public IAESDecrypter
...
-  uint64_t GetNowTime() override { return mock_time; }
-  std::chrono::system_clock::time_point GetNowTimeChrono() { return mock_time_chrono; };
+  DASHTestTree(UTILS::PROPERTIES::KodiProperties kodiProps);
+  uint64_t GetNowTime() override { return m_mock_time; }
+  void SetNowTime(uint64_t time) { m_mock_time = time; }
   void SetLastUpdated(std::chrono::system_clock::time_point tm) override { lastUpdated_ = tm; };
+  std::chrono::system_clock::time_point GetNowTimeChrono() { return m_mock_time_chrono; };
+
+private:
+  uint64_t m_mock_time = 10000000L;
```
glennguy: no big deal but should we be dropping the underscores in the name i.e `m_mockTime` ?

CastagnaIT: yes my intention but also my oversight :-D

glennguy: All good! I'm not too worried about getting everything perfect first time. We can always do a couple of cleanups later on to get things that were missed beforehand.

glennguy (approved):
Looks great!

Issue #544, opened 2020-10-29
With Facebook live streams, sound is lost after some seconds


tanty:
https://forum.kodi.tv/showthread.php?tid=311335

Still reproducible with latest Kodi (Leia)

glennguy:
@tanty is this issue still current? If so could you please post a full debug log? Don't turn on component logging btw.

tanty:
Yes, it's still happening.

Kodi v18.7 from Raspbian Buster repositories (RPi4). inputstream.adaptive addon v2.4.4
[debug-kodi.log](https://github.com/xbmc/inputstream.adaptive/files/6625127/debug-kodi.log)

tanty:
Ugh, just realized that there is v2.4.7 for Leia and Debian has this version for Buster but Raspbian only has v.2.4.4 as the latest version. Not sure how to do to push a newer package there ...

... all in all, I cannot say if the latest version has fixed this. I'll try to figure out.

dagwieers:
@tanty We have a list of contacts on the Wiki at: https://github.com/xbmc/inputstream.adaptive/wiki/Stakeholders#platforms

But the release process currently does not involve contacting them when a new release is ready. We should ask all maintainers to track ISA releases on GitHub. (I think I did ask this in a previous communication)

Raspbian does not seem to be listed. And the releases have not been updated on the Wiki matrix either.

**Update**: I updated the matrix with the most current version.
**Update2**: @tanty I assumed Raspbian was getting updates from Debian, but if that is not the case (or it needs to be acknowledged first) we ought to add a link for Raspbian as well. Feel free to update the Wiki!

dagwieers:
Ok, I add @basilgello as the maintainer for Raspbian now.

basilgello:
I do maintain Debian, not specifically Raspbian, sorry!

tanty:
@peak3d appears as maintainer in the v2.4.4 package I've installed:

```
pi@raspberrypi:~ $ apt show kodi-inputstream-adaptive
Package: kodi-inputstream-adaptive
Version: 2.4.4-1~buster
Priority: extra
Section: libs
Maintainer: peak3d <info@peak3d.de>
Installed-Size: 2,287 kB
Depends: libc6 (>= 2.17), libexpat1 (>= 2.0.1), libgcc1 (>= 1:3.5), libstdc++6 (>= 6)
Conflicts: kodi-inputstream-mpd, kodi-inputstream-smoothstream
[...]

PR #913, opened 2022-03-08
[cleanup] Implemented UrlUtils and relative code cleanup

CastagnaIT:
Moved all common URL operations in utils
and to make URLS operations more safe

This PR also add the support to handle Relative URLS with parent levels, so example:
`../../this/sun`

when will be joined with another base URL will be automatically adjusted, e.g.
base URL: `http://www.example.com/cat/dog/birds/`

After the join operation will be:
`http://www.example.com/cat/this/sun`

before this PR instead result as:
`http://www.example.com/cat/dog/birds/../../this/sun`
cause of wrong url stream download with Facebook live videos ref. #544

For the url validation `URL::IsValidUrl` i have implemented partially the isURL method from https://github.com/validatorjs/validator.js i think it is enough for our purposes, should be a task of the developer who integrates ISAdaptive into his add-on that should take care of providing valid links

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -10,6 +10,7 @@
 
 #include "../oscompat.h"
 #include "../utils/StringUtils.h"
+#include "../utils/UrlUtils.h"
 #include "../utils/Utils.h"
 #include "../utils/log.h"
 #include "PRProtectionParser.h"
@@ -79,7 +80,6 @@ static uint8_t GetChannels(const char** attr)
 
 static uint64_t ParseSegmentTemplate(const char** attr,
                                          std::string baseURL,
-                                         std::string baseDomain,
                                          DASHTree::SegmentTemplate& tpl,
                                          uint64_t startNumber)
 {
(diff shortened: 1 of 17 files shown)


CastagnaIT:
Please before merge CI migration, after i will rebase this PR to make CI working

src/parser/DASHTree.cpp
```diff
@@ -1635,40 +1630,44 @@ void DASHTree::RefreshLiveSegments()
...
-               br != er; ++br)
+      for (auto period : periods_)
+      {
+        if (!period)
+          continue;
+        for (auto adaptSet : period->adaptationSets_)
+        {
+          if (!adaptSet)
+            continue;
+          for (auto repr : adaptSet->representations_)
```
glennguy, line 1649: So much better, thankyou!

src/parser/HLSTree.cpp
```diff
@@ -102,8 +103,8 @@ int HLSTree::processEncryption(std::string baseUrl, std::map<std::string, std::s
   if (map["METHOD"] == "AES-128" && !map["URI"].empty())
   {
     current_pssh_ = map["URI"];
-    if (current_pssh_[0] != '/' && current_pssh_.find("://") == std::string::npos)
-      current_pssh_ = baseUrl + current_pssh_;
+    if (!URL::IsValidUrl(current_pssh_))
+      current_pssh_ = URL::Join(baseUrl, current_pssh_);
```
glennguy, line 107: From this we should be able to remove https://github.com/xbmc/inputstream.adaptive/blob/ae9f92f40ef52df8b6ee12ef7a808a8816b61d1b/src/parser/HLSTree.cpp#L807

glennguy (commented):
Great work, thanks!

src/parser/HLSTree.cpp
CastagnaIT: not sure to understand exaclty
you mean that in this specific case is not needed call `BuildDownloadUrl`
or you mean delete `BuildDownloadUrl` method?
if this last case, delete this helper method dont seems good, since it is called from different points

CastagnaIT: now that i see for the line 106 `if (!URL::IsValidUrl(current_pssh_))`
the url should come from the manifest need really to check validation?
do you think we can replace it as:
 `if (!URL::IsUrlAbsolute(current_pssh_))`

CastagnaIT: comment removed -- never mind my overisght

glennguy:
Thanks @CastagnaIT 

Wrt HLSTree and the issue with Dailymotion (and separate from this PR, but talking about here because very much related to all of this) we should look at once this goes in. I have solved but it probably needs looking at properly in conjunction with these changes.

The main issue being that BuildDownloadUrl only cares about the global (master manifest) base url/domain

We can't rely on effective_url_ IMO because it's global to the tree and I fear that it won't work in the case of different reps that have different domains to eachother, best bet is to keep these values on the reps themselves.

I tried by making this function and PreparePaths take a reference to a base/domain, and solving the HLS pssh download by also passing the rep to OnDataArrived so we can access it.

Another way would be to always resolve full url for HLS. Downside is it may be bad performance on parsing playlists with thousands of segments (rather than at download time). 4 hour playlist of 4 second segments == 3600. Having said that, we already do it half the time anyway. Then we can drop these handful of conditionals in HLSTree and just do a BuildDownloadUrl (that respects the rep's base url) instead. Thoughts?

src/parser/HLSTree.cpp
glennguy: yes, just that specific case, with the code above you'll always get a full url right?

CastagnaIT: yes you have right i will remove it

CastagnaIT:
seems reasonable but i can't judge completely
I should have at least an example of reps with different domains to understand better
I am making small improvements and this helping me to understand better how works but not yet under all aspects

glennguy (approved):

PR #914, opened 2022-03-09
[CI] Migrated to windows 2019 and VS 2019

CastagnaIT:
Migration to windows 2019 and VS 2019
Windows-2016 environment will be removed and not works anymore


CastagnaIT:
i have bumped SDK to `10.0.22000.0` if it is a bad idea i can rollback to the old version (seems still available)

glennguy (approved):

PR #916, opened 2022-03-14
[cleanup] Cleanup to segment url and SPINCACHE

CastagnaIT:
This cleanup a bit the `Segment` struct and the related code
such as solve the big mess of casts with different variables types where should be size_t and nothing else,
and remove all non sense code related to `Segment->url` was handled as char array instead of a simple std::string

SPINCACHE is a bit changed, before was used the overloading of `operator[]` with a pointer as output,
where instead the source of data are not pointers, this implementation is such confusing, usually expect type-consistent output, e.g. like happen with vector's index access, but here we have a custom behaviour with different output type
so better handle it with an appropriate method with doc

another thing is that all code which gave access to SPINCACHE with `operator[]` was totally unsafe code with the risk of dereferencing a nullptr, now fixed

Last thing, this was not my initial intention, but i have reworked in full `DASHTree::RefreshLiveSegments()` in DASHTree.cpp
Initially, i thought it was a problem with a constant variable because i had "url" as string like readonly,
so i had started to rework it to understand the cause and then simplify the code,
found the cause i was halfway through the work so i finished it... i have triple checked the code i hope no errors

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -60,9 +60,6 @@ AdaptiveStream::~AdaptiveStream()
 {
   stop();
   clear();
-
-  for (SEGMENTBUFFER& buf : segment_buffers_)
-    delete[] buf.segment.url;
 }
 
 void AdaptiveStream::Reset()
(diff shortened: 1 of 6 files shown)


src/common/AdaptiveStream.cpp
```diff
@@ -268,28 +265,33 @@ bool AdaptiveStream::start_stream()
...
-          if (!pos)
+          if (pos == 0)
             pos = 1;
         }
         uint64_t duration(current_rep_->get_segment(pos)->startPTS_ -
                           current_rep_->get_segment(pos - 1)->startPTS_);
-        pos -= static_cast<uint32_t>((tree_.live_delay_ * current_rep_->timescale_) / duration);
-        current_rep_->current_segment_ = current_rep_->get_segment(pos < 0 ? 0 : pos);
+        size_t segmentPos{0};
+        if (pos > (tree_.live_delay_ * current_rep_->timescale_) / duration) {
```
glennguy: new line for curly brace

src/common/AdaptiveStream.cpp
```diff
@@ -268,28 +265,33 @@ bool AdaptiveStream::start_stream()
...
+        current_rep_->current_segment_ = current_rep_->get_segment(segmentPos);
       }
       else // switching streams, align new stream segment no.
       {
-        std::int64_t segmentId = segment_buffers_[0].segment_number;
+        uint64_t segmentId = segment_buffers_[0].segment_number;
         if (segmentId >= current_rep_->startNumber_ + current_rep_->segments_.size())
           segmentId = current_rep_->startNumber_ + current_rep_->segments_.size() - 1;
-        current_rep_->current_segment_ = current_rep_->get_segment(static_cast<uint32_t>(segmentId - current_rep_->startNumber_));
+        current_rep_->current_segment_ = current_rep_->get_segment(static_cast<size_t>(segmentId - current_rep_->startNumber_));
```
glennguy: line length

src/common/AdaptiveStream.cpp
```diff
@@ -529,20 +538,28 @@ bool AdaptiveStream::ensureSegment()
 
     if (nextSegment)
     {
+      if (!current_rep_->segments_.Get(0))
```
glennguy: Is there a possibility that we will have first segment == null while having at least some segments?

src/common/AdaptiveTree.h
```diff
@@ -214,44 +228,41 @@ class ATTR_DLL_LOCAL AdaptiveTree
     std::chrono::time_point<std::chrono::system_clock> repLastUpdated_;
     const Segment *current_segment_;
     const Segment *get_initialization()const { return (flags_ & INITIALIZATION) ? &initialization_ : 0; };
-    const Segment *get_next_segment(const Segment *seg)const
+    const Segment* get_next_segment(const Segment* seg)
```
glennguy: Please say if I have this wrong - is there a reason to remove const qualifier for these functions? My understanding is const functions guarantee that member variables won't be changed which looks to be the case here.

glennguy (changes requested):

src/common/AdaptiveTree.h
CastagnaIT: no i have forgot to readd it

CastagnaIT: however the left const make the vars/members constants
`**const** Segment *get_next_segment(const Segment *seg) const`

instead the right const make the function itself const
`const Segment *get_next_segment(const Segment *seg) **const**`
that will use const functions

the "const" is always applied to the thing on the left of where this word is placed
if of the left of "const" word there is nothing then the "const" will be applied to the right,
example:
`const Segment *get_next_segment(const Segment *seg)`
on the left of const there is nothing then will be applied to the right, so to the `Segment` contained in the pointer

src/common/AdaptiveStream.cpp
CastagnaIT: right the various thing called have confused me, removed

src/common/AdaptiveTree.h
```diff
@@ -29,32 +29,54 @@ namespace adaptive
 template<typename T>
 struct ATTR_DLL_LOCAL SPINCACHE
 {
-  SPINCACHE() :basePos(0) {};
-
-  size_t basePos;
+  /*! \brief Get the <T> value pointer from the specified position
+   *  \param pos The position of <T>
+   *  \return <T> value pointer, otherwise nullptr if not found
+   */
```
CastagnaIT, line 35: @glennguy do you think here is appropriate this description? need to be improved in some way?

glennguy: I think it's fine, easy to read

glennguy (approved):
Great effort - thanks!

Especially the live updates in DASHTree - it's readable now! :)

PR #924, opened 2022-03-29
[cleanup] Refactor of representation chooser

CastagnaIT:
This refactor aims to move code through a more clean reusable interface,
the idea (in the future) is have the ability to provide more types of representation chooser's
that could be for example:

- Default (the current WIP adaptive)
- Agressive
- Fixed (no adaptive)
- Test
- etc ...

Then an user in future will can set the preferred "chooser" from the addon settings, and can customize it as you wish. (the addon settings will need to be reworked)

OFC this is a first step, ~i have cleaned a lot of code but not in full~
The functionality of the code has remained unchanged, only the prints in the log are a little more accurate.

There are parts of the "chooser" code that i have not yet been able to understand and also will need to be improved in the future (e.g. `ChooseNextRepresentation`/`ChooseRepresentation` have also duplicate code, magic numbers, and so on...)

Just a note, i have changed the type of variables referred to resolution height/width from `unisigned int` to `int`, the code was really messed up of type casts, which made no sense at all since every point in the code was handling number operations as integer

I think this is a good way


src/main.cpp
```diff
@@ -1617,43 +1619,16 @@ Session::Session(const PROPERTIES::KodiProperties& kodiProps,
...
-
-  manual_streams_ = kodi::addon::GetSettingInt("STREAMSELECTION");
-  LOG::Log(LOGDEBUG, "STREAMSELECTION selected: %d ", manual_streams_);
-
-  allow_no_secure_decoder_ = kodi::addon::GetSettingBoolean("NOSECUREDECODER");
-  LOG::Log(LOGDEBUG, "FORCENONSECUREDECODER selected: %d ", allow_no_secure_decoder_);
+  m_settingNoSecureDecoder = kodi::addon::GetSettingBoolean("NOSECUREDECODER");
+  LOG::Log(LOGDEBUG, "Setting NOSECUREDECODER value: %d", m_settingNoSecureDecoder);
 
   int buf = kodi::addon::GetSettingInt("MEDIATYPE");
```
glennguy: maybe a better name here for buf, like mediaType - or does it work to do `switch (kodi::addon::GetSettingInt("MEDIATYPE"))`

src/main.cpp
```diff
@@ -2295,82 +2252,124 @@ bool Session::InitializePeriod(bool isSessionOpened /* = false */)
...
 
-    do
+    if (isManualStreamSelection)
     {
-      m_streams.push_back(std::make_unique<STREAM>(
-          *adaptiveTree_, adp, media_headers_, representationChooser_,
-          m_kodiProps.m_playTimeshiftBuffer, repId, first_period_initialized_));
-      STREAM& stream(*m_streams.back());
+      // Add all stream representations
+      for (size_t reprIdx{0}; reprIdx < adp->representations_.size(); reprIdx++)
```
glennguy: this `reprIdx` can be a little confusing to read next to `reprIndex`, maybe just sticking with the usual `i` would make for cleaner reading and less chance to accidentally use the wrong one?

glennguy:
This looks fantastic btw

src/main.cpp
CastagnaIT: done

src/main.cpp
CastagnaIT: restored as `i`

glennguy (approved):

PR #925, opened 2022-03-30
Add add-on icon/fanart/info

CastagnaIT:
Add icon / fanart / info have never been added,
although of little importance it gives a better appearance to the component

Icon and fanart images come from rawpixel.com and are:
royalty-free, without license, no need to add copyright author reference, for personal and commercial use
so in line with add-ons rules

SShot01
SShot02


glennguy (approved):
Thanks @CastagnaIT

PR #927, opened 2022-04-01
Print a log warning for unsupported container/codec

CastagnaIT:
While testing unsupported codecs that cause a stream to be discarded
i have noticed that it helps to have a warning of the reason


glennguy (approved):
👍

PR #929, opened 2022-04-03
[cleanup] Removed KodiAdaptiveStream class

CastagnaIT:
As discussed in the past i have removed not needed `KodiAdaptiveStream` class
and merged the code to the parent class, no functional change

i do not like too much add `INPUTSTREAM_TEST_BUILD` conditions
but i don't know of any better way to avoid breaking the test build
due to references to objects of Kodi interface includes

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -8,16 +8,23 @@
 
 #include "AdaptiveStream.h"
 
-#include "RepresentationChooser.h"
-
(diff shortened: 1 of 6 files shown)


glennguy (approved):

glennguy:
Happy to put this in now and fix up later once the refactoring is largely finished. Hopefully we can come up with a design that is less tightly coupled and makes separating the stuff with kodi #includes a lot easier.

PR #932, opened 2022-04-13
Representation chooser improvements and new settings

### LINKED ISSUE #528: Maximum resolution - please provide more options.
This menu cannot currently set any maximum resolution limit between 1280p and 4K. But unfortunately hardware limits of my raspi4 seem to be exactly somewhere in between, so i have to choose the resolution manualy every time.

Can you please add following max resolution choices?

1440p, 1620p, 1920p, 2160p, 2540p, 2K (and possibly other sensible values)

image

CastagnaIT:
Another step to clean the representation chooser interface to prepare it for future new choosers

~This is full WIP is well on track but there are still a number of cases that i have not tested yet~

Main changes:

### Settings
has been reworked the add-on settings with the idea to be able to select the type of representation chooser, and change the available settings according to the type chosen.

Since there are a lot of changes in settings, i have added a first new representation chooser **Manual GUI**, to restore the manual stream selection mode, in separately way, which would otherwise conflict with the automatic selection of streams.

preview:
immagine
immagine

### Initial bandwidth change
Has been removed the bandwith value saved/restored to disk, that value not reflect the actual network condition.
We are at a disadvantage because with the web players, it is the browser that provides this information. Therefore, the only way to determine it is when we downloading the manifest files. 

As manifest files very small file, it is downloaded too quickly and the bandwidth is not determined properly, so I have proportioned the value, not so correct but i don't think you can do better (other ideas welcome).
Since is not perfect way, as you can see in the second screenshot, there is a settings that allow to disable the init. bandwidth autodetermination.

### Add new ISA Kodi properties "inputstream.adaptive.stream_selection_type"
This allow an addon to force choose a specific "Chooser type" so e.g. to have a fixed stream quality instead an adaptive stream

### Related cleanups
some bigger cleanup are:
- `AdaptiveTree::download` full reworked which was well messed up
- The assured_buffer_length_/max_buffer_duration_ has been moved out of repr. chooser
- In settings has been changed from bps to kbps, much more user-friendly
- Add a proper way to clone a tree `AdaptiveTree::Clone` to prevents override problems with derived classes (currently used only for DASH case)
- Introduced `test/KodiStubs.h`, this file have functions/classes mokups to replace Kodi API includes for the test build, then this we allow to remove all the if/endif build conditions in code.
It works as follow:
```
#ifdef INPUTSTREAM_TEST_BUILD
#include "../test/KodiStubs.h"
#else
#include <kodi/AddonBase.h>
#endif
```

fixes #528, i agreed to add only a few resolutions that fall within the common standards

### DIFF
--- a/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
+++ b/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
@@ -20,125 +20,246 @@ msgctxt "Addon Description"
 msgid "InputStream client for adaptive streams"
 msgstr ""
 
+#. Category title
 msgctxt "#30100"
 msgid "General"
 msgstr ""
 
-# The minimum bandwidth which should not be deceeded.
+#. The minimum bandwidth which should not be deceeded
 msgctxt "#30101"
-msgid "Min. Bandwidth (Bit/s)"
+msgid "Minimum bandwidth (Kbps)"
 msgstr ""
 
-# The maximum bandwidth which should not be exceeded. 0=unlimited
+#. The maximum bandwidth which should not be exceeded
 msgctxt "#30102"
-msgid "Max. Bandwidth (Bit/s)"
+msgid "Maximum bandwidth (Kbps)"
 msgstr ""
 
-# Assured buffer length duration (seconds)
-msgctxt "#30200"
-msgid "Assured Buffer Duration (sec)"
-msgstr ""
-
-# Max buffer length duration (seconds)
-msgctxt "#30201"
-msgid "Max Buffer Duration (sec)"
-msgstr ""
-
-# Ignore Window Display Resolution Change
-msgctxt "#30202"
-msgid "Ignore Window Change"
-msgstr ""
-
-# Absolute path to the folder containing the decrypters
+#. Description of setting with label #30101, 30102
 msgctxt "#30103"
-msgid "Decrypter path"
+msgid "Establishes the bandwidth limit not to be exceeded with the audio / video streams. Set to 0 to disable it."
 msgstr ""
 
-# Maximum Resolution
+#empty strings from id 30104 to 30109
+
+#. Maximum video resolution non-protected videos
 msgctxt "#30110"
-msgid "Max. Resolution general decoder"
+msgid "Maximum resolution"
 msgstr ""
 
-msgctxt "#30111"
-msgid "Stream Selection"
-msgstr ""
+#empty string with id 30111
 
+#. Type of media streams that will be handled for the playback
 msgctxt "#30112"
-msgid "Media"
+msgid "Type of media streams played"
 msgstr ""
 
-# Maximum allowed resolution if decoded through secure path
+#. Maximum video resolution for DRM-protected videos
 msgctxt "#30113"
-msgid "Max. Resolution secure decoder"
+msgid "Maximum resolution for DRM videos"
 msgstr ""
 
-# Select streams without respecting HDCP status
-msgctxt "#30114"
-msgid "Override HDCP status"
-msgstr ""
+#empty string with id 30114
 
 # Do not respect display resolution when selecting streams
 msgctxt "#30115"
-msgid "Ignore Display Resolution"
+msgid "Ignore screen resolution"
 msgstr ""
 
+#. Description of setting with label #30115
+msgctxt "#30116"
+msgid "If enabled, the screen resolution (or window size when in windowed mode) will no longer be considered when selecting the best video stream resolution when video starts and while in playback."
+msgstr ""
+
+msgctxt "#30117"
+msgid "Manual stream selection mode"
+msgstr ""
+
+#. Description of setting with label #30117
+msgctxt "#30118"
+msgid "Defines which type of streams to make available for manual selection on Kodi OSD settings during playback."
+msgstr ""
+
+#empty string with id 30119
+
+#. Category title
 msgctxt "#30120"
 msgid "Expert"
 msgstr ""
 
-msgctxt "#30121"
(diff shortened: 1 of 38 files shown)


inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
```diff
@@ -20,125 +20,246 @@ msgctxt "Addon Description"
...
+msgstr ""
+
+#. Ignore screen resolution change e.g. window resize
+msgctxt "#30202"
+msgid "Ignore screen resolution change"
+msgstr ""
+
+#. Description of setting with label #30202
+msgctxt "#30203"
+msgid "If enabled, the screen resolution (or window size when in windowed mode) will no longer be considered to selecting the best video stream resolution while in playback."
```
glennguy: > will no longer be considered ~~to~~ **when** selecting the best video stream resolution while in playback

inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
```diff
@@ -20,125 +20,246 @@ msgctxt "Addon Description"
...
+msgid "If enabled, the bandwidth will be determined by the first download, it may not be accurate. If the video quality at the start of playback is too low try disabling it."
+msgstr ""
+
+msgctxt "#30170"
+msgid "Initial bandwidth (Kbps)"
+msgstr ""
+
+#. Description of setting with label #30170
+msgctxt "#30171"
+msgid "Defines the initial bandwidth when cannot be automatically determined. This value can be overridden by the minimum bandwidth setting."
```
glennguy: Defines the initial bandwidth when *it* cannot be automatically determined

inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
```diff
@@ -20,125 +20,246 @@ msgctxt "Addon Description"
...
+msgid "WARNING: This is a TEST feature, may not work appropriately and may change on future versions."
+msgstr ""
+
+msgctxt "#30168"
+msgid "Auto determines initial bandwidth"
+msgstr ""
+
+#. Description of setting with label #30168
+msgctxt "#30169"
+msgid "If enabled, the bandwidth will be determined by the first download, it may not be accurate. If the video quality at the start of playback is too low try disabling it."
```
glennguy: If enabled the bandwidth will be determined by the first download, *however* it may not be accurate. If the video quality at the start of playback is too low try disabling it.

inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
```diff
@@ -20,125 +20,246 @@ msgctxt "Addon Description"
...
 msgctxt "#30115"
-msgid "Ignore Display Resolution"
+msgid "Ignore screen resolution"
 msgstr ""
 
-msgctxt "#30120"
-msgid "Expert"
+#. Description of setting with label #30115
+msgctxt "#30116"
+msgid "If enabled, the screen resolution (or window size when in windowed mode) will no longer be considered to selecting the best video stream resolution when video starts and while in playback."
```
glennguy: .... will no longer be considered ~~to~~ *when* selecting the best video....

src/common/AdaptiveTree.cpp
```diff
@@ -410,4 +421,92 @@ namespace adaptive
...
+    while (!isEOF)
+    {
+      // Read the data in chunks
+      ssize_t byteRead{file.Read(bufferData.data(), bufferSize)};
+      if (byteRead == -1)
+      {
+        LOG::Log(LOGERROR, "An error occurred in the download: %s", url.c_str());
+        break;
+      }
+      else if (byteRead == 0) // EOF or undectetable error
```
glennguy: typo - undetectable

src/common/AdaptiveTree.cpp
```diff
@@ -410,4 +421,92 @@ namespace adaptive
...
+        data.write(bufferData.data(), byteRead);
+      }
+    }
+
+    if (isEOF)
+    {
+      long dataSizeBytes{static_cast<long>(data.tellp())};
+      if(dataSizeBytes > 0)
+      {
+        double downloadSpeed{file.GetFileDownloadSpeed()};
```
glennguy, line 487: just a question here more than anything else - are we getting a skewed download speed because of gzip/deflate compression? i.e if dataSizeBytes is the uncompressed size then of course transferring compressed text will be much smaller and faster... quick research suggests 3-4 times smaller size

glennguy (changes requested):

src/common/AdaptiveTree.cpp
CastagnaIT: oh you have right i will use Content-Length header to get the original body size

CastagnaIT: change added

src/common/AdaptiveTree.cpp
```diff
@@ -410,4 +421,99 @@ namespace adaptive
...
+
+    if (isEOF)
+    {
+      long dataSizeBytes{static_cast<long>(data.tellp())};
+      if(dataSizeBytes > 0)
+      {
+        // Get body lenght (could be gzip compressed)
+        std::string contentLenghtStr{
+            file.GetPropertyValue(ADDON_FILE_PROPERTY_RESPONSE_HEADER, "Content-Length")};
+        long contentLenght{std::atol(contentLenghtStr.c_str())};
```
glennguy: just a typo on this variable name

glennguy (approved):

glennguy:
Great work!

CastagnaIT:
force pushed only to fix a clang formatting

PR #939, opened 2022-04-18
Moved HDCP check temporarily to AdaptiveTree

CastagnaIT:
Check HDCP compliant is not a task of a representation chooser,
and this allows me to create new choosers without still keeping this in the interface

i have temporary moved this check in AdaptiveTree in awaiting for the future refactor of Session class, so a middle step

sidenote: `hdcpVersion` on android side was unset (not even in the structure) then i have add the value


glennguy (approved):
Looks good!

PR #941, opened 2022-04-19
Adaptive chooser improvements and cleanup

### LINKED ISSUE #596: [FR] Property max_resolution
Is it possible to add max_resolution property for listitem? It would be nice to be able to set max requested resolution and override it's settings. I guess max_bandwidth works just like that but it's less intuitive for normal user to set bandwidth instead of resolution.

Use case will be let users to set maximum resolution in my addon's settings and pass that value to imputstream.adaptive when starting stream.
`list_item.setProperty('inputstream.adaptive.max_resolution', '720p')`

CastagnaIT:
Another round of improvements and cleanup to make all more structured for new future chooser's properties

Has been add 2 new properties for Adaptive chooser:
- `inputstream.adaptive.chooser_resolution_max` -> overwrites the respective setting only if the user has not set it, or if it exceeds the value
- `inputstream.adaptive.chooser_resolution_secure_max` -> overwrites the respective setting only if the user has not set it, or if it exceeds the value

and deprecated the old one:
- `inputstream.adaptive.max_bandwidth` -> for now still working, but will raise a log warning to notice future its removal. Now replaced by `inputstream.adaptive.chooser_bandwidth_max`

Has been fixed in adaptive chooser the user bandwidth limit setting, that was respected in `ChooseRepresentation` method but not in `ChooseNextRepresentation` method

fix #596

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ set(ADP_SOURCES
 	src/utils/Base64Utils.cpp
 	src/utils/PropertiesUtils.cpp
 	src/utils/StringUtils.cpp
+	src/utils/SettingsUtils.cpp
 	src/utils/UrlUtils.cpp
 	src/utils/Utils.cpp
 	src/oscompat.cpp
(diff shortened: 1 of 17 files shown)


glennguy (approved):
Great ideas in here, thanks!

PR #942, opened 2022-04-20
New representation choosers: Fixed resolution / Ask quality

CastagnaIT:
This PR add two new choosers a lot requested over the time that will make happy a lot of users

### Fixed resolution
The stream is fixed for the full duration of the video, when playback starts will be chosen the best video resolution that fit the screen resolution, an user can also limit to a max resolution (e.g. for not powerful CPUs),
of course network bandwidth is not considered.

### Ask quality
Just before playing a video, the user will be asked to choose the preferred video quality from a dialog window,
which will remain fixed for the full duration of the video (network bandwidth is not considered).

immagine

here i have left a "todo" in sourcecode for cases of multi-codec manifests which may be discussed and improved in some way in the future

As always a video add-on can force the choice of one of the available choosers despite the user setting

### DIFF
--- a/src/common/RepresentationChooserAskQuality.cpp
+++ b/src/common/RepresentationChooserAskQuality.cpp
@@ -1,0 +1,111 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "RepresentationChooserAskQuality.h"
+
+#include "../utils/StringUtils.h"
+#include "../utils/Utils.h"
+#include "../utils/log.h"
+#include "RepresentationSelector.h"
+#include "kodi/tools/StringUtils.h"
+
+#ifndef INPUTSTREAM_TEST_BUILD
+#include <kodi/gui/dialogs/Select.h>
+#endif
+
+#include <vector>
+
+using namespace adaptive;
+using namespace CHOOSER;
+using namespace UTILS;
+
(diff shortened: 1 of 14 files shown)


src/common/RepresentationChooserAskQuality.cpp
```diff
@@ -0,0 +1,111 @@
...
+                                                  preselIndex, 10000)};
+
+    AdaptiveTree::Representation* selRep{bestRep};
+
+    if (selIndex >= 0) // If was <0 has been cancelled by the user
+      selRep = adp->representations_[selIndex];
+
+    m_selectedResWidth = selRep->width_;
+    m_selectedResHeight = selRep->height_;
+    m_isFirstVideoAdptationSetChosen = true;
```
glennguy: small typo here - `m_isFirstVideoAdaptationSetChosen`

src/common/RepresentationChooserFixedRes.h
```diff
@@ -0,0 +1,40 @@
...
+  virtual void PostInit() override;
+
+  adaptive::AdaptiveTree::Representation* ChooseRepresentation(
+      adaptive::AdaptiveTree::AdaptationSet* adp) override;
+
+  adaptive::AdaptiveTree::Representation* ChooseNextRepresentation(
+      adaptive::AdaptiveTree::AdaptationSet* adp,
+      adaptive::AdaptiveTree::Representation* currentRep) override;
+
+protected:
```
glennguy: does this need to be protected or would private be fine?

glennguy (changes requested):

glennguy (approved):

glennguy:
Awesome work! :)

PR #947, opened 2022-04-24
Fix webvtt/mp4 subtitles crash

### LINKED ISSUE #946: Subtitles causes video to crash
HI there. How come you guys do not allow Media stream to only have Audio+VIdeo? Because having it set to "all" causes video to crash in some addons such as Discovery+. I believe it has to do with the way that the website handles subtitles, because setting Media to Video or Audio only solves the issue. But obviously I want both video and audio.
Could you guys please push an update and fix this?
Thank you.

glennguy:
When we hit the situation in this while loop where `remainingVttcBoxSize` is 8, `uint32_t boxSize = ReadNextUnsignedInt(cbuf + pos);` ends up reading whatever junk is in memory. In my testing case it was some text left over from the previous mp4 segment giving a very large boxSize!

I slightly simplified by removing the redundant condition check and just making part of the loop.

@CastagnaIT does this look alright to you?

fixes #946 

Many thanks to @Dis90 for spending so much time and effort getting me a test stream

### DIFF
--- a/src/parser/WebVTT.cpp
+++ b/src/parser/WebVTT.cpp
@@ -53,11 +53,8 @@ bool WebVTT::Parse(uint64_t pts, uint32_t duration, const void *buffer, size_t b
       uint32_t remainingVttcBoxSize = ReadNextUnsignedInt(cbuf);
       int pos = 8;
 
-      while (remainingVttcBoxSize > 0)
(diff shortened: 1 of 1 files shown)


CastagnaIT (approved):
i think go well thanks

PR #951, opened 2022-04-26
Add new "Test" chooser representation

CastagnaIT:
This is the first implementation of the "Test" chooser

we can finally have the possibility of a controlled stream switching,
so we can test the stream adaptive feature to investigate for potential problems and to make improvements that could be needed also on Kodi core, without being tied to finding ways of limiting network bandwidth to be able to force stream quality switching

I thought of integrating this Chooser in an expandable way
the "Test mode" setting allow to choose the type of test to be executed while in playback
and if needed we can also expand the functionality with new test modes and/or new settings

immagine


glennguy (approved):
This is a great feature, thanks!

PR #953, opened 2022-04-29
[cleanup] Cleanup and moved md5 to utils

CastagnaIT:
As title, code funcitonality kept same as before

cleanups are:
add namespace
just fixed camel case in var/methods
removed `size_type` typedef not needed C++ provide `uint32_t` by default


glennguy (approved):
👍

PR #954, opened 2022-04-29
Fixed estimate segments count method

### LINKED ISSUE #152: Arithmetic exception causing Kodi to crash due to inputstream adaptive
```
Thread 97 "kodi.bin" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7f6441737700 (LWP 19312)]
0x00007f64f2f4aa23 in adaptive::AdaptiveTree::estimate_segcount (this=this@entry=0x7f64f8007430, duration=0, timescale=1000)
    at /build/kodi-inputstream-adaptive-DzFzop/kodi-inputstream-adaptive-2.0.20/src/common/AdaptiveTree.cpp:91
91	/build/kodi-inputstream-adaptive-DzFzop/kodi-inputstream-adaptive-2.0.20/src/common/AdaptiveTree.cpp: No such file or directory.
(gdb)
```

I installed the -dpg package for kodi-inputstream-adaptive, and used gdb to catch this problem that causes Kodi to crash to the X server for me. This is trying to load a DASH youtube livestream.

CastagnaIT:
This method is used only to optimize initial memory allocation of the vector not so important if we dont have an accurate value

but the calculation to get the lenght in seconds was done between two uint variables
then if a < b the result will be 0
and the following division will cause an exception

Fix #152

i will do the backport after approvation

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -154,12 +154,12 @@ namespace adaptive
     return false;
   }
 
-  uint32_t AdaptiveTree::estimate_segcount(uint64_t duration, uint32_t timescale)
+  size_t AdaptiveTree::EstimateSegmentsCount(uint64_t duration, uint32_t timescale)
   {
-    LOG::Log(LOGDEBUG,"estimate_segcount  duration=%llu , timescale=%u",duration , timescale);
-
-    duration /= timescale;
-    return static_cast<uint32_t>((overallSeconds_ / duration)*1.01);
+    double lengthSecs{static_cast<double>(duration) / timescale};
+    if (lengthSecs < 1)
+      lengthSecs = 1;
+    return static_cast<size_t>(overallSeconds_ / lengthSecs);
   }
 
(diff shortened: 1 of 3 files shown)


src/common/AdaptiveTree.cpp
```diff
@@ -154,12 +154,12 @@ namespace adaptive
...
+  size_t AdaptiveTree::EstimateSegmentsCount(uint64_t duration, uint32_t timescale)
   {
-    LOG::Log(LOGDEBUG,"estimate_segcount  duration=%llu , timescale=%u",duration , timescale);
-
-    duration /= timescale;
-    return static_cast<uint32_t>((overallSeconds_ / duration)*1.01);
+    double lenghtSecs{static_cast<double>(duration) / timescale};
+    if (lenghtSecs < 1)
+      lenghtSecs = 1;
+    return static_cast<size_t>(overallSeconds_ / lenghtSecs);
```
glennguy: Just a typo in variable name - `lengthSecs`

glennguy (changes requested):

glennguy (approved):

PR #955, opened 2022-04-30
[chooser][askquality] Show fps value on stream info

CastagnaIT:
When available show fps value on stream info

immagine


glennguy (approved):
Nice, thanks!

PR #956, opened 2022-04-30
add missing assignment of effective url

glennguy:


glennguy:
Fixes paths for sub m3u8 files that are not in the same folder as master

CastagnaIT (approved):

PR #958, opened 2022-04-30
Fix crashing on 2nd period start

glennguy:
In start stream we're sending a null rep over to ChooseNextRepresentation which becomes defereneced

Calling ChooserRepresentation instead will do the usual, but now we already have some decent data for the current bandwidth (compared to very weak data on playback start which is just from downloading the manifest).
All things being equal with the adaptive chooser we should expect to select the same representation as the previous period if the adaptationsets/reps are 1:1 with the new period, otherwise the rep which best suits the current conditions.

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -451,8 +451,7 @@ bool AdaptiveStream::start_stream()
   if (choose_rep_)
   {
     choose_rep_ = false;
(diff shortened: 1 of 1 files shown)


CastagnaIT (approved):
now that i have take a better look with last cleanups done i think i can merge method
`ChooseRepresentation`
with
`ChooseNextRepresentation`

so your change for now it's fine
i will try merge these both methods next days

CastagnaIT (commented):
superseed by #959

PR #959, opened 2022-04-30
[chooser] Merged ChooseRepresentation with ChooseNextRepresentation

CastagnaIT:
Merged ChooseRepresentation with ChooseNextRepresentation
both methods do more or less the same things with small differences
implicitly this contains the fix mentioned in #958 then superseed PR #958
i preferred doing another cleaning step

Log has been improved, now you can understand when chooser select the representation for the first start or for period change by:
`[Repr. chooser] Selected representation`
vs
`[Repr. chooser] Changed representation`
that happens when stream is changed for a chooser specified behaviour

you can see by testing multiperiod mpd that when a period is switched, there are two sequentials `Selected representation` that sound a bit weird, that could be need to investigate in future

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -451,8 +451,7 @@ bool AdaptiveStream::start_stream()
   if (choose_rep_)
   {
     choose_rep_ = false;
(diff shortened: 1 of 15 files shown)


glennguy (approved):
Thanks, looks great

PR #962, opened 2022-05-04
[cleanup] Removed CheckHDCP from RefreshLiveSegments

CastagnaIT:
this part of code confuses me, this check is not necessary as we look for updates based on existing representations (already HDCP checked)

add two non-thread safe todo's, to keep in mind for future reworking

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1653,6 +1653,7 @@ void DASHTree::RefreshSegments(Period* period,
 }
 
 //Can be called form update-thread!
+//! @todo: we are updating variables in non-thread safe way
 void DASHTree::RefreshLiveSegments()
 {
   if (has_timeshift_buffer_ && !update_parameter_.empty())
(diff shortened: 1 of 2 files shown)


glennguy:
Agreed, my reading of DASH specs says that rep details should not change with updates (emphasis mine):

>4.4.3.3. Permitted Updates in an MPD16
According to clause 5.4 of ISO/IEC 23009-1, when the MPD is updated
• the value of MPD@id, if present, shall be the same in the original and the updated MPD;
• the values of any Period@id attributes shall be the same in the original and the updated
MPD, unless the containing Period element has been removed;
• the values of any AdaptationSet@id attributes shall be the same in the original and
the updated MPD unless the containing Period element has been removed;
• any Representation with the same id and within the same Period as a Representation
appearing in the previous MPD ***shall provide functionally equivalent attributes and ele-
ments***, and shall provide functionally identical Segments with the same indices in the cor-
responding Representation in the new MPD.

glennguy (approved):

CastagnaIT:
thanks for confirmation

PR #971, opened 2022-05-10
[bento4] Fix VP9 GetData for metadata

CastagnaIT:
Fix regression VP9 metadata was not stored
then `AP4_VpccAtom` in `VP9CodecHandler` always return no data

from https://github.com/xbmc/inputstream.adaptive/commit/e016f6b68f81973ce301fcb194df4e4a21143c5e

i made a separate patch file, for future same changes to other atoms


glennguy (approved):
Thanks for finding and fixing!

Issue #525, opened 2020-10-06
Short netfx VP9 movie shown in wrong colors and lines


CastagnaIT:
~I found a movie that can not be played with VP9 stream~
UPDATE: (now starts) There is a VP9 movie shown with wrong colors and lines
~at least on Windows the Kodi player stuck in preloading state (no picture at all),~

This happen on Windows and also on CoreELEC
i do not know where can be the problem because in the log i not found nothing, some suggestion?

Title: Cosmos laundromat
ID: 80114804
Length: 12 minutes

![Clipboard02](https://user-images.githubusercontent.com/3257156/118498643-8b619300-b726-11eb-9d6a-80aa4654d913.jpg)

Logs:
[kodi_ffmpegdbg.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6494157/kodi_ffmpegdbg.log.txt)
[kodi_std.log.txt](https://github.com/xbmc/inputstream.adaptive/files/6494158/kodi_std.log.txt)
[manifest.json.txt](https://github.com/xbmc/inputstream.adaptive/files/6494159/manifest.json.txt)

CastagnaIT:
for now i close this issue,
the video specified use vp9 profile 2 that it in theory could contains HDR data it could be something mishandled in kodi core
for now net***x seems to have put in standby the vp9 profile 2
if it is reintroduced in the future and the problem recurs, I will see what to do.

PR #972, opened 2022-05-10
Implement AV1 codec support

CastagnaIT:
This implement AV1 codec support

What is the current status, has been implemented for DASH and unofficially for HSL
the unencrypted streams are played with success, some examples:

https://storage.googleapis.com/bitmovin-demos/av1/stream.mpd
https://storage.googleapis.com/bitmovin-demos/av1/stream_chrome.mpd
http://ftp.itec.aau.at/datasets/mmsys18/testing/Beauty_4sec/multi-codec.mpd

For encrypted AV1 streams i havent found sample streams to test,
the only way that i found to test it is by using n@tflix, that provide MP4/CBCS.
Unfortunately, these encrypted streams are not being played in the right way,
~a possible cause is the use of CMAF~ (EDIT: the problem is wrong video format set), the image result as:
immagine
this problem looks similar to my old test with VP9.2 in Issue https://github.com/xbmc/inputstream.adaptive/issues/525 (still unresolved, not more tested over the time not knowing how to solve)

To do tests with N@tflix is possible to use this PR branch:
https://github.com/CastagnaIT/plugin.video.netflix/pull/1343
(clearly on browser Chrome N@tflix AV1 playback works fine)

I can suggest to add this support to allow to play at least AV1 unencrypted streams,
and when we will find an encrypted sample in future we will see how goes

**This depends from:** https://github.com/xbmc/xbmc/pull/21390

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,8 +9,9 @@ find_package(Kodi REQUIRED)
 set(ADP_SOURCES
     src/AdaptiveByteStream.cpp
 	src/main.cpp
-	src/codechandler/AVCCodecHandler.cpp
 	src/codechandler/CodecHandler.cpp
+	src/codechandler/AV1CodecHandler.cpp
+	src/codechandler/AVCCodecHandler.cpp
 	src/codechandler/HEVCCodecHandler.cpp
 	src/codechandler/MPEGCodecHandler.cpp
 	src/codechandler/TTMLCodecHandler.cpp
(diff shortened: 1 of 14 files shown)


glennguy:
Is there any info here that could help - https://aomediacodec.github.io/av1-isobmff/

~~One thing I've noticed is that although we seem to need to user codec private data, this info looks like it would stay empty in the sample streams. Maybe we need to extract it from the init segment if it's not in the manifest?~~ nevermind, re-read the OP and can see it works for unencrypted sample streams.

CastagnaIT:
already tried read in link but i have found nothing that help me
i am thinking of something not implemented in wvdecrypter video decode but i dont know what
(this codec do not have things like sps or pps as hevc is much simplier)
another think, that i dont know if it is relevant is that n@tflix have AV1 with CMAF
we should test an encrypted AV1 without CMAF and see how works

AlwinEsch:
@CastagnaIT please make a rebase and to combine with https://github.com/xbmc/inputstream.adaptive/pull/989 to have ready about new release.

**NOTE:** We need to wait until launchpad have newest Kodi Version before we can merge requests on add-ons!

AlwinEsch:
@glennguy has brought in the other, seen are a bit different, @CastagnaIT to become Jenkins happy you need to make again, Sorry!

CastagnaIT:
yes now i will see to fix it

glennguy (approved):
Thanks @CastagnaIT 
The wvdecrypter changes in an upcoming PR should address the CBCS/multi-subsample issues

Issue #963, opened 2022-05-04
kodi crash with some streams with inputstream.adaptive


TermeHansen:
### Describe the bug
When starting playback of some specific streams via the plugin.video.drnu addon, kodi will crash when using the adaptive inputstream. When using the normal ffmpeg demuxer it will play but will crash when trying to skip.

### Expected Behavior
Normal playback of hls stream

### Actual Behavior
kodi crashes with no errors in debug log

### To Reproduce
Play a .strm file with:

```
#KODIPROP:inputstream=inputstream.adaptive
#KODIPROP:inputstream.adaptive.manifest_type=hls
https://drod22d.akamaized.net/dk/encrypted/none/b4/625ef6edaa5a612268e778b4/00022162620/stream_fmp4/master_manifest.m3u8

``` 
[master_manifest.m3u8](https://pastebin.com/m0vT184b)

### Debuglog

The debug log can be found here:
https://pastebin.com/qPSALQ9v

crashlog
[https://pastebin.com/fk89XGBc](https://pastebin.com/fk89XGBc)

### Additional context
tried with ffmpegdirect. It starts but kodi crash when trying to skip...
```
#KODIPROP:inputstream=inputstream.ffmpegdirect
#KODIPROP:mimetype=application/x-mpegURL
#KODIPROP:inputstream.ffmpegdirect.manifest_type=hls
https://drod22d.akamaized.net/dk/encrypted/none/b4/625ef6edaa5a612268e778b4/00022162620/stream_fmp4/master_manifest.m3u8

```

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [X] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Ubuntu 20.04
 - Kodi version: 19.4.0

CastagnaIT:
i need two more things:
attach here the https://drod22d.akamaized.net/dk/encrypted/none/b4/625ef6edaa5a612268e778b4/00022162620/stream_fmp4/master_manifest.m3u8 file

and provide stacktrace error, you can get it by running this in your ubuntu terminal:
- open terminal
- `gdb /usr/lib/x86_64-linux-gnu/kodi/kodi.bin`
- a gdb shell should be opened with a text like `Reading symbols from...`
- now open kodi
- make it crash
- Return on the shell in terminal and type: `bt full`

should display the error stack trace list

[...]

PR #986, opened 2022-05-16
Set byteranged init segments explicitly

glennguy:
Current behavior has been that while byteranged segments are parsed and set as the playlist is parsed, the init segment was not set until all lines were parsed. The byterange part of the init segment line is ignored and instead the range inferred by the start range of the first segment, and url being the same as the segments too.

This change now sets the init segment byterange and url at the time the line is read.

Allows for the very strange playlist scenario seen in #963 to playback correctly (other players support this too)


CastagnaIT (approved):
I do not have enough understanding of the logical set of this parser
i approve but untested

glennguy:
I tested with a few streams without any issues - byteranged, regular segments, and the weird case from the issue.

PR #987, opened 2022-05-16
Set byteranged init segments explicitly

TermeHansen:
this fixes issue https://github.com/xbmc/inputstream.adaptive/issues/963

tested!


glennguy:
@TermeHansen could you please amend the commit name and PR title to the same or similar to the one in #986?
Once done I can merge it in and do a release

PR #989, opened 2022-05-18
Handled limited GUI resolution with Adjust refresh rate and window resizing

CastagnaIT:
This handle two use cases
- when user limit Kodi GUI to 1080p to allow tv upscaling and use Adjust refresh rate to play until to 4k
- when Kodi is windowed mode we can have a callback from kodi of the updated resolution values after window resize

**It depends from:** https://github.com/xbmc/xbmc/pull/21319

Follow addon workaround settings are now not more needed and should not be used:
`Ignore screen resolution change`
`Ignore screen resolution`

they has been moved to `Expert` settings for a more likely future removal
also the id's of these settings are changed to make effective the new change on existing old addon installations

PS. i will do a cleanup to change all variables from `int` to `unsigned int` on another PR

### DIFF
--- a/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
+++ b/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
@@ -306,3 +306,8 @@ msgstr ""
 msgctxt "#30235"
 msgid "Segments"
 msgstr ""
+
+#. Category group title
+msgctxt "#30236"
+msgid "Override settings"
+msgstr ""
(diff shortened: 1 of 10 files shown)


CastagnaIT:
@glennguy you can prepare a review?
in theory we are close to the merge of PR's on xbmc side

glennguy (approved):
Looks fine to me, great feature to have.

AlwinEsch:
@CastagnaIT please make an rebase and an version increase to have usable about the API changes.

PR #997, opened 2022-05-31
Refactor of Session class

glennguy:
@CastagnaIT So far have done:

* Move KodiHost out of main.cpp
* Move Session out of main.cpp
* Move STREAM out of Session
* Changed some raw pointers to managed in Session. For this HLSTree now also creates its own AES decrypter instead of being injected
* Comments/docstrings for all methods in Session

I don't want to get too much more changes made, would rather limit the scope and have this reviewed/merged first. 

At the moment could you have a quick look and let me know if you think there's anything majorly incorrect about the work so far. Thanks.

### DIFF
--- a/src/KodiHost.cpp
+++ b/src/KodiHost.cpp
@@ -1,0 +1,119 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "KodiHost.h"
+
+#include "utils/log.h"
+
+void* CKodiHost::CURLCreate(const char* strURL)
+{
+  kodi::vfs::CFile* file{new kodi::vfs::CFile};
+  if (!file->CURLCreate(strURL))
+  {
+    delete file;
+    return nullptr;
+  }
+  return file;
+}
+
+bool CKodiHost::CURLAddOption(void* file, CURLOPTIONS opt, const char* name, const char* value)
+{
+  const CURLOptiontype xbmcmap[]{ADDON_CURL_OPTION_PROTOCOL, ADDON_CURL_OPTION_HEADER};
+  return static_cast<kodi::vfs::CFile*>(file)->CURLAddOption(xbmcmap[opt], name, value);
+}
+
+const char* CKodiHost::CURLGetProperty(void* file, CURLPROPERTY prop, const char* name)
+{
+  const FilePropertyTypes xbmcmap[]{ADDON_FILE_PROPERTY_RESPONSE_HEADER};
+  m_strPropertyValue = static_cast<kodi::vfs::CFile*>(file)->GetPropertyValue(xbmcmap[prop], name);
+  return m_strPropertyValue.c_str();
+}
+
+bool CKodiHost::CURLOpen(void* file)
+{
+  return static_cast<kodi::vfs::CFile*>(file)->CURLOpen(ADDON_READ_NO_CACHE);
+}
+
+size_t CKodiHost::ReadFile(void* file, void* lpBuf, size_t uiBufSize)
+{
+  return static_cast<kodi::vfs::CFile*>(file)->Read(lpBuf, uiBufSize);
+}
+
+void CKodiHost::CloseFile(void* file)
+{
+  return static_cast<kodi::vfs::CFile*>(file)->Close();
+}
+
+bool CKodiHost::CreateDir(const char* dir)
+{
+  return kodi::vfs::CreateDirectory(dir);
+}
+
+void CKodiHost::LogVA(const SSD::SSDLogLevel level, const char* format, va_list args)
+{
+  std::vector<char> data;
+  data.resize(256);
+
+  va_list argsStart;
+  va_copy(argsStart, args);
+
+  int ret;
+  while (static_cast<size_t>(ret = vsnprintf(data.data(), data.size(), format, args)) > data.size())
+  {
+    data.resize(data.size() * 2);
+    args = argsStart;
+  }
+  LOG::Log(static_cast<LogLevel>(level), data.data());
+  va_end(argsStart);
+}
+
+void CKodiHost::SetLibraryPath(const char* libraryPath)
+{
+  m_strLibraryPath = libraryPath;
+
+  const char* pathSep{libraryPath[0] && libraryPath[1] == ':' && isalpha(libraryPath[0]) ? "\\"
+                                                                                         : "/"};
+
+  if (m_strLibraryPath.size() && m_strLibraryPath.back() != pathSep[0])
+    m_strLibraryPath += pathSep;
+}
+
+void CKodiHost::SetProfilePath(const std::string& profilePath)
+{
+  m_strProfilePath = profilePath;
+
+  const char* pathSep{profilePath[0] && profilePath[1] == ':' && isalpha(profilePath[0]) ? "\\"
+                                                                                         : "/"};
+
+  if (m_strProfilePath.size() && m_strProfilePath.back() != pathSep[0])
+    m_strProfilePath += pathSep;
+
+  //let us make cdm userdata out of the addonpath and share them between addons
+  m_strProfilePath.resize(m_strProfilePath.find_last_of(pathSep[0], m_strProfilePath.length() - 2));
+  m_strProfilePath.resize(m_strProfilePath.find_last_of(pathSep[0], m_strProfilePath.length() - 1));
+  m_strProfilePath.resize(m_strProfilePath.find_last_of(pathSep[0], m_strProfilePath.length() - 1) +
+                          1);
+
+  kodi::vfs::CreateDirectory(m_strProfilePath.c_str());
+  m_strProfilePath += "cdm";
+  m_strProfilePath += pathSep;
+  kodi::vfs::CreateDirectory(m_strProfilePath.c_str());
+}
+
+bool CKodiHost::GetBuffer(void* instance, SSD::SSD_PICTURE& picture)
+{
+  return instance ? static_cast<kodi::addon::CInstanceVideoCodec*>(instance)->GetFrameBuffer(
+                        *reinterpret_cast<VIDEOCODEC_PICTURE*>(&picture))
+                  : false;
+}
+
+void CKodiHost::ReleaseBuffer(void* instance, void* buffer)
+{
+  if (instance)
+    static_cast<kodi::addon::CInstanceVideoCodec*>(instance)->ReleaseFrameBuffer(buffer);
+}
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1,0 +1,1458 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "Session.h"
+
+#include "aes_decrypter.h"
+#include "parser/DASHTree.h"
+#include "parser/HLSTree.h"
+#include "parser/SmoothTree.h"
+#include "samplereader/ADTSSampleReader.h"
+#include "samplereader/DummySampleReader.h"
+#include "samplereader/FragmentedSampleReader.h"
+#include "samplereader/SubtitleSampleReader.h"
+#include "samplereader/TSSampleReader.h"
+#include "samplereader/WebmSampleReader.h"
+#include "utils/Base64Utils.h"
+#include "utils/SettingsUtils.h"
+#include "utils/StringUtils.h"
+#include "utils/Utils.h"
+#include "utils/log.h"
+
+#include <kodi/addon-instance/Inputstream.h>
+
+using namespace UTILS;
+using namespace kodi::tools;
+using namespace SESSION;
+
+static const AP4_Track::Type TIDC[adaptive::AdaptiveTree::STREAM_TYPE_COUNT] = {
+    AP4_Track::TYPE_UNKNOWN, AP4_Track::TYPE_VIDEO, AP4_Track::TYPE_AUDIO,
+    AP4_Track::TYPE_SUBTITLES};
+
+
+CSession::CSession(const PROPERTIES::KodiProperties& kodiProps,
+                   const std::string& manifestUrl,
+                   const std::map<std::string, std::string>& mediaHeaders,
+                   const std::string& profilePath)
+  : m_kodiProps(kodiProps),
+    m_manifestUrl(manifestUrl),
+    m_mediaHeaders(mediaHeaders),
+    m_dummySampleReader(std::make_unique<CDummySampleReader>()),
+    m_KodiHost(std::make_unique<CKodiHost>()),
+    m_reprChooser(CHOOSER::CreateRepresentationChooser(kodiProps))
+
+{
+  m_KodiHost->SetProfilePath(profilePath);
+
+  switch (kodiProps.m_manifestType)
+  {
+    case PROPERTIES::ManifestType::MPD:
+      m_adaptiveTree = new adaptive::DASHTree(kodiProps, m_reprChooser);
+      break;
+    case PROPERTIES::ManifestType::ISM:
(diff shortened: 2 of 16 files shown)


CastagnaIT:
my god so big size PR i will need some time :)
I will try check minor things

src/main.cpp
```diff
@@ -33,6 +33,8 @@
 #include "utils/SettingsUtils.h"
```
CastagnaIT: all includes: `codechandler/*` can be removed

do you think we can move part of `OpenStream` to Session to remove all includes of `samplereader/*` from main?

src/main.h
```diff
@@ -48,210 +48,4 @@ namespace XBMCFILE
 Kodi Streams implementation
```
CastagnaIT: this comment can be removed
and the namespace `XBMCFILE` with all his static variables seem to be unused can be removed

src/main.cpp
CastagnaIT: there a bit of mess to be fixed also with all other includes:
example:
//#include "ADTSReader.h"
//#include "TSReader.h"
//#include "WebmReader.h"
//#include "aes_decrypter.h"
//#include "oscompat.h"
//#include "parser/DASHTree.h"
//#include "parser/HLSTree.h"
//#include "parser/SmoothTree.h"
//#include "utils/Base64Utils.h"
//#include "utils/SettingsUtils.h"
//#include "utils/StringUtils.h"
`#include <chrono>`
`#include <algorithm>`
are unused and probably also others

//#include <string.h> can be changed to `#include <string>`
and miss to add:
`#include <map>`

src/main.cpp
```diff
@@ -64,1577 +66,18 @@ using namespace kodi::tools;
...
-      if (stream->GetReader())
-        stream->GetReader()->Reset(true);
-    }
-    return true;
-  }
 
-  return false;
-}
 
 /***************************  Interface *********************************/
```
CastagnaIT: i know that there are a lot of changes but i think is appropriate
move to the header the definition of classes
And also the forward `class CInputStreamAdaptive;` of line 84 should be not more needed

src/main.h
```diff
@@ -10,7 +10,7 @@
 
 #include "AdaptiveByteStream.h"
 #include "SSD_dll.h"
-#include "common/AdaptiveStream.h"
+
 #include "common/AdaptiveTree.h"
 #include "common/Chooser.h"
 #include "samplereader/SampleReader.h"
```
CastagnaIT: also here we need to keep only the required includes for the definitions of classes

src/Session.cpp
```diff
@@ -0,0 +1,1456 @@
...
+    m_reprChooser(CHOOSER::CreateRepresentationChooser(kodiProps))
+
+{
+  m_KodiHost->SetProfilePath(profilePath);
+
+  switch (kodiProps.m_manifestType)
+  {
+    case PROPERTIES::ManifestType::MPD:
+      m_adaptiveTree =
+          std::make_unique<adaptive::DASHTree>(adaptive::DASHTree(kodiProps, m_reprChooser.get()));
```
CastagnaIT: i think that you have done a bit of mess with unique ptr's
you are using unique ptr's and sharing his raw pointers i think this is a bad implementation

I would rather leave everything as it was originally with raw pointers
and with a later PR change them with a shared_ptr instead of unique ptr
which has exactly the purpose of being shared safely

src/Session.cpp
```diff
@@ -0,0 +1,1456 @@
...
+      break;
+    case PROPERTIES::ManifestType::ISM:
+      m_adaptiveTree = std::make_unique<adaptive::SmoothTree>(
+          adaptive::SmoothTree(kodiProps, m_reprChooser.get()));
```
CastagnaIT: Sidenote for `std::make_unique`
when you create a unique ptr is rendundant create the object inside another contstructor,
std::make_unique is already a constructor then for example this become:
`m_adaptiveTree = std::make_unique<adaptive::SmoothTree>(kodiProps, m_reprChooser.get());`

src/common/Chooser.cpp
```diff
@@ -21,28 +21,28 @@ using namespace CHOOSER;
 
 namespace
 {
-IRepresentationChooser* GetReprChooser(std::string_view type)
+std::unique_ptr<IRepresentationChooser> GetReprChooser(std::string_view type)
```
CastagnaIT: same here as above, unique ptr shared as raw pointer
preferable as it was originally or use shared ptr maybe on other PR

CastagnaIT: another sidenote: you have changed the return value (here and other places) to `unique_ptr`
but in all places when you use a unique_ptr you should use `std::move`
because you must transfer ownership of the ptr, this is missing everywhere

CastagnaIT:
i have add some first comments but not done it will be for another day

src/main.cpp
glennguy: > do you think we can move part of OpenStream to Session to remove all includes of samplereader/* from main?

Yes but probably better for a separate PR. Looks like `CStream` would be a good place to handle this and use the factory pattern to create the readers (pass in rep to get flags/containertype). ~150 lines is too much for a single function

src/Session.cpp
glennguy: Thanks for the tip, I thought it looked very verbose!

src/Session.cpp
glennguy: > i think that you have done a bit of mess with unique ptr's
Yeah sorry, I did make a mess. I've removed for chooser because yes it was being abused, but do you think it's appropriate to leave any of the adaptiveTree/dllHelper/KodiHost ones as unique_ptr?

CastagnaIT: dllHelper and KodiHost unique_ptr are ok
but m_adaptiveTree is not, better keep raw for now and change to sharedptr in future prs

src/main.h
```diff
@@ -8,250 +8,88 @@
...
 
-#include "AdaptiveByteStream.h"
-#include "SSD_dll.h"
-#include "common/AdaptiveStream.h"
-#include "common/AdaptiveTree.h"
-#include "common/Chooser.h"
-#include "samplereader/SampleReader.h"
-#include "utils/PropertiesUtils.h"
-
 #include <float.h>
```
CastagnaIT: to remove

CastagnaIT: please do a clang format to whole header file

src/main.cpp
```diff
@@ -33,1689 +26,29 @@
...
-#include <sstream>
 #include <stdarg.h> // va_list, va_start, va_arg, va_end
-#include <stdio.h>
-#include <string.h>
 
 #include <kodi/Filesystem.h>
 #include <kodi/General.h>
-#include <kodi/addon-instance/VideoCodec.h>
 #include <kodi/addon-instance/inputstream/StreamCodec.h>
 #include "kodi/tools/StringUtils.h"
```
CastagnaIT: unused includes:
//#include "oscompat.h"
//#include "utils/Base64Utils.h"
//#include "utils/SettingsUtils.h"
//#include "utils/StringUtils.h"
//#include <kodi/Filesystem.h>
//#include <kodi/General.h>
//#include <kodi/addon-instance/inputstream/StreamCodec.h>
//#include "kodi/tools/StringUtils.h"

src/main.cpp
```diff
@@ -33,1689 +26,29 @@
...
-}
-
-STREAM_CRYPTO_KEY_SYSTEM Session::GetCryptoKeySystem() const
-{
-  if (m_kodiProps.m_licenseType == "com.widevine.alpha")
-    return STREAM_CRYPTO_KEY_SYSTEM_WIDEVINE;
-#if STREAMCRYPTO_VERSION_LEVEL >= 1
-  else if (m_kodiProps.m_licenseType == "com.huawei.wiseplay")
-    return STREAM_CRYPTO_KEY_SYSTEM_WISEPLAY;
+#ifdef CreateDirectory
```
CastagnaIT: this is unused

src/main.cpp
```diff
@@ -33,1689 +26,29 @@
...
-  bool IsRealTimeStream() override;
 
-#if INPUTSTREAM_VERSION_LEVEL > 1
-  int GetChapter() override;
-  int GetChapterCount() override;
-  const char* GetChapterName(int ch) override;
-  int64_t GetChapterPos(int ch) override;
-  bool SeekChapter(int ch) override;
-#endif
+#define STREAM_TIME_BASE 1000000
```
CastagnaIT: to be removed

src/main.cpp
```diff
@@ -33,1689 +26,29 @@
...
 
-  std::shared_ptr<Session> GetSession() { return m_session; };
+using namespace UTILS;
+using namespace kodi::tools;
```
CastagnaIT: to be removed

src/Stream.cpp
```diff
@@ -0,0 +1,33 @@
...
+void CStream::Reset()
+{
+  if (m_isEnabled)
+  {
+    m_streamReader.reset();
+    m_streamFile.reset();
+    m_adByteStream.reset();
+    m_mainId = 0;
+  }
+}
```
CastagnaIT: missing empty end line

src/Stream.h
```diff
@@ -0,0 +1,99 @@
...
+  adaptive::AdaptiveStream m_adStream;
+  kodi::addon::InputstreamInfo m_info;
+  bool m_hasSegmentChanged;
+  bool m_isValid;
+
+private:
+  std::unique_ptr<ISampleReader> m_streamReader;
+  std::unique_ptr<CAdaptiveByteStream> m_adByteStream;
+  std::unique_ptr<AP4_File> m_streamFile;
+};
```
CastagnaIT: missing empty end line

src/common/Chooser.cpp
```diff
@@ -42,7 +42,7 @@ IRepresentationChooser* GetReprChooser(std::string_view type)
 IRepresentationChooser* CHOOSER::CreateRepresentationChooser(
     const UTILS::PROPERTIES::KodiProperties& kodiProps)
 {
-  IRepresentationChooser* reprChooser{nullptr};
+  IRepresentationChooser* reprChooser;
```
CastagnaIT: forgot to restore `{nullptr}`

src/samplereader/SubtitleSampleReader.h
```diff
@@ -9,7 +9,7 @@
 #include "../codechandler/TTMLCodecHandler.h"
 #include "../codechandler/WebVTTCodecHandler.h"
 #include "../common/AdaptiveStream.h"
-#include "../main.h" //! @todo: Relying on for STREAM objects, to be removed in a future cleanup
+#include "../Session.h" //! @todo: Relying on for STREAM objects, to be removed in a future cleanup
```
CastagnaIT: you can full remove the "todo" comment

src/test/TestDASHTree.cpp
```diff
@@ -21,10 +21,10 @@ class DASHTreeTest : public ::testing::Test
   {
     UTILS::PROPERTIES::KodiProperties kodiProps;
 
-    m_reprChooser = new CTestRepresentationChooserDefault();
+    m_reprChooser = std::make_unique<CTestRepresentationChooserDefault>(CTestRepresentationChooserDefault());
```
CastagnaIT: to restore raw pointer behaviour for chooser in whole file
also the other test file

src/parser/HLSTree.cpp
```diff
@@ -85,13 +85,13 @@ static std::string getAudioCodec(const std::string& codecs)
 }
 
 HLSTree::HLSTree(const HLSTree& left)
-  : AdaptiveTree(left.m_kodiProps, left.m_reprChooser), m_decrypter(left.m_decrypter)
+  : AdaptiveTree(left.m_kodiProps, left.m_reprChooser)
 {
+  m_decrypter = std::make_unique<AESDecrypter>(AESDecrypter(left.m_kodiProps.m_licenseKey));
```
CastagnaIT: rendundant constructor

src/Stream.h
```diff
@@ -0,0 +1,99 @@
...
+      m_mainId{0},
+      m_adStream{tree, adp, initialRepr, mediaHeaders, playTimeshiftBuffer, chooseRep},
+      m_hasSegmentChanged{false},
+      m_isValid{true} {};
+
+
+  ~CStream() { Disable(); };
+
+  /*!
+     * \brief Stop/disable the AdaptiveStream and reset
```
CastagnaIT: i think copy/paste has broken all comments spaces

src/main.cpp
glennguy: For some reason this was needed back when adding using dllhelper, it wouldn't compile on some platforms. Seems to not be an issue now :)

src/KodiHost.cpp
```diff
@@ -0,0 +1,121 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#pragma once
```
CastagnaIT: to be removed

src/Session.cpp
```diff
@@ -0,0 +1,1459 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#pragma once
```
CastagnaIT: to be removed

src/Stream.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+ *  Copyright (C) 2022 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#pragma once
```
CastagnaIT: to be removed

CastagnaIT:
there is this linux warning needs to be solved
```
/home/runner/work/inputstream.adaptive/inputstream.adaptive/inputstream.adaptive/src/Stream.h:18:7: warning: ‘CStream’ declared with greater visibility than the type of its field ‘CStream::m_adStream’ [-Wattributes]
 class CStream
       ^~~~~~~
```
could be due to namespaces because m_adStream is public
a my possible idea is to make `CStream` derived from `AdaptiveStream` but i think this require more works,
and since also all other members should not be exposed directly,
we can do it later

Only just last observation,
i noticed that `CSession` class is without a namespace
perhaps it is worth add a namespace to group possible future classes or other definitions that concern session scope
where maybe also `CStream` could be part of "session" namespace?

glennguy:
I think that should be fixed with adding ATTR_DLL_LOCAL macro to the class declaration

glennguy:
@CastagnaIT how are we looking now?

src/parser/HLSTree.h
```diff
@@ -30,10 +29,8 @@ class ATTR_DLL_LOCAL HLSTree : public AdaptiveTree
     ENCRYPTIONTYPE_UNKNOWN = 4,
   };
   HLSTree(const UTILS::PROPERTIES::KodiProperties& kodiProps,
-          CHOOSER::IRepresentationChooser* reprChooser,
-          IAESDecrypter* decrypter)
-    : AdaptiveTree(kodiProps, reprChooser), m_decrypter(decrypter){};
-  HLSTree(const HLSTree& left);
+          CHOOSER::IRepresentationChooser* reprChooser);
+  HLSTree(const HLSTree& left) : AdaptiveTree(left.m_kodiProps, left.m_reprChooser){};
```
CastagnaIT: this method was on cpp, can you keep it on cpp?

CastagnaIT:
> I think that should be fixed with adding ATTR_DLL_LOCAL macro to the class declaration

good to know
only methods descriptions still formatted with bad spaces where before was ok
after that i think we can go on
good work

src/parser/HLSTree.h
glennguy: Sure thing

glennguy:
>only methods descriptions still formatted with bad spaces where before was ok 

Damn, I think that must have been adding the namespace (VS then indents everything) and clang format only moved the first line of each comment back... all good now!

CastagnaIT (approved):

glennguy:
Thanks for all your help @CastagnaIT

Issue #996, opened 2022-05-19
large audio sync error after skipping in some streams, making it jump back


TermeHansen:
### Describe the bug
When playing a certain stream it fails to skip ahead correctly in the first part of the stream. In the kodi log I can see `large audio sync error` so I think the problem is related to some discrepancy between audio and video stream time segment for the first 10min of the stream.

## Expected Behavior
video skipping correctly forward

## Actual Behavior
Starting video and skipping anything up to 10min ahead and it will jump back to only 10 seconds ahead. Jumping 20min ahead works perfect, and moreover skipping backwards again into anywhere in the first 10min of the stream also works correct. So something around the definitions of the first segment I believe as the problem only seem to come when stream is playing from the start.

### To Reproduce
Stream is geo locked to danish ips I believe

```
#KODIPROP:inputstream=inputstream.adaptive
#KODIPROP:inputstream.adaptive.manifest_type=hls
https://drod20c.akamaized.net/dk/encrypted/none/ea/62694ba3ac5a6120fcf966ea/00022163280/stream_fmp4/master_manifest.m3u8

```

### Debuglog and  M3U8s
The debuglog,, master.m3u8 and m38u8 files for video and audio an be found here:

[https://paste.c-net.org/HondaRanking](https://paste.c-net.org/HondaRanking)
[full_debug.log](https://paste.c-net.org/PricesBruno)

```
2022-05-19 22:40:53.465 T:192929    INFO <general>: CDVDVideoCodecFFmpeg::CDropControl: calculated diff time: 40000
2022-05-19 22:40:54.571 T:192924    INFO <general>: AddOnLog: inputstream.adaptive: PosTime (601345)
2022-05-19 22:40:54.611 T:192924    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(601.3) for Stream:1 continues at 610.0 (PTS: 610080000)
2022-05-19 22:40:54.680 T:192924    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(610.0) for Stream:3 continues at 610.0 (PTS: 610069333)
2022-05-19 22:40:54.707 T:192924    INFO <general>: AddOnLog: inputstream.adaptive: seekTime(610.0) for Stream:4 continues at 596.9 (PTS: 596960000)
2022-05-19 22:40:54.751 T:192929 WARNING <general>: OutputPicture - timeout waiting for buffer
[...]

PR #998, opened 2022-06-04
Handle AES PKCS7 padding

glennguy:
To fix the issue in #996 

Quick rundown of how padding works:
Before encryption padding is added to all files to round up to the nearest 16 byte boundary. The characters added all have the value of how many padding bytes are added. And if the file length is already divisible by 16 then 16 bytes are added eg.
21 byte file will be padded out to 32 bytes, last 11 bytes are all `0b`. 32 byte file will be padded out to 48 bytes, last 16 bytes are all `10`
After decryption this padding is detected and removed.

This does not happen currently however 99% of AES-128 encrypted HLS is TS files, which are apparently robust enough to not cause any problems. However in the issue above it's fmp4 and bad things happen when there's random stuff added to the end of the file.

Changes made to accomplish this:
I don't know if this is a very nice way to do this but I stuck largely with how the buffers are laid out - the underlying object in `segment_buffers_` is a string put on the stack and now this must be available to operate on inside aes_decrypter, as it needs to be resized once the whole file is processed.
Also the aes decrypter needs to have knowledge of when the end of a file is being decrypted, so the `lastChunk` flag needs to be passed through to it.

@CastagnaIT maybe this is not for this PR but do you have any ideas around how this chain of functions could be improved? There's lots of parameters being passed around and I wonder if it would be better to wrap it into an object?

@TermeHansen could you please test out one of the builds [here](https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-998/1/artifacts) for Nexus?

### DIFF
--- a/src/aes_decrypter.cpp
+++ b/src/aes_decrypter.cpp
@@ -7,11 +7,18 @@
  */
 
 #include "aes_decrypter.h"
-#include <bento4/Ap4Protection.h>
+#include "utils/log.h"
+#include <bento4/Ap4StreamCipher.h>
 #include <kodi/Filesystem.h>
 #include <vector>
 
-void AESDecrypter::decrypt(const AP4_UI08 *aes_key, const AP4_UI08 *aes_iv, const AP4_UI08 *src, AP4_UI08 *dst, size_t dataSize)
+void AESDecrypter::decrypt(const AP4_UI08* aes_key,
+                           const AP4_UI08* aes_iv,
+                           const AP4_UI08* src,
+                           std::string& dst,
+                           size_t dstOffset,
+                           size_t& dataSize,
+                           bool lastChunk)
 {
   AP4_BlockCipher* cbc_d_block_cipher;
   AP4_DefaultBlockCipherFactory::Instance.CreateCipher(
@@ -23,9 +30,16 @@ void AESDecrypter::decrypt(const AP4_UI08 *aes_key, const AP4_UI08 *aes_iv, cons
     16,
     cbc_d_block_cipher);
 
-  cbc_d_block_cipher->Process(src, dataSize, dst, aes_iv);
-
-  delete cbc_d_block_cipher;
+  AP4_CbcStreamCipher cbcStreamCipher{cbc_d_block_cipher};
+  cbcStreamCipher.SetIV(aes_iv);
+  AP4_Result result{
+      cbcStreamCipher.ProcessBuffer(src, dataSize, reinterpret_cast<AP4_UI08*>(&dst[0] + dstOffset),
+                                    reinterpret_cast<AP4_Size*>(&dataSize), lastChunk)};
+  if (!AP4_SUCCEEDED(result))
+  {
+    LOG::LogF(LOGERROR, "AES decryption failed: %d", result);
+  }
+  dst.resize(dstOffset + dataSize);
 }
 
 std::string AESDecrypter::convertIV(const std::string &input)
--- a/src/Iaes_decrypter.h
+++ b/src/Iaes_decrypter.h
@@ -20,8 +20,10 @@ public:
   virtual void decrypt(const AP4_UI08* aes_key,
                        const AP4_UI08* aes_iv,
                        const AP4_UI08* src,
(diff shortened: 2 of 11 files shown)


src/aes_decrypter.cpp
```diff
@@ -23,8 +30,21 @@ void AESDecrypter::decrypt(const AP4_UI08 *aes_key, const AP4_UI08 *aes_iv, cons
     16,
     cbc_d_block_cipher);
 
-  cbc_d_block_cipher->Process(src, dataSize, dst, aes_iv);
+  cbc_d_block_cipher->Process(src, dataSize, reinterpret_cast<AP4_UI08*>(&dst[0] + dstOffset),
```
CastagnaIT: instead to use `cbc_d_block_cipher->Process` is possible that this method:
```c++
AP4_CbcStreamCipher cbcStreamCipher{cbc_d_block_cipher};
cbcStreamCipher.ProcessBuffer(...)
```
already do the padding check?

CastagnaIT:
> maybe this is not for this PR but do you have any ideas around how this chain of functions could be improved? There's lots of parameters being passed around and I wonder if it would be better to wrap it into an object?

atm i dont have a clear vision on this but maybe there is also the possibility to remove a bit of parameters
i can try give a look but atm i am busy with other

TermeHansen:
@glennguy Just tested my side on Nexus - and it solves the issue reported (for both streams) - great stuff!

src/aes_decrypter.cpp
glennguy: Thanks for pointing this out. I tried to do this way initially but was missing passing the block cipher into the stream cipher constructor - what I came up with originally was just copy/paste/adjust what was in AP4_CbcStreamCipher::DecryptBuffer to remove the padding.
Looks much nicer now and cbcStreamCipher also takes care of deleting the block cipher :)

glennguy:
Thanks @TermeHansen for all of your help getting me access to debug this properly.

src/aes_decrypter.cpp
CastagnaIT: nice! for safety better if @TermeHansen give another test
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-998/2/artifacts

TermeHansen: yep, still works...

CastagnaIT: Thank you

CastagnaIT (approved):

TermeHansen:
> Thanks @TermeHansen for all of your help getting me access to debug this properly.

no problem, this shows how open-source community works the best, and I also learned a new feature of how to use zerotier...

TermeHansen:
I will make a backport branch to Matrix and test in my fork and share in a PR when ready...

glennguy:
> I will make a backport branch to Matrix and test in my fork and share in a PR when ready...

Awesome, thank you so much

L2501:
This seems to fail on segments that are using chunked Transfer-Encoding (without Content-Length header)

glennguy:
@L2501 thanks for the report - so I can understand more quickly (I don't have something that returns chunked encoded to test), is it:
* trying to strip PKCS7 padding from the end of every chunk; or
* not stripping PKCS7 padding at all from chunked transfers

L2501:
@glennguy file.AtEnd() is True at every chunk so it is trying to strip PKCS7 padding from the end of every chunk

glennguy:
Thanks for that.

I've quickly done up https://github.com/xbmc/inputstream.adaptive/pull/1010 - would you be able to check if that fixes things on your end?

@TermeHansen could you double check I haven't broken for you with this either?
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1010/1/artifacts

I'm mindful that the next (and possibly last) version of Matrix is dropping imminently and don't want to leave problems for those platforms with it baked in (e.g Xbox)

TermeHansen:
I will check tonight (CET), and write here...

Issue #976, opened 2022-05-12
Incorrect behaviour when with data interruptions


glennguy:
details of issue in https://github.com/xbmc/xbmc/issues/20742

IA handles DemuxRead calls appropriately when we reach the live edge of the manifest (sets `WAITFORSEGMENT`) and empty demux packets are returned until this condition changes.

For cases when we're waiting on data, (need to confirm properly) it seems we just make VP wait.

glennguy:
@david-morris @kontell this is behind the issues you opened

kontell:
Well done on narrowing down the issue, let me know if you want any testing done.

glennguy:
Closed with #1008

Issue #973, opened 2022-05-10
Buffering/ Fast-Forwarding (re-syncing) with Marginal Internet Connection


kontell:
### Describe the bug
Here is a clear and concise description of what the problem is:
The buffering behaviour is much improved with this release however it doesn't pause/ buffer gracefully when the network connection is inadequate. This isn't new behaviour for Kodi, I've seen it in V19. 

## Expected Behavior
Here is a clear and concise description of what was expected to happen:
Playback should pause with the buffer symbol on screen and playback should resume from previously paused location when the bufffer is created.

## Actual Behavior
When live streaming with a marginal internet connection the buffering behaviour is jittery. 

The video freezes - for a period of less than 1s to over a minute - and will then fast-forward to what seems to be live. Sometimes the audio will keep playing and the video will resync with the audio. The buffer circle symbol is not displayed when the video is frozen.

Other times the Kodi buffer symbol will show with a percentage and playback will alternate between playing/ buffering rapidly rather than a real buffer being created.

## Possible Fix

Choose a lower bitrate stream so the connection is no longer marginal and playback is fine.

### To Reproduce

1. Choose "Manual OSD" for stream selection type in Inputstream Adaptive settings.
2. Play live stream with a bitrate higher than the network connection can comfortably handle.

### Debuglog

The debuglog can be found here:
[kodi_2022-05-10.log](https://github.com/xbmc/inputstream.adaptive/files/8662291/kodi_2022-05-10.log)

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
skysportnow.co.nz (subscription service, can provide login for testing).

### Screenshots 
Here are some links or screenshots to help explain the problem:

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:
Same/ similar problem:
https://forum.kodi.tv/showthread.php?tid=309627

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Android 11 (Nvidia Shield 9.0.1)
 - Kodi version: 20.0-ALPHA1 (19.90.101) (2022-05-09)
 - inputstream.adaptive version: 20.2.0

CastagnaIT:
is a know problem and it's a much bigger problem this could could be improved for final Kodi 20
[...]

PR #1000, opened 2022-06-04
Send empty demux packets when waiting for data

glennguy:
Very early test to see if this gives better behaviour when a stream is starved of data. To solve #976, and related to #973 and https://github.com/xbmc/xbmc/issues/20742

The idea here is to give the dummy sample reader when requesting the next sample for DemuxRead under the right conditions. So far the 'right' condition in my mind is that the download worker is flagged as processing (i.e trying to download), we only have 1 buffer that is valid, and the available bytes to read are 0 (segment read position is the size of the buffer currently).

Assuming that is the correct condition, I'm not sure about the threading part of this. I have it set to get the reading lock, I just get nervous because the way the locks are setup at the moment can cause deadlocking already in test conditions (unlikely to happen in real world though...)

Testing with normal streams seems to be fine for regular playback, however I'm yet to simulate a bad connection.

@kontell would you be able to try and test? Zips are at https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1000/1/artifacts

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1107,7 +1107,7 @@ ISampleReader* CSession::GetNextSample()
     {
       if (!res || streamReader->DTSorPTS() < res->GetReader()->DTSorPTS())
       {
-        if (stream->m_adStream.waitingForSegment(true))
+        if (stream->m_adStream.WaitingForData() || stream->m_adStream.waitingForSegment(true))
           waiting = stream.get();
         else
           res = stream.get();
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -1075,6 +1075,19 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
   return false;
 }
 
+bool AdaptiveStream::WaitingForData() const
+{
+  {
+    std::unique_lock<std::mutex> lckrw(thread_data_->mutex_rw_);
+    if (worker_processing_ && valid_segment_buffers_ == 1 &&
+      segment_buffers_[0].buffer.size() - segment_read_pos_ == 0)
+    {
+      return true;
+    }
+  }
+  return false;
+}
+
 bool AdaptiveStream::waitingForSegment(bool checkTime) const
 {
   if (tree_.HasUpdateThread())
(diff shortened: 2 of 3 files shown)


kontell:
I'll start testing this tomorrow, thanks.

kontell:
This is difficult to test as my network quality is variable and the majority of time playback is ok with the buffering issues being erratic. 

Subjectively it seems that the new version is doing better but the connection is too random to be able to tell for sure (the buffering appears to be more graceful and I haven't seen the freeze/ fast-forwarding behaviour).

I tried to be more rigorous about it and have a bandwidth limited network connection so as to force buffering. I did tests with 3Mb & 4.5Mb streams on a connection limited to 4Mb.

The tests were with the Jenkins version of IA from above and with the repository version. Logs are attached with comments at the beginning of the files. The buffering is still erratic with some of the fast-forwarding seen previously.

I'll continue to use the new version for normal streaming and in a week it should be clear if it's performing better and I'll see if I can get some consistent behaviour out of the bandwidth limited tests.

[01_4.5Mb@4Mb_Jenkins_kodi.log](https://github.com/xbmc/inputstream.adaptive/files/8840128/01_4.5Mb%404Mb_Jenkins_kodi.log)
[02_3Mb@4Mb_Jenkins_kodi.log](https://github.com/xbmc/inputstream.adaptive/files/8840129/02_3Mb%404Mb_Jenkins_kodi.log)
[03_3Mb@4Mb_Repository_kodi.log](https://github.com/xbmc/inputstream.adaptive/files/8840130/03_3Mb%404Mb_Repository_kodi.log)
.

Arias800:
Hello,
I have tested this pull request with the wnbc addon, and it seems that's fix the playback issue on my side too.
Now, I'm able to watch the content without any weird behavior.

Arias800:
Well, I find something.
Sometimes, the video stopped, and the subtitles get out of sync.
This is a video of the behavior, the log and the manifest: 
https://www.youtube.com/watch?v=3cHkfr1L1vU
Log : https://github.com/Arias800/Fichier-divers/blob/master/kodi.log
Manifest : https://github.com/Arias800/Fichier-divers/blob/master/master.m3u8

If you want to reproduce this, you can use this subtitles files.
https://github.com/Arias800/Fichier-divers/blob/master/subtitle.ssa

glennguy:
Thanks for testing @Arias800 @kontell 

I think the code so far in this PR is incorrect. I have made some adjustments on my end and can now get the desired behaviour if I hard code in download delays, however I haven't been able yet to find the best place to test for the condition. Will give you an update when there's something to test again.

glennguy:
Closing this in favor of incoming solution that actually works :)

PR #1001, opened 2022-06-05
Reinitialize cdm video decoder only when codec change

CastagnaIT:
`OpenVideoDecoder` is called each time that the stream quality change and currently force call `InitializeVideoDecoder` in the cdm during playback, at least with N@tflix this cause cdm errors when the stream quality is switched during the playback, so freeze the playback

A possible solution is call `DeinitializeDecoder` before recall `InitializeVideoDecoder`,
but as is implemented now cause a delay (image freezed for some secs),

from a search on chromium sources seem that they allow call `InitializeVideoDecoder`
during playback, but only in the situation that the codec change,
trying to do this i can see that the playback now works when stream quality change,
we will see if someone report a regression but it should not

Since i compared the sources, i took the opportunity to do some cleaning, and to make some parts more similar to chromium conterpart. Changes:
- `DecodeVideo` method is now splitted in to: `DecryptAndDecodeVideo`, `VideoFrameDataToPicture`
- Moved `host` global variable to new Helper.h, to allow have access to log methods without being restricted by each class (in a future prospective to have the main/ssd projects merged)
- Add cdm_type_conversions.h (adapted from chromium sources) to contains all cdm data type conversions
- Removed unnecessary reallocations and member variables in `WV_CencSingleSampleDecrypter` in favour of simple vectors
- Implemented all fields in `SSD_SAMPLE` struct, to have full `DEMUX_PACKET` crypto data

The only functional change in the cleanup is that the `cdm::InputBuffer_2 cdm_in;` defined in to previous `DecodeVideo` method
is now constructed by using all info of SSD_SAMPLE (DEMUX_PACKET) data, instead of take some info from members variables like m_CryptBlocks, m_SkipBlocks, m_EncryptionScheme

~WIP: not done android side cleanups~

### DIFF
--- a/src/SSD_dll.h
+++ b/src/SSD_dll.h
@@ -42,7 +42,7 @@ namespace SSD
     {
       PROPERTY_HEADER
   };
-    static const uint32_t version = 18;
+    static const uint32_t version = 19;
 #if defined(ANDROID)
     virtual void* GetJNIEnv() = 0;
     virtual int GetSDKVersion() = 0;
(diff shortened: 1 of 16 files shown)


glennguy (approved):
Amazing, looks really good!

glennguy:
So does this then solve the issue we would have on amlogic/CE devices where video is lost after the first stream change?

CastagnaIT:
yes tested (with n@tflix) on CoreElec when DEMUX_SPECIALID is signaled not stuck the playback anymore,
although not yet perfect, the quality switching:
-with H264, works but frames seem to be mixed for 1-2 secs, after that playback continues as normal
-with VP9, no problems, as usual the little black screen transition only

comparing to Windows build, the quality switching:
-with H264, works but transition can freeze image for 1 sec, after that playback continues as normal
-with VP9, no problems, transition looks smooth without black screen transition

glennguy:
Great! It's certainly better than no picture at all. Maybe some improvements on Kodi side can make it more seamless.

Issue #1003, opened 2022-06-07
[20.2.0-Nexus and up] Different behavior of Base64 decoder


dobo90:
After updating inputstream adaptive to 20.2.1-Nexus one of video addons has stopped working. Git bisect has shown that commit **1b9e2d01 [cleanup] Rework to new utils** breaks it. After some hours of digging it came out that stream provider sends a license which contains dummy backslash characters.

An example of received license:
[...]

PR #1006, opened 2022-06-09
[utils][base64] Improved decoding

CastagnaIT:
Improved base 64 decode method,
compared to the method implemented in kodi core this has been backported from cpython
which performs more secure decoding checks
and also automatically filters out characters that do not fit into the table of 64 ascii
so allow to decoding malformed base64 string case like #1003


glennguy (approved):
Thanks @CastagnaIT

PR #1008, opened 2022-06-16
Send empty packets instead of blocking DemuxRead

glennguy:
To solve https://github.com/xbmc/inputstream.adaptive/issues/976, and related to https://github.com/xbmc/inputstream.adaptive/issues/973 and https://github.com/xbmc/xbmc/issues/20742

Currently DemuxRead calls ReadSample on the sample reader whose data was sent in the packet. If there's no data left in the buffers to read then we block here until downloading of the next segment has finished.
VideoPlayer is unaware of this and what can be observed is the 'clock' still runs and VP wants to still be where the stream would otherwise be if data was not interrupted, giving us the 'hyperspeed' fast forwarding when data comes back. If data isn't downloaded fast enough to get back to the 'now' point, we don't get proper playback with audio. My Australian neighbours with crap data connections will know this very well!

This PR gives VP what it expects when IA's buffer empties - empty demux packet(s) until more data is available. This happens by calling ReadSample asynchronously. If the call isn't finished by the time DemuxRead is called again, then we give it an empty packet.

Advice from fernetmenta has been to not keep VP waiting longer than 10ms, so I've left the wait time for checking the future object at 0. When decrypting for Widevine it usually always takes longer for ReadSample to finish than by the time the next DemuxRead is called, and you can observe if you uncomment https://github.com/glennguy/inputstream.adaptive/blob/1e51da47b408c5033358392a9ddfa5c836f2e911/src/main.cpp#L490 that we alternate between real packet and 0 packet. Nothing bad has been observed from this.

There are checks included to make sure ReadSample has returned before stopping streams and when starting a stream (in case of changing period).

@Arias800 @kontell, as far as I can tell this works well. No more 'hyperspeed' behaviour.


Arias800:
Hello,
I did some tests, and here is the result:
- The first episode played without any problem.
- The second one, I tested 3 times to be sure, the video froze during playback.
The audio goes back a few seconds and the video is forced to wait.

It's strange that this doesn't happen in the first episode (maybe they're a difference between these two hls stream).
This is similar to what I reported for the previous pull request.

And of course, the subtitle problem is still there.
When the video froze, the subtitles are out of sync.

Apart from these 2 issues, which may not even related to these changes, the pull requests work because there is no "hyperspeed" behavior.

glennguy:
Thanks for the feedback.
I'd be interested to know if the video freezing in your second video happens with the release version of 20.2.2

I've seen the subtitle issue where it gets out of sync without these changes, so I'm sure it's not related.

Arias800:
I have manually compiled IA with the current master branch.
So yes, the version 20.2.2 doesn't fix the issue.

But, I have updated Kodi 20 with the last change and I can't reproduce anymore.
I don't know if it's a coincidence, but I think it's fixed.

glennguy:
Ok thanks for the feedback. I'll take it as there being no regressions

CastagnaIT:
i tried do a couple of tests,

as first i tried https://github.com/MrKrabat/plugin.video.crunchyroll, this always had buffer problems when starting/seek the video, that now seems are solved, i will need to try also my HT pc

instead trying this http://media.axprod.net/ExoPlayer/Captions2/Manifest.mpd when press stop crashes

will do other test on next days

for GetNextSample method you have described the return as the sample reader for the next chronological sample
but should return the reader of the stream with the lower PTS value, its right?
to me sound a more understandable description

is possible that the dummy reader could be removed, I have not looked in depth but it seems unnecessary

i am a bit worried for `stream->m_readSample = std::async(std::launch::async, &ISampleReader::ReadSample, sr);`
its the first time that i see the use of `std::async` then i need to do some research
but this should generate a thread, so what happens if we try access to some shared variables with reader methods in the middle of an async process? is possible that happens? need add mutex?

> Advice from fernetmenta has been to not keep VP waiting longer than 10ms

i think we should add a comment in the DemuxReader method to keep in mind for the future this max latency of about 10ms

glennguy:
Yes you're probably right that it's not safe at the moment, infact GetNextSample itself is checking variables that may be updated by the separate thread. 

With std::async my understanding is it's a nice wrapper for detaching a thread which calls back to update the future variable when it's done. I'll look into the crashing test stream during waking hours :)

glennguy:
@CastagnaIT I've been checking the stream you wrote but can't reproduce the crash. Could it be the libass thing (I kept getting caught out with this, would see crashes and then be disappointed, then remember about copying kodi.exe over existing installation and attach debugger, find out that crash is unrelated...)

Anyway I've reviewed my code again and made a couple of changes - now `WaitForReadSample` is called in `GetNextSample` before looking at any sample reader values. I've fixed the other locations where we wait as well so that anything wanting to access the reader from the main thread will wait to ensure the detached thread is finished first.

~~There is one exception at the moment: `CSession::OnSegmentChanged`
This is called from `AdaptiveStream::ensureSegment` which is normally always called from the main thread, *except* in SubtitileSampleReader `ensureSegment` is called from `ReadSample` itself: https://github.com/xbmc/inputstream.adaptive/blob/90d28beadeddeda572f58d79ee17072e5a3defd8/src/samplereader/SubtitleSampleReader.cpp#L77~~
~~I will look further into it but is there any other way we can achieve what you want to happen in `SubtitleSampleReader::ReadSample` without the call to `ensureSegment`?~~ nvm, realised that this seems to come from all `ReadSample` calls

glennguy:
@CastagnaIT I think it's now safe. `GetNextSample` will now only proceed with reader operations once the async task has finished. I've gone and checked all other functions that access the reader on the main thread and they're all protected with the wait call, so we should be fine without needing to do mutex locks. Please let me know what you think.

CastagnaIT:
yes now i think its better
these are the changes that i was telling you, please give a look and if possible merge it with your changes
https://github.com/CastagnaIT/inputstream.adaptive/commit/1406112d85556885968ae6f259976f22ea0f0c29

what i have done,
since we are modify GetNextSample method its better do a little cleanup,
to do this cleanup, as said in previous message we delete the dummy reader class because is a noop,
then can be replaced by an empty package from AllocateDemuxPacket nothing else

the std::future its now a private member of reader interface, there is no need to expose the member only add more clear methods
in the CSession::SeekTime method i have deleted an error log, this is not an oversight, was just spamming the log

Before doing the review i would like to test these changes better
I will dedicate some time in the next few days

glennguy:
> please give a look and if possible merge it with your changes
Done, thanks for your help. Looks much better now.

glennguy:
@kontell @Arias800 latest changes I hope are good with no issues - win/mac/android builds here https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1008/6/artifacts

@Arias800 I know why you were seeing issues in the last one you tried and that is fixed now.

CastagnaIT:
i have done some tests,
unfurnately i think my wifi usb adaptor that i was using last year is broken too much error timeouts, 
i am unable to have a working connection then i cannot test with my HTPC on wifi
where usually hyperspeed was happening

i tried test on my CoreElec arm device (wifi), but nothing
i saw buffering only one time just after playback starts
and while the loading circle was filling some frames kept being displayed which is a bit strange
however no hyperspeed happened in this case
I will continue using this device for this week to see if i catch a buffering loading

src/Session.cpp
```diff
@@ -1102,37 +1099,55 @@ ISampleReader* CSession::GetNextSample()
...
-    return res->GetReader();
+    return true;
   }
-  else if (waiting)
+  else if (res)
   {
-    std::this_thread::sleep_for(std::chrono::milliseconds(100));
-    return m_dummySampleReader.get();
+    CheckFragmentDuration(*res);
+    ISampleReader* sr{ res->GetReader() };
```
CastagnaIT: please clang it

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
     m_failedSeekTime = ~0;
   }
 
-  ISampleReader* sr(m_session->GetNextSample());
+  ISampleReader* sr{ nullptr };
```
CastagnaIT: please clag all the modified code from here

src/samplereader/SampleReader.h
```diff
@@ -55,4 +57,34 @@ class ATTR_DLL_LOCAL ISampleReader
...
+   */
+  void ReadSampleAsync()
+  {
+    m_readSampleAsyncState = std::async(std::launch::async, &ISampleReader::ReadSample, this);
+  }
+
+  /*!
+   * \brief Wait for the asynchronous ReadSample to complete
+   */
+  void WaitReadSampleAsyncComplete() {
```
CastagnaIT: clang also here

CastagnaIT (changes requested):
only minor things and the `DummySampleReader.h` file need to be deleted

glennguy:
Changes made

If you want to simulate a crappy connection you can try this: http://jagt.github.io/clumsy/download.html
I'm using these settings with the SBS test stream 
image



Have the filter stopped, start the test stream (I choose the 720p rep), start the filter and then seek (stay in same period is probably best). The seek will complete after some seconds, playback resumes, then it will stop. Once data starts coming in we see the blue buffering circle:
image

kontell:
The long freeze followed by a fast-forward appears to be gone but I'm still seeing short duration freeze/ fast-forwarding. Video and log below, the fast-forward is clearest at 26s.

[kodi.log](https://github.com/xbmc/inputstream.adaptive/files/8948597/kodi.log)

https://user-images.githubusercontent.com/91025994/174792522-e200a406-a1a8-47ca-86e9-dd6185ddfc2a.mp4

glennguy:
Thanks for testing @kontell, I'll try to reproduce with a live stream and see if I get the 'mini' freezes.

glennguy:
@kontell I have some findings that need more investigation
@CastagnaIT the issue seen by kontell seems to be controlled by what we send to Kodi for `IsRealTimeStream`. Setting this to false gives a nicer experience where Kodi will stop to buffer properly. Without looking into VP (not that I'm likely to fully understand anyway) it kind of seems like sending true is asking Kodi "please don't let the 'now' time slip behind, resume playback after getting minimal buffer back".  

To reproduce:
```
https://livesim.dashif.org/livesim/segtimeline_1/testpic_2s/Manifest.mpd
```
limit d/l rate to 20K/s

glennguy:
@kontell also what are your settings for stream selection type? Are you using adaptive (automatic switching)? Is the 960*540 stream the lowest bandwidth one available?
edit - you'll need to change the settings in that slyguy add-on to bypass/disable the stream quality selector in order for IA to be able to see all the available streams.

kontell:
I've been using manual stream selection, there are two lower bandwidth streams available. I just chose a stream that gave buffering at the time. Do you want me to test with a stream that plays successfully?

glennguy:
This stream of yours is great to test. It would be great to see if the adaptive stream selection works well for you or not, this is a perfect use case.

I'm having a look over what's in VideoPlayer relating to live streams to see if these mini-fastforwards are actually intended behaviour or not - it definitely treats things differently if we flag it as a 'real-time' stream

kontell:
I'm using Android so the black screen when changing streams isn't great (should a bug report be created with Kodi for this?).

I've also found that it will often settle on the lowest bitrate stream even though higher ones will play ok (I mentioned this on the Kodi [forum](https://forum.kodi.tv/showthread.php?tid=364745&pid=3090818#pid3090818)). Happy to provide login details if you want or give logs.

glennguy:
Ah I see, I remember now.
@CastagnaIT Looking into VP in Kodi I can see that the realtime flag definitely affects how things are treated when the buffer runs low. I guess that things are affected differently depending on whether video is dry or audio is dry. On my test stream here I'm not really getting the mini fast forwards though (I don't think, hard to tell sometimes looking at a video of clock timer with beeps)

IMO there's not anything more to be done for this PR but for the future might be worth checking in with fernetmenta to see if there's any more advice to be had regarding this.

Stefano - do you have any ideas at the moment on the black screen on switching with Android?

CastagnaIT:
> Looking into VP in Kodi I can see that the realtime flag definitely affects ...

yesterday i put a bit of log prints to try debug more,
i noticed that in `DemuxRead` sometimes when we are in wait for "video" data stream (for buffering),
after a bit seem download an audio data and so return an audio packet
maybe its correct and no relevant, unfortunately it is still not clear to me how this relates to the buffering mechanism
this should happens when (with the test stream) the buffer loading screen resume playback, and we have the mini skipping
but difficult to say exactly...

> Stefano - do you have any ideas at the moment on the black screen on switching with Android?

depends what means for @kontell blackscreen,
- if we are talking of the little black screen transition (~1 sec) when change video stream quality, this happens also on others operative systems, VP in kodi need optimizations and we need find someone that know this area
- if we are talking that when change video stream quality, the screen become black and no more image after, then need to be investigated

I have a question that i keep asking myself
`DemuxRead` is called from Kodi many times very quickly
and with `async` we are creating hundreds of new threads every sec
creating new threads do an overhead in the CPU it is possible that this on low end devices is not so good
im not super expert in threads, perhaps one more review would be appropriate to have a voice from someone else on this

kontell:
What I've seen is a black screen when changing quality followed by a video resync (audio usually keeps playing). The duration of the black screen can be from <1s to multiple seconds.

The black screen isn't present on raspberry pi, the video just does a minor resync. I think there's an Android specific black-screen issue and a more general resyncing issue.

CastagnaIT:
yes at least on your android device the black screen transition is more prolonged,
but the behaviour is different from system to system and by video codec used, IMO its more generic problem

is possible reply this problem in different ways,
on Windows video can freeze for some secs,
on CoreElec i can see the little blackscreen, this usually with h264, VP9 seems to be more stable
however i think you will have to do something in Kodi VP

glennguy:
> yesterday i put a bit of log prints to try debug more,
i noticed that in DemuxRead sometimes when we are in wait for "video" data stream (for buffering),
after a bit seem download an audio data and so return an audio packet

You should be able to see if you uncomment the log statement in demuxread that the samples still come in chronological order. Even though an audio packet is the next one that we need to send, if the video stream is waiting for async task to finish then we don't send this audio until the task is done

In VP there are different scenarions affected when realtime is true but in general it will start playback well before filling the buffer. Different actions are taken as well depending on if audio buffer is low or video buffer is low - one is perhaps more important than the other, I wonder if audio continuity is perceived as more important than video.

> DemuxRead is called from Kodi many times very quickly
and with async we are creating hundreds of new threads every sec
creating new threads do an overhead in the CPU it is possible that this on low end devices is not so good


Maybe so, I'm no expert either. @phunkyfish, would you have any thoughts?
Quick run down: only one std::async is running at a time per stream opened (video/audio/sub), the whole point being to allow demuxread to return a packet (empty or not) straight away instead of getting stuck waiting for a download to finish when IA buffer has run dry.

Some analysis related to threads created by std::async here https://stackoverflow.com/questions/26730287/c11-thread-vs-async-performance-vs2013 seems to suggest that it may use a sort of thread pool under the hood and in our case only a handful of threads would be created, and checking by printing to the log in `ReadSample` I can only see 4 or 5 different thread ids after a couple of minutes.

CastagnaIT:
i tried do a search because i was confused
some comments found on the web said that async for some platform create new threads and with others said that it use pools

now i found this comment: https://over.wiki/ask/is-it-worth-replacing-the-stdthread-approach-with-stdasync/
(below) that seems clarify better this ambiguity, old gcc implementation create new threads the new gcc use pools, this in 2017 year

then for me its ok

CastagnaIT (approved):
ok for me,
i dont know if you want try give a bit more time to @phunkyfish to see if he want left a comment

src/Session.cpp
```diff
@@ -1449,8 +1464,12 @@ bool CSession::SeekChapter(int ch)
     m_adaptiveTree->next_period_ = m_adaptiveTree->periods_[ch];
     for (auto& stream : m_streams)
     {
-      if (stream->GetReader())
-        stream->GetReader()->Reset(true);
+      ISampleReader* sr{stream->GetReader()};
```
phunkyfish, line 1467: Better to use `ISampleReader* sr = stream->GetReader();` No need to a different assignment operator that is not often used.

src/Session.cpp
```diff
@@ -1102,37 +1099,55 @@ ISampleReader* CSession::GetNextSample()
...
-    return res->GetReader();
+    return true;
   }
-  else if (waiting)
+  else if (res)
   {
-    std::this_thread::sleep_for(std::chrono::milliseconds(100));
-    return m_dummySampleReader.get();
+    CheckFragmentDuration(*res);
+    ISampleReader* sr{res->GetReader()};
```
phunkyfish, line 1140: Better to use `ISampleReader* sr = stream->GetReader();` No need to a different assignment operator that is not often used.

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
     m_failedSeekTime = ~0;
   }
 
-  ISampleReader* sr(m_session->GetNextSample());
+  ISampleReader* sr{nullptr};
```
phunkyfish, line 431: Use assignment operator `=` instead.

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
-      p->pts = static_cast<double>(sr->PTS() + m_session->GetChapterStartTime());
-      p->duration = static_cast<double>(sr->GetDuration());
-      p->iStreamId = sr->GetStreamId();
-      p->iGroupId = 0;
-      p->iSize = iSize;
-      memcpy(p->pData, pData, iSize);
-    }
+      AP4_Size iSize(sr->GetSampleDataSize());
+      const AP4_UI08* pData(sr->GetSampleData());
+      bool srHaveData{ iSize > 0 && pData };
```
phunkyfish: Use assignment operator `=` instead.

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
+
+      if (sr->IsEncrypted() && srHaveData)
+      {
+        const unsigned int numSubSamples(*(reinterpret_cast<const unsigned int*>(pData)));
+        pData += sizeof(numSubSamples);
+        p = AllocateEncryptedDemuxPacket(iSize, numSubSamples);
+        memcpy(p->cryptoInfo->clearBytes, pData, numSubSamples * sizeof(uint16_t));
```
phunkyfish: Always qualify standard functions: `std::memcpy`

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
+        pData += (numSubSamples * sizeof(uint16_t));
+        memcpy(p->cryptoInfo->cipherBytes, pData, numSubSamples * sizeof(uint32_t));
```
phunkyfish: Always qualify standard functions: `std::memcpy`

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
+        pData += (numSubSamples * sizeof(uint32_t));
+        memcpy(p->cryptoInfo->iv, pData, 16);
```
phunkyfish: Always qualify standard functions: `std::memcpy`

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
+        pData += 16;
+        memcpy(p->cryptoInfo->kid, pData, 16);
```
phunkyfish: Always qualify standard functions: `std::memcpy`

src/main.cpp
```diff
@@ -429,60 +428,71 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
+
+      if (srHaveData)
+      {
+        p->dts = static_cast<double>(sr->DTS() + m_session->GetChapterStartTime());
+        p->pts = static_cast<double>(sr->PTS() + m_session->GetChapterStartTime());
+        p->duration = static_cast<double>(sr->GetDuration());
+        p->iStreamId = sr->GetStreamId();
+        p->iGroupId = 0;
+        p->iSize = iSize;
+        memcpy(p->pData, pData, iSize);
```
phunkyfish: Always qualify standard functions: `std::memcpy`

phunkyfish (changes requested):
Nice work, this is exactly how VP expects to be communicated with. Maybe also clean up other unusual edge cases.

Just nits really, but they may point to some smells elsewhere in the code based.

src/Session.cpp
CastagnaIT: i started use brackets `{}` since enen has required it, as per xbmc rules documentation
if there are different use cases it would be better to explain more in the documentation

CastagnaIT: I am happy to abandon `{}` but we need to clarify its uses so as not to create diatribes between developers

phunkyfish: Fair point. Let me check the docs. They may have changed.

phunkyfish: Ha. Well you learn something new every day. This comment is just noise please ignore.

glennguy:
Thanks for review @phunkyfish  - changes made (memcpy -> std::memcpy) if I can get your approval?

phunkyfish (approved):

PR #1010, opened 2022-06-22
Don't do PKCS7 stripping on chunked transfers

glennguy:
As per discussion at https://github.com/xbmc/inputstream.adaptive/pull/998#issuecomment-1162779734

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -228,6 +228,9 @@ bool AdaptiveStream::download(const std::string& url,
     std::vector<char> bufferData(bufferSize);
     ssize_t totalReadBytes{0};
     bool isEOF{false};
+    std::string transferEncodingStr{
+        file.GetPropertyValue(ADDON_FILE_PROPERTY_RESPONSE_HEADER, "Transfer-Encoding")};
+    bool isChunked{transferEncodingStr.find("hunked") != std::string::npos};
 
     while (!isEOF)
     {
@@ -245,7 +248,7 @@ bool AdaptiveStream::download(const std::string& url,
       else
       {
         // Store the data
-        if (write_data(bufferData.data(), byteRead, lockfreeBuffer, file.AtEnd()))
+        if (write_data(bufferData.data(), byteRead, lockfreeBuffer, (!isChunked && file.AtEnd())))
         {
           totalReadBytes += byteRead;
         }


L2501:
good for me

glennguy:
Thanks @L2501!

glennguy:
Hi @CastagnaIT would you be able to review this change when you have a chance? I'd still like to wait for @TermeHansen to validate there's no regression before merge but also be ready to bring in ASAP (mainly for backport)

src/common/AdaptiveStream.cpp
```diff
@@ -228,6 +228,10 @@ bool AdaptiveStream::download(const std::string& url,
     std::vector<char> bufferData(bufferSize);
     ssize_t totalReadBytes{0};
     bool isEOF{false};
+    std::string transferEncodingStr{
+        file.GetPropertyValue(ADDON_FILE_PROPERTY_RESPONSE_HEADER, "Transfer-Encoding")};
+    bool isChunked{!transferEncodingStr.empty() &&
+                   transferEncodingStr.find("hunked") != std::string::npos};
```
CastagnaIT:
```suggestion
    std::string transferEncodingValue{
        file.GetPropertyValue(ADDON_FILE_PROPERTY_RESPONSE_HEADER, "Transfer-Encoding")};
    bool isChunked{transferEncodingValue.find("chunked") != std::string::npos};
```

CastagnaIT: is there a reason why you have kept `hunked` ? in documentation is specified `chunked`
for `transferEncodingStr` from rules we have to avoid specify the type of variable on the var name, thats why i change it

glennguy: It captures both `Chunked` and `chunked` :)

TermeHansen:
Just tested backport build 19.0.6 on Android and it still works(tm)

glennguy:
Thanks @TermeHansen!

src/common/AdaptiveStream.cpp
glennguy: @CastagnaIT to give more detail - I have seen it done this way for other implementations of libcurl. 
If you prefer we can do `bool isChunked{transferEncodingValue.find("chunked") != std::string::npos || transferEncodingValue.find("Chunked") != std::string::npos};` but I figure that the one test for "hunked" is a little less verbose

CastagnaIT: its ok, yesterday there was internet down in italy many countries have been affected, after i forgot to answer..

CastagnaIT (approved):

Lantash77:
#1019 
Looks like my problems started since release 19.0.6

PR #1017, opened 2022-07-03
Switched bento4 sources to xbmc fork

CastagnaIT:
Switched bento4 sources to xbmc fork
the sources are the same with same patches,
plus add a couple of new CVE's patches


glennguy (approved):
Great!

basilgello:
Meh… And now we landed in the same situation as before.

Suppose I want to add a bunch of new security fixes from bento4 master. How do I PR them?
Should I set target branch to 'release/1.0.639-kodi' in xbmc/bento4 ?
And after this PR gets merged, we need to tag new release tarball (which semantic versioning has to be greater than current, i.e `1.0.639-20220806-Nexus`) and update dependency in inputstream.adaptive?

Looks like much more work than just adding a bunch of patches on top of upstream library :(.

What is the benefit of forking bento4 @CastagnaIT ?

CastagnaIT:
meh yes some more steps but also clearer to maintain, pro are:
no need to maintains tons of single patches files
modify/rebase existing commits its more easy without become crazy with each single file patch
its possible browse commits easily
check diff changes with a future old ISA version its more easy

so we act in similar way of other kodi repos, this is how
the current release name is `1.6.0-639-Nexus` on `v1.6.0-639-Nexus` branch

if there is no new official bento4 version:
then we use same branch to push new commits and to make a new release,
in this case the new release will have the name of `1.6.0-639-Nexus-2`, in next future become `1.6.0-639-Nexus-3`, `1.6.0-639-Nexus-4` etc...

if there is a new official bento4 version:
then we have to create a new branch, for example `v1.6.0-640-Nexus` and rebase all our commits on top of upstream,
the new release will have the name of `1.6.0-640-Nexus`

in case of PR, who do the PR merge (repo admins), will have to select `Rebase and merge` instead of `Create a merge commit`
so to avoid create the not needed merge commit (and avoiding to point the new release to the merge commit)

on ISA repo OFC is needed make a commit for the bento release version bump

basilgello:
I'd suggest using `1.6.0-639+20220806-Nexus` or `1.6.0-639+1-Nexus` for releases rather than using `Nexus-2`. Generally, Kodi release tarballs end with release cidename (Nexus) and most  Linux distro watcher tools have this rule in their config files. Any variant:

* `1.6.0-639-20220806-Nexus`,
* `1.6.0-639+20220806-Nexus`, 
* `1.6.0-639-1-Nexus`,
* `1.6.0-639+1-Nexus`

will work.

If you want I can try crafting a Github Action script which will tag the update and PR the updated links to IA.

CastagnaIT:
yeah i like better `1.6.0-639+1-Nexus` its more immediate to read

> If you want I can try crafting a Github Action script which will tag the update and PR the updated links to IA

im not sure if its a good idea automatize things if we need to do preventive tests
perhaps it is better to do it manually for now, but good to know
ref. @glennguy

PR #1027, opened 2022-07-30
Further fixes relating to subtitle playback

glennguy:
Please see the comments left throughout the code changes

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -1006,10 +1006,6 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
   if (!current_rep_)
     return false;
 
-  if (state_ == STOPPED)
-    // For subtitles which come in one file we should return true!
-    return current_rep_->segments_.empty();
-
   if (current_rep_->flags_ & AdaptiveTree::Representation::SUBTITLESTREAM)
     return true;
 
@@ -1055,7 +1051,7 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
   if (newSeg)
   {
     needReset = true;
-    if (newSeg != old_seg)
+    if ((newSeg != old_seg) || (!preceeding && state_ == STOPPED))
     {
       StopWorker(STOPPED);
       // EnsureSegment loads always the next segment, so go back 1
@@ -1063,14 +1059,16 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
           current_rep_->get_segment(current_rep_->get_segment_pos(newSeg) - 1);
       // TODO: if new segment is already prefetched, don't ResetActiveBuffer;
       ResetActiveBuffer(false);
+      if (newSeg == old_seg && !preceeding)
+      {
+        absolute_position_ -= segment_read_pos_;
+        segment_read_pos_ = 0;
+      }
     }
-    else if (!preceeding)
+    else if (preceeding)
     {
-      absolute_position_ -= segment_read_pos_;
-      segment_read_pos_ = 0;
-    }
-    else
       needReset = false;
+    }
     return true;
   }
   else
--- a/src/parser/HLSTree.cpp
+++ b/src/parser/HLSTree.cpp
@@ -619,6 +619,7 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
           FreeSegments(period, rep);
           rep->segments_.swap(newSegments);
           rep->startNumber_ = newStartNumber;
+          rep->duration_ = period->duration_;
 
           if (segmentInitialization)
           {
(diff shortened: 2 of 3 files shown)


src/common/AdaptiveStream.cpp
```diff
@@ -1006,10 +1006,6 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
   if (!current_rep_)
     return false;
 
-  if (state_ == STOPPED)
-    // For subtitles which come in one file we should return true!
-    return current_rep_->segments_.empty();
```
glennguy, line 1011: Works fine for single file but our state becomes 'STOPPED' once we're in the last segment of a multi-segment period. Since subtitle stream segments are quite long (several minutes usually) the effect of the code here was seeking once in this segment would never be able to recover the sub stream - subs are now off until we stop/start playback or for multiperiod you would have to exit/re-enter the period.
For audio/video streams, trying to seek once already in the last segment would stop playback.
Removing this allows for the appropriate things to happen to 'restart' the stopped stream (other changes help address this also)

src/common/AdaptiveStream.cpp
```diff
@@ -1055,22 +1051,24 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
   if (newSeg)
   {
     needReset = true;
-    if (newSeg != old_seg)
+    if ((newSeg != old_seg) || (!preceeding && state_ == STOPPED))
```
glennguy, line 1054: Condition checks of this entire section rewritten to avoid having some parts of the blocks copy/pasted. Please check to see I haven't stuffed it up :)

src/common/AdaptiveStream.cpp
```diff
@@ -1055,22 +1051,24 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
...
     }
-    else if (!preceeding)
+    else if (preceeding)
     {
-      absolute_position_ -= segment_read_pos_;
-      segment_read_pos_ = 0;
-    }
-    else
       needReset = false;
+    }
```
glennguy, line 1071: Here the goal is to make sure that if the state is stopped we start it again. Not needed to check state when preceeding is only true when seeking across periods (not sure why this is btw)

On another note, `StopWorker` is horribly named

src/parser/HLSTree.cpp
```diff
@@ -619,6 +619,7 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
           FreeSegments(period, rep);
           rep->segments_.swap(newSegments);
           rep->startNumber_ = newStartNumber;
+          rep->duration_ = period->duration_;
```
glennguy, line 622: Previously this was only being set on the last period (an oversight of HLS multi period implementation) and causing issues with the seeking logic here: https://github.com/xbmc/inputstream.adaptive/blob/74205c0cf1625f016a21d1b924cf26af66e2a8ff/src/common/AdaptiveStream.cpp#L1041

src/samplereader/SubtitleSampleReader.cpp
```diff
@@ -153,7 +153,10 @@ AP4_Result CSubtitleSampleReader::ReadSample()
 void CSubtitleSampleReader::Reset(bool bEOS)
 {
   if (m_adByteStream || bEOS)
+  {
+    m_eos = bEOS;
```
glennguy, line 157: `m_eos` not being set was causing the sub reader to be overlooked in `GetNextSample`

src/parser/HLSTree.cpp
glennguy: This also exposes an issue that I believe is in VP - seeking in to the very last moments of a period (not the last one) causes a big sync issue. In a nutshell: with a 2 period video, 10 minutes long and period 2 starts at 6 minutes, if I seek to 5:59 the OSD timer goes back to zero with a black screen, and I will be waiting for 6 minutes for it to 'catch up' and playback of period 2 to begin. Will log an issue for this once I find an easy to reproduce internet stream

src/common/AdaptiveStream.cpp
CastagnaIT: I do not have enough knowledge to verify this code, but i have done various tests and not found side effects,
we will see if there are regressions reported by users

src/common/AdaptiveStream.cpp
CastagnaIT: yeah the name is inappropriate

CastagnaIT (approved):
i have done tests with all my test streams that i have with subs, and all works good! great works!

just one stream not works good, but is due to a bug in Kodi, see PR: https://github.com/xbmc/xbmc/pull/21727
if you can please make some tests with my kodi PR and your fixes, as confirmation

glennguy:
@CastagnaIT sorry for such a delay in responding, been very busy with work recently.

I've tested with https://github.com/xbmc/xbmc/pull/21727 and all looks good. Thanks for the review!

PR #1033, opened 2022-08-20
Fix use after free crash from readsample thread

### LINKED ISSUE #1028: Use-after-free crash caused by lack of synchronization between CStream::Reset() and IStreamReader::ReadSampleAsync
### Describe the bug

Recently refactored Session class introduces a regression when the `OpenStream` first resets `CTSSampleReader` and then `DemuxRead` in another thread tries to access the freed `m_AVContext` and crashes Kodi with SIGSEGV.

## Expected Behavior

No crash should be observed.

## Actual Behavior

The `TSReader::~TSReader()` destructor is called here:

```gdb
Thread 25 "VideoPlayer" hit Breakpoint 1, TSReader::~TSReader (this=0x7fcd241719f8, __in_chrg=<optimized out>) at ./src/TSReader.cpp:33
33	./src/TSReader.cpp: Немає такого файла або каталогу.
#0  TSReader::~TSReader (this=0x7fcd241719f8, __in_chrg=<optimized out>) at ./src/TSReader.cpp:33
No locals.
#1  0x00007fcce14e7066 in CTSSampleReader::~CTSSampleReader (this=0x7fcd241719e0, __in_chrg=<optimized out>) at ./src/samplereader/TSSampleReader.h:12
No locals.
#2  CTSSampleReader::~CTSSampleReader (this=0x7fcd241719e0, __in_chrg=<optimized out>) at ./src/samplereader/TSSampleReader.h:12
No locals.
#3  0x00007fcce14f7f61 in std::default_delete<ISampleReader>::operator() (__ptr=<optimized out>, this=<optimized out>) at /usr/include/c++/12/bits/unique_ptr.h:89
No locals.
#4  std::__uniq_ptr_impl<ISampleReader, std::default_delete<ISampleReader> >::reset (__p=0x0, this=0x7fcd24886ea0) at /usr/include/c++/12/bits/unique_ptr.h:203
        __old_p = <optimized out>
#5  std::unique_ptr<ISampleReader, std::default_delete<ISampleReader> >::reset (__p=0x0, this=0x7fcd24886ea0) at /usr/include/c++/12/bits/unique_ptr.h:501
No locals.
#6  SESSION::CStream::Reset (this=this@entry=0x7fcd24886cb0) at ./src/Stream.cpp:30
No locals.
#7  0x00007fcce14f809a in SESSION::CStream::Reset (this=this@entry=0x7fcd24886cb0) at ./src/Stream.cpp:28
No locals.
#8  0x00007fcce14a608b in CInputStreamAdaptive::OpenStream (this=0x7fcd2425ff10, streamid=1001) at ./src/main.cpp:255
        stream = 0x7fcd24886cb0
        needRefetch = false
        rep = <optimized out>
        __FUNCTION__ = "OpenStream"
        movie = <optimized out>
#9  0x000055fce6dc7d02 in CDVDDemuxClient::OpenStream(int) ()
No symbol table info available.
#10 0x000055fce6e4fff6 in CVideoPlayer::OpenStream(CCurrentStream&, long, int, int, bool) ()
No symbol table info available.
#11 0x000055fce6e514e2 in CVideoPlayer::OpenDefaultStreams(bool) ()
No symbol table info available.
#12 0x000055fce6e52345 in CVideoPlayer::ReadPacket(DemuxPacket*&, CDemuxStream*&) ()
No symbol table info available.
#13 0x000055fce6e5938d in CVideoPlayer::Process() ()
No symbol table info available.
#14 0x000055fce70f2005 in CThread::Action() ()
No symbol table info available.
#15 0x000055fce70f3b4d in ?? ()
No symbol table info available.
#16 0x000055fce70f494b in ?? ()
No symbol table info available.
#17 0x00007fcd408d49c3 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#18 0x00007fcd44b14d80 in start_thread (arg=0x7fccd91fa640) at pthread_create.c:481
        ret = <optimized out>
        pd = 0x7fccd91fa640
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {140517792785984, -3906111645919552856, 140729902391870, 140729902391871, 0, 140517792785984, 3914479203012237992, 3913695446271974056}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = 0
#19 0x00007fcd40b31baf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
No locals.
```
while the "main" stream keeps playing and the next `ReadSampleAsync` crashes in `CTSSampleReader::ReadPacket()` because the `m_AVContext` member is already destroyed.

## Possible Fix

As a very first bandaid solving the crash **for me**, I added the following code block to https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/TSReader.cpp#L187 :

```c++
if (!m_AVContext) return false;
```

However, I see more proper to use `stream->Disable()` instead of `stream->Reset()` here
https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/main.cpp#L255

Or even better, just guard `TSReader` class methods so they block on each enter.

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [x] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Debian 12 bookworm/testing
 - Kodi version: 20.0 alpha2
 - inputstream.adaptive version: 20.2.3

glennguy:
fixes #1028 (see for full explanation) - thanks @basilgello 

Decided to go with altering the `Reset` function to wait for the async readsample thread to finish. Changing the call in main.cpp from `stream->Reset()` to `stream->Disable()` although solving the problem seems the wrong way to do it (at the very least from how the functions are named) - we only want to reset the stream, not fully disable.

@basilgello are you able to verify this fixes the issue for you? (cannot reproduce here)

### DIFF
--- a/src/Stream.cpp
+++ b/src/Stream.cpp
@@ -27,6 +27,8 @@ void CStream::Reset()
 {
   if (m_isEnabled)
   {
+    if (m_streamReader)
+      m_streamReader->WaitReadSampleAsyncComplete();
     m_streamReader.reset();
     m_streamFile.reset();
     m_adByteStream.reset();


CastagnaIT (approved):
IMO has sense, in any case better wait the test feedback

basilgello:
Yes, can check today against tbe streams and report back!

basilgello (approved):
I tested the PR on a dozen streams (HD and non-HD) including channels known to cause issues. All streams work fine for 20+ minutes.

basilgello:
As a side note: shouldnt we later implement a ring buffer to get rid of `std::async` thread spamming? I.e one thread downloads and decrypts segments and pkaces them to a period buffer while DemuxRead from the main thread reads them (lock-free or read-lock) and marks as read (under the lock).

glennguy:
Edit - thanks for testing @basilgello !

I was worried about the overhead when doing the `ReadSample` on a separate thread, had a look into `std::async` and what I was told that it is somewhat 'smart' about reusing threads. Testing on Windows on a video/audio/sub stream it re-used the same 5 or 6 threads the whole time.

The whole download/write/read threading in `AdaptiveStream` needs to be overhauled IMO. Seems to be fine in normal playback but the unit tests can cause a deadlock (without hack) which doesn't seem right: https://github.com/xbmc/inputstream.adaptive/blob/1b8ca82879e1edba885a03b54a6985eff9a76981/src/test/TestDASHTree.cpp#L95-L102
There are locks which are acquired inside another lock, but sometimes by themselves. I tried and failed to fix this :/

basilgello:
@phunkyfish summoning you as you are much better at C++ multithreading!

basilgello:
On Linux, threads are spawned like from a machine gun, so let's implement these properly

glennguy:
@basilgello do you have a log showing this?
On linux I tested with 

```
https://bitmovin-a.akamaihd.net/content/MI201109210084_1/mpds/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.mpd
```
separate audio/video streams and only 2 different thread ids for downloading.

[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/9439617/kodi.log.txt)

basilgello:
I tested with the example I gave you on Slack @glennguy (the hls manifest_type). It still worke.

CastagnaIT:
i think this for now can be merged,
to make the threading rework is need to change a lot of things
we can discuss on a separate Issue or on a Slack group without spam this PR

PR #1035, opened 2022-08-31
Fix video resume with seek

CastagnaIT:
fix the broken video resume after PR #1027
so i have based my changes on the previous implementation
i think is fine but @glennguy better if you can double check

tested also webvtt regressions with:
our akamai test stream for webvtt segmented chapters
dash ISO/MP4 webvtt

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -1057,7 +1057,7 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
   if (newSeg)
   {
     needReset = true;
-    if ((newSeg != old_seg) || (!preceeding && state_ == STOPPED))
+    if ((old_seg && newSeg != old_seg) || (!preceeding && state_ == STOPPED))
     {
       StopWorker(STOPPED);
       // EnsureSegment loads always the next segment, so go back 1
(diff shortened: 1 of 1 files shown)


CastagnaIT:
just found a problem with akamai test stream
if you play the video without do seeks between chapters (normal playback) when change chapter the subtitles become broken

EDIT:
also "Nexus" branch has this bug... I thought it worked a while ago

so subtitles packet arrives to kodi but:
- with normal playback when chapter change the "m_chapterStartTime" is set and subtitles are not displayed (something like wrong offset timing?)
- if you seek middle of another chapter the "m_chapterStartTime" is not set (value 0), and subtitles are displayed

i am pretty confused...

glennguy:
Hmm will need to look at a bit more
`m_chapterStartTime` in Session seems to be set on start of period but never used.

glennguy:
Yes I can see now that I've made an oversight in how I reorganised this section.
There was nothing to handle the case where `newSeg == old_seg && !preceeding && state_ != STOPPED`

I was trying to avoid having duplicate blocks of 
```
StopWorker(STOPPED);
      // EnsureSegment loads always the next segment, so go back 1
      current_rep_->current_segment_ =
          current_rep_->get_segment(current_rep_->get_segment_pos(newSeg) - 1);
      // TODO: if new segment is already prefetched, don't ResetActiveBuffer;
      ResetActiveBuffer(false);
```
But maybe it would be better to set a bool and execute that after the if block?
I'll try to reproduce the bug when playing through 1 chapter to the next and see if I can work it out.

CastagnaIT:
if you want reuse same block of code yes there are not many alternatives
free to superseed this PR

glennguy:
Yes it's confusing (regarding 'playing through' chapters)
Maybe there's something on the Kodi side that isn't being initialized/reset. If I hardcode the offset value in IA to always send 0, then playback is fine when playing through (and broken when seeking).
When playing through, at the point the first packet of the second chapter is sent IA is streaming the second chapter, current_period_ is second chapter etc., but VP is still presenting the first chapter. Haven't looked at Kodi side yet but this could be the key...

glennguy:
Merging this now to at least fix the resume issue so release can be made

Issue #1007, opened 2022-06-16
CMAF Support


matthuisman:
This is an issue to track CMAF support in Inputstream Adaptive / Kodi

CMAF is basically fMP4 (fragmented mp4) + M3U8/Dash manifest + optional encryption (usually AES-CBC)
It allows a single set of media segments for many benefits (better caching etc)

Current Issues

1) 
CBCS support
Added in Inputstream 20 with PR: https://github.com/xbmc/inputstream.adaptive/pull/887
Unable to backport to 19 due to Kodi core needing changes for Android support
Technically, Kodi 19 support for other platforms could be possible - but not great having a feature that doesn't work on a certain platform (especially a big one like Android)
    
2)
Support for ignoring unsupported EXT-X-KEY in M3U8
The below will error out with LOGERROR, "Unsupported encryption method: SAMPLE-AES"
Due to the first EXT-X-KEY being for apple streaming (actual sample-aes)
The 2nd Sample-aes is for CMAF CBCS (Widevine)
If I remove the first line - we support the 2nd line already due to looking for the KEYFORMAT.

So, we just need to change the code to not fail on SAMPLE-AES, but to just ignore it.
I think i've had a PR about that in the past.
```
#EXTM3U
#EXT-X-TARGETDURATION:5
#EXT-X-VERSION:6
#EXT-X-DISCONTINUITY-SEQUENCE:0
#EXT-X-MEDIA-SEQUENCE:5
#EXT-X-INDEPENDENT-SEGMENTS
#EXT-X-START:TIME-OFFSET=-15,PRECISE=YES
#EXT-X-PROGRAM-DATE-TIME:2022-06-14T10:42:30.211Z
#EXT-X-KEY:METHOD=SAMPLE-AES,URI="skd://03b47c8e-5bdf-41cb-bbef-29be6a18de9e:813A1D12BE0A6D2475F810BEE5AFC5E2",IV=0x813A1D12BE0A6D2475F810BEE5AFC5E2,KEYFORMAT="com.apple.streamingkeydelivery",KEYFORMATVERSIONS="1"
[...]

PR #1048, opened 2022-10-01
CBC refactor, HLS fixes

glennguy:
A few things in this change:

* Byteranged HLS playlists - some playlists don't show the start byte in the byterange tag, this commit allows for this type to parse correctly
* `CryptoMode` enum is moved into it's own file. Since these values are used in multiple classes it makes sense for it to live on its own
* Refactoring of decrypt-only part of the `DecryptSampleData` function:

For some time now it seems that the CDM will not accept decrypting samples with multiple subsamples. It works fine when decrypting and decoding, but with decrypt only it returns an error. Maybe this is an oversight by google/wv because browser may always do decrypt/decode by default? Hard to know.

The existing code tests first to see what the CDM will do and if it fails to decrypt 2 subsamples it attempts to decrypt with 1. If this is successful then a flag is set to first combine the subsample data and decrypt as 1 contiguous sample. Now this works fine for CTR cipher, however with CBC the output of one decryption block is XORed with the previous block - therefore we end up with incorrect output as soon as we move on from the first subsample.

The change here adapts to the cypher type and allows for proper CBCS decryption when there are multiple subsamples.


glennguy:
Fixes #1007 in conjunction with https://github.com/glennguy/Bento4-axiomatic/commit/96c9abfc603be06310cb9902d87f1d12959b46ab - @matthuisman I'll send you a release build on slack

I'll fix the Android builds shortly

matthuisman:
Amazing work. For now, could you include a bento4 patch in this PR so we could use Jenkins builds? Or does IA no longer support patches? It could always be removed once upstream merged?

glennguy:
I think it would be preferable to just update xbmc/bento4. I know the patch files have caused a bunch of issues with other automated builds outside of here so I'll prepare a PR for this as well.

In the mean time though here's a release build for Win for you to test: 
edit: please see next post for correct build

glennguy:
[addon-inputstream.adaptive-20.3.0-windows-x86_64.zip](https://github.com/xbmc/inputstream.adaptive/files/9689598/addon-inputstream.adaptive-20.3.0-windows-x86_64.zip)

CastagnaIT:
i have no time today to review it just a note
if you want add new changes to xbmc/bento4 since we are admins better if you push commit directly to the branch without making PR's

or you want make a PR to explaining more things outside of commit description,
then after use "Rebase and merge" instead of "Merge pull request"

this will make easier future rebases to new bento upstream version

matthuisman:
works really well :) thank you for the hard work!

src/common/AdaptiveDecrypter.h
```diff
@@ -8,19 +8,15 @@
...
-#include <bento4/Ap4.h>
+#include "../CryptoMode.h"
 
 #include <stdexcept>
 #include <string_view>
 
+#include <bento4/Ap4.h>
+
 // These values must match their respective constant values
 // defined in the Android MediaCodec class
```
CastagnaIT: this comment must be moved with to the enum

src/common/AdaptiveTree.h
```diff
@@ -505,7 +507,10 @@ class ATTR_DLL_LOCAL AdaptiveTree
   bool adp_timelined_, period_timelined_;
 
   bool current_hasRepURN_, current_hasAdpURN_;
-  std::string current_pssh_, current_defaultKID_, current_iv_;
+  std::string current_pssh_;
+  std::string current_defaultKID_;
+  std::string current_iv_;
+  CryptoMode m_currentCryptoMode;
```
CastagnaIT: i know that it was existing code but i have not understand because there is a prefix with "current" when there is a fixed value, i am wrong?
in case i would prefer to drop this prefix

src/parser/HLSTree.cpp
```diff
@@ -496,10 +502,13 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
         {
           std::string::size_type bs = line.rfind('@');
           if (bs != std::string::npos)
-          {
             segment.range_begin_ = atoll(line.c_str() + (bs + 1));
```
CastagnaIT:
```suggestion
            segment.range_begin_ = std::atoll(line.c_str() + (bs + 1));
```

src/parser/HLSTree.cpp
```diff
@@ -496,10 +502,13 @@ HLSTree::PREPARE_RESULT HLSTree::prepareRepresentation(Period* period,
...
-            segment.range_end_ = segment.range_begin_ + atoll(line.c_str() + 17) - 1;
-          }
+          else
+            segment.range_begin_ = newSegments.size() > 0
+                                       ? newSegments.Get(newSegments.size() - 1)->range_end_ + 1
+                                       : 0;
+
+          segment.range_end_ = segment.range_begin_ + atoll(line.c_str() + 17) - 1;
```
CastagnaIT:
```suggestion
          segment.range_end_ = segment.range_begin_ + std::atoll(line.c_str() + 17) - 1;
```

wvdecrypter/wvdecrypter.cpp
```diff
@@ -1166,99 +1253,135 @@ AP4_Result WV_CencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 pool_id,
...
+  // from it before passing to CDM.
+  // CBCS:
+  // Due to the nature of this cipher subsamples must be decrypted separately
+  else
   {
-    char buf[36]; buf[32] = 0;
-    AP4_FormatHex(fragInfo.key_, 16, buf);
-    LOG::LogF(SSDDEBUG, "Decrypt failed with error: %d and key: %s", ret, buf);
-  }
+    const size_t iterations{useCbcDecrypt ? subsample_count : 1};
```
CastagnaIT: has been discussed recently on slack that for better reading is some cases we can still use `=` instead of brackets (yes docs should be updated....)
```suggestion
    const size_t iterations = useCbcDecrypt ? subsample_count : 1;
```

CastagnaIT: since subsample_count is `unsigned int` and there is no places where `size_t` is used i think its better change all size_t vars with same type of subsample_count  or better `uint32_t`
this also on all new pack methods

wvdecrypter/wvdecrypter.cpp
```diff
@@ -1166,99 +1253,135 @@ AP4_Result WV_CencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 pool_id,
...
+      decrypt_in_.SetDataSize(0);
+      size_t decryptInPos = absPos;
+      if (useCbcDecrypt)
+      {
+        UnpackSubsampleData(data_in, decryptInPos, i, bytes_of_cleartext_data,
+                            bytes_of_encrypted_data);
+      }
+      else
+      {
+        for (unsigned int j{0}; j < subsample_count; ++j)
```
CastagnaIT: Please rename `j` with a more explicit var name e.g. `subIndex` or similar
I would like to avoid easy oversights between i and j

CastagnaIT (commented):
only minor things,
please note that in the wvdecrypter.cpp has been introduced a CRLF that need to be removed

src/common/AdaptiveTree.h
glennguy: Sure - I'll change this one.
For the others - example being for AES-128/HLS. PSSH is inserted and the function uses the values of the 'current' IV/KID etc to add to the vector of PSSHs.

wvdecrypter/wvdecrypter.cpp
glennguy: Since Bento4 uses unsigned int I'll switch to that

glennguy:
Ok @CastagnaIT comments are addressed

wvdecrypter/wvdecrypter.cpp
```diff
@@ -234,7 +234,26 @@ class WV_CencSingleSampleDecrypter : public Adaptive_CencSingleSampleDecrypter
...
+  void RepackSubsampleData(AP4_DataBuffer& dataIn,
+                           AP4_DataBuffer& dataOut,
+                           size_t& startPos,
+                           size_t& cipherPos,
+                           const unsigned int subsamplePos,
+                           const AP4_UI16* bytesOfCleartextData,
+                           const AP4_UI32* bytesOfEncryptedData);
+  void UnpackSubsampleData(AP4_DataBuffer& data_in,
+                           size_t& startPos,
+                           const size_t subsamplePos,
```
CastagnaIT:
```suggestion
                           const unsigned int subsamplePos,
```

wvdecrypter/wvdecrypter.cpp
```diff
@@ -1166,97 +1253,133 @@ AP4_Result WV_CencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 pool_id,
     bytes_of_cleartext_data = &clearb;
     bytes_of_encrypted_data = &cipherb;
   }
-
-  cdm::InputBuffer_2 cdm_in;
+
```
CastagnaIT: remove whitespaces

CastagnaIT (commented):
your branch need also to be rebased with current Nexus because show unwanted commits

glennguy:
Ok all done :)

glennguy:
Once approved I'll push changes to bento4 and make a release there.
Happy to do a release here once this is merged?

CastagnaIT:
> Once approved I'll push changes to bento4 and make a release there.
> Happy to do a release here once this is merged?

yes no problem just rename commit name like `[Bento4] Version bump 1.6.0-639-xyz-Nexus`
so can be better identified in history

CastagnaIT (approved):

matthuisman:
Awesome work!!!

Issue #1047, opened 2022-09-29
Nexus H265 issues on newer version of ISA


dev77-07:
### Describe the bug
Here is a clear and concise description of what the problem is:
Playing H265 media from the Hulu addon doesn't work on ISA 20.2.0-Nexus and higher. Tested with 20.1.2-Nexus and it works. 20.2.0 crashes when trying to play H265, 20.3.0 just spins and never loads, Kodi freezes if I try and exit.
Moving issue from https://github.com/matthuisman/slyguy.addons/issues/267 to here

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

Play H265 media without crashing or not loading

## Actual Behavior

Kodi crashes or doesn't load media

### To Reproduce

1. 
2.
3.

### Debuglog

The debuglog can be found here:
ISA 20.3.0: https://paste.kodi.tv/rorejuveke.kodi
ISA 20.2.0: https://paste.kodi.tv/jiwicijuxa.kodi
ISA 20.1.2: https://paste.kodi.tv/uhivalinax.kodi

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:

### Screenshots 
Here are some links or screenshots to help explain the problem:

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Operating system version/name:
 - Kodi version: 20.0-ALPHA3 (19.90.710) Git:20220927-b7332d015c
 - inputstream.adaptive version: 20.3.0, 20.2.0

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

CastagnaIT:
MPD of ISA 20.1.2 log: [171249737.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/9681941/171249737.mpd.txt)

Your kodi version is one of last builds, so since there is the API breaking change with these kodi versions
all ISA versions below 20.3.0 will not works
or you have self changed/builded source code to make these works but this will make only confusion to understand the causes

the log of "ISA 20.1.2" it is also weird is missing ISA stream selection logging appears like another oldest version

[...]

PR #1049, opened 2022-10-04
Fix junk being sent as kid to decrypter

glennguy:
If we have a pssh without a default KID (DASHTree doesn't parse properly due to order of tags):
```
<ContentProtection schemeIdUri="urn:mpeg:dash:mp4protection:2011" value="cenc" cenc:default_KID="2384f91217d34162a995187704117759"/>
<ContentProtection schemeIdUri="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed"    >
  <cenc:pssh>AAAASXBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAACkIARIQI4T5EhfTQWKplRh3BBF3WRoEaHVsdSIJMTcxMjQ5NzM3KgJIRA==</cenc:pssh>
</ContentProtection>
```
Then we end up sending a pointer to an empty string and crap gets sent to the decypter, then more bad stuff happens when `GetCapabilities` tries to use it.

Small clean-up included. With the full refactor of wvdecrypter we can abolish the `defkid` variable and just send the string instead.

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -599,9 +599,8 @@ bool CSession::InitializeDRM(bool addDefaultKID /* = false */)
 
       CCdmSession& session{m_cdmSessions[ses]};
       std::string defaultKid{sessionPsshset.defaultKID_};
-
-      const char* defkid{sessionPsshset.defaultKID_.empty() ? nullptr
-                                                            : sessionPsshset.defaultKID_.data()};
+      const uint8_t* defkid{
+          defaultKid.empty() ? nullptr : reinterpret_cast<const uint8_t*>(defaultKid.data())};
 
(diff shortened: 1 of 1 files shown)


glennguy:
Should fix #1047 

@matthuisman @yeahme49 could you try the builds generated from this PR?

matthuisman:
Interesting, has this always been the case? Do you know what broke it recently?

glennguy:
The dash thing has always been there, but the bad kid is introduced since the `Session` refactor (I think)

dev77-07:
Tested and works great, fixes my issue in 1047

CastagnaIT (approved):

PR #1062, opened 2022-10-29
Fix memory leak in AdaptiveTree::SortTree

dobo90:
AdaptationSet is erased from the container but it is not freed.


CastagnaIT (approved):
it think its ok thanks, backport needed?

glennguy:
Thanks @dobo90 !

PR #1067, opened 2022-11-05
Full refactor manifest parsers and adaptive tree structs

CastagnaIT:
## Description
This PR rework all manifest parsers with more effective and simpler XML data management
this is mandatory in order to fix various problems, that due to wrong old code implementations, things like manage DASH multi BaseURL's, is currently almost impossible to implement without add new bad hacks.

Related to this also adaptive tree structs will be reworked where each component will have an appropriate class with more organized code that will allow easier future reworks/cleanups

this will be a radical and enormous code change
since touches many project parts that are closely related so making separate commits will probably be impossible

Main todo:
- [x] Add new dependency ~tinyxml2~ pugixml (in a final future to full replace expat dependency)
- [x] Rework adaptive tree strucs by creating an each proper class implementation (e.g. structs as period, representation, adaptionset...) --- there is a lot of code that can be improved with separate PR's
- [x] Dash parser rework
- [x] Adapt various code parts of ISAdaptive core
- [x] Make some basic Dash streaming tests:
  - [x] played dash with SegmentTemplate having and not the SegmentTimeline
  - [x] played dash with SegmentList
  - [x] played dash with SegmentBase
  - [x] played dash widevine DRM (n€tflix) ~works but video seek is broken~
  - [x] played dash widevine DRM cbcs
  - [x] tested PlayReady header data parsing, we have a sample stream? 
  - [ ] ...other tests may be needed

To do after a first simple dash streaming test works
- [x] Dash parser rework, live update method, tests:
  - [x] played dash with fb live ~(broken as master branch, but starts)~ fixed
  - [x] played a couple of live samples, are not good handled, so there are same result of master branch
  - [ ]  ...other live tests may be needed
- [x] HLS parser rework
- [x] Make some basic HLS streaming tests:
  - [x] played HLS unencrypted VOD
  - [x] played HLS AES128 VOD
  - [x] played HLS AES128 with chapters
  - [x] played live HLS
  - [ ]  ...other live tests may be needed
- [x] Smooth stream parser rework
  - [x] played sample test with manifest UTF16 xml format
- [x] Re-adapt code of test project

This will remove many code hacks/workarounds like using adaptive tree class to store temporary data, flags, ...

I will try to maintain a similarity to the original code

due to the amount of work, specifics code refactor/rework will be done on separate PR's,
and no documentation will be added to the new methods (separate PR's),
the current goal for now is to make it all code working possibly with few or none regressions

Todos:
- ~2 dash tests that fails: `CalculateCorrectSegmentNumbersFromSegmentTimeline`, `replacePlaceHolders` i have commented the old test code to allow compare before/after, these two failure however might not be relevants, i have fixed them by add `SetNowTime`, i noticed that other live tests have it, but these two do not~ fixed
- I added many `\\! @todo:` in the code, while reworking the code i noticed weaknesses to be resolved and code that needs to be better managed, to be done in future
- Dash with playready drm streams its a particolar use case that i have not tested

## Note to particular code changes

CPeriod inherited from CCommonSegAttribs,
CAdaptationSet/CRepresentation are inherited from CCommonAttribs/CCommonSegAttribs,
that allow to share common data with parent levels

The rework highlighted customised DASH implementations, where some of them are not documented in wiki, and nor in the github history e.g. `SegmentDurations` tag, `codecPrivateData`

DASH multi BaseURL has been now fixed, so will be taken in account the first BaseURL defined, instead of the last one (OFC the full support to multi base url must be done in future)

HLS parser its the same of before just some cleanups/adaptations, has been fixed a possible crash when extract the file extension on segment url.

SmoothStream parser has been reworked with new XML implementation, no particular changes to mention.

`SPINCACHE` class has been moved in to AdaptiveUtils as `CSpinCache` and has been added two security fixes that can cause crashes for out-of-range or due to empty data.

## How has this been tested?
On Windows OS only, see list above where i played differents sample streams

Streaming add-ons tested:
- N€tflix: vod
- La7: live and vod
- Twitch: vod
- Youtube: vod

## Types of change

- [x] **None of the above** (please explain below)
code refactor

## Checklist:


CastagnaIT:
i think i have found a bug in the tinyxml2 library,
seem not able to parse correctly xmls with tags arguments in to multiple lines (also removing return chars seem multiple spaces cause problems)

maybe is worth switching to pugixml library (discovered now) that have also support for common unicodes, it should also allow to remove our utf conversion, methods in both libraries are very similar should not be too complex transition
i will have to do a test

EDIT: confirmed pugixml works as expected its a tinyxml2 bug, i will switch to pugixml
UPDATE: code switched to pugixml

src/test/TestDASHTree.cpp
```diff
@@ -141,89 +141,99 @@ TEST_F(DASHTreeTest, CalculateBaseUrlFromRedirect)
...
-  adaptive::SPINCACHE<adaptive::AdaptiveTree::Segment> segments =
-      tree->periods_[0]->adaptationSets_[0]->representations_[0]->segments_;
+  auto& segments =
+      tree->m_periods[0]->GetAdaptationSets()[0]->GetRepresentations()[0]->SegmentTimeline();
+
+  EXPECT_EQ(segments.GetSize(), 13);
+  EXPECT_EQ(segments.Get(0)->range_end_, 853663);
+  EXPECT_EQ(segments.Get(12)->range_end_, 853675);
+
+  /*! @todo: without SetNowTime this now return wrong range_end_ values, is this right?
```
glennguy: I think this might be wrong. For SegmentTimeline we definitely have to go by system clock but in this case it should be anchored to the start number.

Consider if my system clock is 2 minutes fast - the parser will have all of its segments in the future for this test manifest.

For easy reference, here is the template:
```
<SegmentTemplate initialization="$RepresentationID$/init.mp4?id=8c1bbb709229e110fbb0d1e544266504" 
  media="$RepresentationID$/segment_$Number$.m4s" 
  presentationTimeOffset="1738747575" 
  startNumber="487050"
  timescale="90000">
  <SegmentTimeline>
    <S d="540000" r="12" t="263007000000"/>
  </SegmentTimeline>
</SegmentTemplate>
```

glennguy:
Apart from above comment everything else looks great. I spent a couple of hours going over this, amazing work!!!

src/test/TestDASHTree.cpp
glennguy: Really the segments need to always match the template since we have no idea if the duration and t values will be the same for future segments, if they will even exist.

CastagnaIT: found the problem and fixed
now all tests pass ok

glennguy (approved):
Well done!

glennguy:
Please merge when you're ready

CastagnaIT:
it is fine with me if some regression is found later it will be addressed separately

Issue #1072, opened 2022-11-29
KODI crashes in ism file. (KODI 20)


mbebe:
It only appears on KODI 20 (prereleased and all nightly).
ism file: 
[ism.txt](https://github.com/xbmc/inputstream.adaptive/files/10116845/ism.txt)
url: https://ec04-szz1.waw5.cache.orange.pl/canal/v/canal/vod/store01/CAP_21042687/_/hd4-hssdrm02.ism/Manifest
 In KODI 19 base url is created correctly (without Manifest at the end) - 
`https://ec04-szz1.waw5.cache.orange.pl/canal/v/canal/vod/store01/CAP_21042687/_/hd4-hssdrm02.ism/QualityLevels(128000)/Fragments(Polski=0)` and it works perfectly.

but in KODI20 ```https://ec04-szz1.waw5.cache.orange.pl/canal/v/canal/vod/store01/CAP_21042687/_/hd4-hssdrm02.ism/Manifest/QualityLevels(128000)/Fragments(Polski=0)```
so it's cause errors.
After fix it via proxy KODI is crashed.

[lg.txt](https://github.com/xbmc/inputstream.adaptive/files/10116911/lg.txt)

Paco8:
I came here to report the same problem. I attach a sample plugin to reproduce the problem. In the settings you can turn on a proxy. Without the proxy the video won't play because of wrong urls. With proxy kodi crashes with a segmentation fault. The crash happens in all systems I tested (Windows, Android and Mac OS).

[plugin.video.ism-test-0.0.1.zip](https://github.com/xbmc/inputstream.adaptive/files/10246517/plugin.video.ism-test-0.0.1.zip)

glennguy:
Thanks very much @Paco8 for the easy reproduction!

I've solved the crashing I think in #1080, I'll look at the malformed url next.

glennguy:
#1081 should remove the need for a proxy

CastagnaIT:
we can close the issue?

PR #1080, opened 2022-12-17
Remove unneeded unique ptr, resolves crash

glennguy:
`fragment` was being destroyed straight after construction (not an expert so not sure why, something went out of scope??) so removed this variable being defined as a unique pointer. `AP4_MovieFragment` takes ownership of the moof pointer anyway and deletes it upon destruction. Partial fix for issues in #1072


CastagnaIT (approved):

pipapapipipapa:
@glennguy Perhaps `std::make_unique<AP4_MovieFragment>(moof->Clone())`?
Also, why do you need to `moof->Clone()`?
Maybe we can use `std::make_unique<AP4_MovieFragment>(moof)`?
https://github.com/axiomatic-systems/Bento4/blob/master/Source/C%2B%2B/Core/Ap4LinearReader.cpp#L295

PR #1082, opened 2022-12-18
Fix seek audio/video, various subtitle fixes

### LINKED ISSUE #1043: v20.3.0-Nexus Abnormal behavior of FF and REW
### Describe the bug

When I fastforwards a netflix video then back to play(press play key) sometimes an unexpected abnormal behavior happens such as rewind or video freezing. And sometimes during FF x4 if press REW key once, normaly it shoud be FF x2. But very fast rewind (maybe REW x8?) happening sometimes. It seems that these weird behaviors and frequent video freezing or crash after FF are more likely happening when video is resumed.

## Expected Behavior

### To Reproduce

1. I fresh installed latest kodi nexus nightly (0917)  and netflix addon on shield tv.
2. Resume a video and test FF and REW, repeated test of FF and Play, FF and REW is needed, cause it happens sometimes.

### Debuglog

https://drive.google.com/file/d/1-Bjk-tY4qoVLQTfTDkW5qf0naU36_Hyz/view?usp=drivesdk

### MPD/M3U8s/ISM

https://drive.google.com/file/d/1-Sq7Kwl2mGQf3_He3j1mZ1EA3DCoGAjR/view?usp=drivesdk

### Screenshots 

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: android 11/shield TV
 - Kodi version: nexus latest nightly 20220917
 - inputstream.adaptive version: v20.3.0-Nexus

glennguy:
Subtitle fixes (regarding webvtt/hls):
* Solves the duplicate subtitle issue. This was happening after a seek because the last segment to be read before seek would always be sent to VP - now it is flushed first
* Subtitle stream was blocking until the segment buffer was full. Now behaves normally
* Fix for all known cases to align the subtitles properly. This requires changes in Kodi to be able to adjust for timing correction in VP

Regarding seeking - this still keeps the code path to deal with re-starting stopped streams. The relevant section currently in `seek_time` is impossible to read, and evidently doesn't match 1:1 with what was before the change that introduced this issue.
Should fix #1043, #1090

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1104,15 +1104,20 @@ void CSession::StartReader(
     CStream* stream, uint64_t seekTime, int64_t ptsDiff, bool preceeding, bool timing)
 {
   bool bReset = true;
+  ISampleReader* streamReader = stream->GetReader();
   if (timing)
+  {
     seekTime += stream->m_adStream.GetAbsolutePTSOffset();
+  }
   else
+  {
     seekTime -= ptsDiff;
+    streamReader->SetStartPTS(GetTimingStartPTS());
+  }
 
   stream->m_adStream.seek_time(static_cast<double>(seekTime / STREAM_TIME_BASE), preceeding,
                                bReset);
 
-  ISampleReader* streamReader = stream->GetReader();
   if (!streamReader)
   {
     LOG::LogF(LOGERROR, "Cannot get the stream reader");
@@ -1137,6 +1142,7 @@ bool CSession::GetNextSample(ISampleReader*& sampleReader)
 {
   CStream* res{nullptr};
   CStream* waiting{nullptr};
+  CStream* timingStream{GetTimingStream()};
 
   for (auto& stream : m_streams)
   {
@@ -1155,18 +1161,30 @@ bool CSession::GetNextSample(ISampleReader*& sampleReader)
         waiting = stream.get();
         break;
       }
-      else if (!streamReader->EOS() &&
-        AP4_SUCCEEDED(streamReader->Start(isStarted)))
+      else if (!streamReader->EOS())
       {
-        if (!res || streamReader->DTSorPTS() < res->GetReader()->DTSorPTS())
+        // Once the start PTS has been acquired for the timing stream, set this value
+        // to the other stream readers
+        if (stream.get() != timingStream &&
+            timingStream->GetReader()->GetStartPTS() != STREAM_NOPTS_VALUE &&
+            streamReader->GetStartPTS() == STREAM_NOPTS_VALUE)
         {
-          if (stream->m_adStream.waitingForSegment(true))
+          // want this to be the internal data's (not segment's) pts of 
+          // the first segment in period
+          streamReader->SetStartPTS(GetTimingStartPTS());
+        }
+        if (AP4_SUCCEEDED(streamReader->Start(isStarted)))
+        {
+          if (!res || streamReader->DTSorPTS() < res->GetReader()->DTSorPTS())
           {
-            waiting = stream.get();
-          }
-          else
-          {
-            res = stream.get();
+            if (stream->m_adStream.waitingForSegment(true))
+            {
+              waiting = stream.get();
+            }
+            else
+            {
+              res = stream.get();
+            }
           }
         }
       }
@@ -1470,6 +1488,16 @@ int64_t CSession::GetChapterPos(int ch) const
   }
 
   return sum / STREAM_TIME_BASE;
+}
+
+uint64_t CSession::GetTimingStartPTS() const
+{
+  if (CStream* timing = GetTimingStream())
+  {
+    return timing->GetReader()->GetStartPTS();
+  }
+  else
+    return 0;
 }
 
 uint64_t CSession::GetChapterStartTime() const
--- a/src/Session.h
+++ b/src/Session.h
@@ -292,10 +292,20 @@ public:
    */
   uint64_t GetChapterStartTime() const;
 
+  /*! \brief Get starting PTS of the timing stream
+   *  \return PTS at start of the timing stream
+   */
+  uint64_t GetTimingStartPTS() const;
+
   /*! \brief Get value of m_chapterSeekTime
    *  \return Time stored in m_chapterSeekTime
    */
   double GetChapterSeekTime() { return m_chapterSeekTime; };
+
+  /*! \brief Get the timing stream
+   *  \return Timing stream if exists
+   */
+  CStream* GetTimingStream() const { return m_timingStream; }
 
   /*! \brief Set m_chapterSeekTime back to 0
    */
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -489,14 +489,15 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
 
       if (srHaveData)
       {
-        p->dts = static_cast<double>(sr->DTS() + m_session->GetChapterStartTime());
-        p->pts = static_cast<double>(sr->PTS() + m_session->GetChapterStartTime());
+        p->dts = static_cast<double>(sr->DTS());
+        p->pts = static_cast<double>(sr->PTS());
         p->duration = static_cast<double>(sr->GetDuration());
         p->iStreamId = sr->GetStreamId();
         p->iGroupId = 0;
         p->iSize = iSize;
         std::memcpy(p->pData, pData, iSize);
-
+        //! @todo: on Kodi 21, sending of subtitles packet side data is no longer 
+        //! needed and can be removed
         sr->SetDemuxPacketSideData(p, m_session);
       }
 
--- a/src/TSReader.cpp
+++ b/src/TSReader.cpp
@@ -10,10 +10,8 @@
 #include <bento4/Ap4ByteStream.h>
 #include <stdlib.h>
 
-TSReader::TSReader(AP4_ByteStream *stream, uint32_t requiredMask)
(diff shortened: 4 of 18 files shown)


CastagnaIT:
unfurnately this broken again subtitles (see akmai local test stream)
problems are always the same:
- if you seek sometimes subtitle data are sent twice to kodi (double subtitles displayed)
- with normal playback when you go through a chapter subtitles are full broken

glennguy:
Thanks @CastagnaIT 
I'll look closer at these points. I do remember for the second point that there's something that needs to be checked out on Kodi's side.. but I'm not sure it relates to the code here as this is only called on seek.
Definitely need to look at the preceeding condition for the first point

glennguy:
@CastagnaIT  I've been looking into the issue of subs not working when playing through a chapter change. Everything that I can see shows that we download the subs for the new chapter, the correct offset is being sent in the demux packet (and received) and the adaptivestream looks to be the same in either case. Basically no difference from IA point of view.

The difference from Kodi side is that the packet for the subs are sent while VP is still presenting chapter compared to being sent in current (chapter 2) when seeking into the new chapter.

My thoughts are:
The current way of sending a chapter start offset to Kodi is wrong and will only work when PTS for each chapter starts at 0. The PTS/DTS that is sent in the demux packet for subs should equal the starting PTS of the first sample of the timing stream. Your thoughts?

matthuisman:
bump please :) could we try to get this in and released?

Better to fix playback at cost of subtitles for now?

glennguy:
I'll need to separate these things and put into a new PR. But, getting closer on one of the other issues this addresses so we'll see

glennguy:
@matthuisman would you be able to try out a build from my pushes just now?

Note that the D+ 3 second subtitles issue will be back, it requires the changes in https://github.com/CastagnaIT/xbmc/tree/pts_correction for this to work as intended. Hoping that webvtt subs will be rock solid now.

matthuisman:
have confirmed https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1082/5/artifacts fixes the seek issue (https://github.com/xbmc/inputstream.adaptive/issues/1043) for me :) thank you so much

couldnt reproduce the subs issue so would need someone else to confirm that fix

glennguy:
@CastagnaIT ready for review :)

src/main.cpp
```diff
@@ -474,15 +474,14 @@ DEMUX_PACKET* CInputStreamAdaptive::DemuxRead(void)
...
+        sr->SetDemuxPacketSideData(p, m_session);
+        p->dts = static_cast<double>(sr->DTS());
+        p->pts = static_cast<double>(sr->PTS());
         p->duration = static_cast<double>(sr->GetDuration());
         p->iStreamId = sr->GetStreamId();
         p->iGroupId = 0;
         p->iSize = iSize;
         std::memcpy(p->pData, pData, iSize);
-
-        sr->SetDemuxPacketSideData(p, m_session);
```
CastagnaIT, line 500: there is a reason why you moved setter `SetDemuxPacketSideData` upper, so before assigning data to the pointer?
if no reason please restore setter after assigning pointer properties

CastagnaIT: just before `sr->SetDemuxPacketSideData(p, m_session);`
please add a note something like:
`//! @todo: on Kodi 21, sending of subtitles packet side data is not more needed to be removed`

src/Session.h
```diff
@@ -293,11 +293,21 @@ class ATTR_DLL_LOCAL CSession : public adaptive::AdaptiveStreamObserver
...
+
   /*! \brief Get value of m_chapterSeekTime
    *  \return Time stored in m_chapterSeekTime
    */
   double GetChapterSeekTime() { return m_chapterSeekTime; };
 
+  /*! \brief Get the timing stream
+   *  \return Timing stream if exists
+   */
+  CStream* GetTimingStream() const { return m_timingStream; };
```
CastagnaIT:
```suggestion
  CStream* GetTimingStream() const { return m_timingStream; }
```

src/Session.cpp
```diff
@@ -1137,6 +1142,7 @@ bool CSession::GetNextSample(ISampleReader*& sampleReader)
 {
   CStream* res{nullptr};
   CStream* waiting{nullptr};
+  CStream* timingStream(GetTimingStream());
```
CastagnaIT:
```suggestion
  CStream* timingStream{GetTimingStream()};
```

CastagnaIT (commented):
just minor things, but i think you forgot a change,
the commit [Fix subtitle AdaptiveStream from blocking](https://github.com/xbmc/inputstream.adaptive/pull/1082/commits/95decb4bfc57ad150c3b8070fa3e3d7dd8d1f859) cause broken subtitles (see royal family subs) if it takes too long to investigate suggest to remove that commit and if needs some rework do on separate PR

CastagnaIT:
last thing if possible always add context prefixes on commits title, for first two commits:
```
[AdaptiveStream] Revert to previous seek segment logic
[SubtitleSampleReader] Invalidate buffer after seeking
```

glennguy:
> the commit [Fix subtitle AdaptiveStream from blocking](https://github.com/xbmc/inputstream.adaptive/pull/1082/commits/95decb4bfc57ad150c3b8070fa3e3d7dd8d1f859) cause broken subtitles (see royal family subs)

I'm not seeing this here. I have rebased on Nexus locally to include the HLS/default TS commit and it plays fine. I noticed some hiccups sometimes where the subs stop displaying for a segment or so - I think this is the very poor download speed from where I am causing VP resyncs and pushing the subs out of alignment. Apart from that there's no issues (except the duplicate subs as discussed previously).

src/main.cpp
glennguy: I had moved it when trying a previous strategy (using this method to set the start pts offset) but no longer needed. Will move back and add note

glennguy:
Updates done and rebased on Nexus. When you get a chance could you please check this build with the royal family test stream again?

Also note - once this is in we will need to do a release before https://github.com/xbmc/xbmc/pull/22621 is merged to avoid breaking things fixed by the previous sidedata method.

glennguy:
Sorry, accidentally left the log statement uncommented in main.cpp - good to go now.

CastagnaIT:
> Updates done and rebased on Nexus. When you get a chance could you please check this build with the royal family test stream again?

yeah rebuilded again from your upstream but same thing, subtitles are broken
debugging kodi core webvtt parser the text data in the demuxer package is broken
then isa is somewhat corrupting the data (more likely deleted from memory)

happens from segment `index_6_1.vtt` (from here and all next segments are full broken)

this data
```
WEBVTT	#Elemental Media Engine(TM) 2.1.3.0
X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:180000

00:00:03.880 --> 00:00:08.074
(dramatic choir singing)

00:00:08.074 --> 00:00:10.774
- [Anthony] They are two of
the world's most famous men:

00:00:10.774 --> 00:00:13.727
princes, soldiers,
celebrities.
```

become corrupted as:
```
WEBVTT	#Elemental Media Engine(TM) 2.1.3.0
X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:180000

00:00:03.880 --> 00:00:08.074
(dramatic choir singing)

00:00:08.074 --> 00:00:10.774
- [Anthony] They are two of t
```

Edit: im using kodi and isa as debug mode on Windows os, maybe the results may differs in case you are using release mode

CastagnaIT:
@glennguy i fixed broken subtitles due to commit `Fix subtitle AdaptiveStream from blocking`
just a oversight of "if" condition, and also better make worker_processing_ atomic because its used between differents threads
follow the patch

```c++
diff --git a/src/common/AdaptiveStream.cpp b/src/common/AdaptiveStream.cpp
index a3c5bb9..6d96c67 100644
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -991,7 +991,7 @@ bool AdaptiveStream::retrieveCurrentSegmentBufferSize(size_t& size)
 
   std::unique_lock<std::mutex> lckrw(thread_data_->mutex_rw_);
 
-  if (worker_processing_)
+  while (worker_processing_)
   {
     thread_data_->signal_rw_.wait(lckrw);
   }
diff --git a/src/common/AdaptiveStream.h b/src/common/AdaptiveStream.h
index 9ff4fec..e42ba20 100644
--- a/src/common/AdaptiveStream.h
+++ b/src/common/AdaptiveStream.h
@@ -10,6 +10,7 @@
 
 #include "AdaptiveTree.h"
 
+#include <atomic>
 #include <condition_variable>
 #include <map>
 #include <mutex>
@@ -183,7 +184,7 @@ namespace adaptive
     uint64_t absolute_position_;
     uint64_t currentPTSOffset_, absolutePTSOffset_;
 
-    bool worker_processing_;
+    std::atomic<bool> worker_processing_;
     bool m_fixateInitialization;
     uint64_t m_segmentFileOffset;
     bool play_timeshift_buffer_;

```

forgot, in the file AdaptiveStream.h you have introduced a CRLF please remove it

glennguy:
Thanks @CastagnaIT have fixed the introduced CRLF and added your suggested changes to that commit

CastagnaIT (approved):
finally we are ready :)

Issue #1087, opened 2022-12-25
(Kodi 20) It's not possible to play drm videos from ism manifests


Paco8:
### Describe the bug
Here is a clear and concise description of what the problem is:

Once [this previous bug](https://github.com/xbmc/inputstream.adaptive/issues/1072) has been fixed, I noticed that I still couldn't play videos from my addon on Kodi 20. Now the videos start to play but it's like it's playing a corrupt file. I think the problem may be that inputstream.adaptive is not downloading the license, and so not decrypting the streams.

### To Reproduce

I created this plugin, with a sample video. Install the plugin and try to play the video:
[plugin.video.ism-test-0.0.3.zip](https://github.com/xbmc/inputstream.adaptive/files/10299991/plugin.video.ism-test-0.0.3.zip)
The video plays fine on Kodi 19.

### Debuglog

The debuglog can be found here:
https://paste.kodi.tv/pogodojoga.kodi

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
https://ampvideos.streaming.mediaservices.windows.net/9b77ed82-94af-41bc-b08b-537e7a530eb6/sintel_trailer-1080p_fp.ism/manifest

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Operating system version/name: Windows 7
 - Kodi version:  Kodi (20.0-RC2 (19.90.905
 - inputstream.adaptive version:
 I took the one from here: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/Nexus/188/artifacts/

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

glennguy:
Thanks again @Paco8 for the great example to test, #1088 should fix this

PR #1088, opened 2022-12-26
[ISM] Fix regression from refactor

glennguy:
`m_strPSSH` is needed to be set for `insert_pssh` to add the pssh set.


glennguy:
Should fix #1087 
@Paco8 could you try with these changes?

CastagnaIT (approved):
good catch

Paco8:
> Should fix #1087 @Paco8 could you try with these changes?

Yes, the problem seems fixed. Thank you very much.

Issue #829, opened 2021-10-17
SIGSEGV in AP4_LinearReader::GetSample (tracker->m_SampleTable is NULL)


dobo90:
### Describe the bug
Here is a clear and concise description of what the problem is:

I've compiled latest inputstream.adaptive from Nexus branch (c08955927f) and it crashes when playing stream from Polish Canal+ service.

I've debbuged it using gdb. And the oputput is following:
```
Thread 40 "VideoPlayer" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff7b7fe640 (LWP 4631)]
AP4_LinearReader::GetSample (this=0x7fff8406ca48, track_id=4294967295, sample=..., sample_index=0) at /tmp/inputstream.adaptive/lib/libbento4/Core/Ap4LinearReader.cpp:640                                                                                                  
640         return tracker->m_SampleTable->GetSample(sample_index, sample);
(gdb) p tracker
$1 = (AP4_LinearReader::Tracker *) 0x7fff840d1e80
(gdb) bt
#0  AP4_LinearReader::GetSample(unsigned int, AP4_Sample&, unsigned int)
    (this=0x7fff8406ca48, track_id=4294967295, sample=..., sample_index=0)
    at /tmp/inputstream.adaptive/lib/libbento4/Core/Ap4LinearReader.cpp:640
#1  0x00007fff7adccd8f in FragmentedSampleReader::ProcessMoof(AP4_ContainerAtom*, unsigned long long, unsigned long long, unsigned long long) (this=0x7fff8406ca40, moof=0x7fff84040a10, moof_offset=0, mdat_payload_offset=1737, mdat_payload_size=32161)                  
    at /tmp/inputstream.adaptive/src/main.cpp:1399
#2  0x00007fff7ae56692 in AP4_LinearReader::AdvanceFragment() (this=0x7fff8406ca48)
    at /tmp/inputstream.adaptive/lib/libbento4/Core/Ap4LinearReader.cpp:412
#3  0x00007fff7ae569ad in AP4_LinearReader::Advance(bool) (this=0x7fff8406ca48, read_data=true)
    at /tmp/inputstream.adaptive/lib/libbento4/Core/Ap4LinearReader.cpp:474
#4  0x00007fff7ae56dd8 in AP4_LinearReader::ReadNextSample(unsigned int, AP4_Sample&, AP4_DataBuffer&)
[...]

PR #1092, opened 2023-01-03
FragmentedSampleReader: prevent crash if track is not present in reader

kszaq:
Despite several fixes for SIGSEGV introduced recently to ISA, I was still getting segmentation faults on a regular basis with 4K HEVC titles from Player.pl addon (paid login required, geofenced, cannot provide a sample stream, unfortunately): https://github.com/xbmc/inputstream.adaptive/issues/829#issuecomment-1001952771

While [a workaround](https://github.com/xbmc/inputstream.adaptive/issues/829#issuecomment-946686094) is available and working, introducing a hack to Bento4 is not something that we want, so I tried to track this issue.

There are only 2 streams in the movie: one video and one audio stream. I added some logging to `CFragmentedSampleReader::ProcessMoof` and noted the following:
* Both Video and Audio track IDs are initialized to a value different than `TRACKID_UNKNOWN`
* Video track Id == 3 from the start
* Audio track Id == 1 from the start
* At some point Video track Id changes to 2 and Kodi segfaults

The reason why video track ID is bitrate change: 15Mbps video track is chosen over 9.6Mbps video track. 

Since I am not familiar with Representation Chooser, I was only able to propose a workaround: commit in this Pull Request simply modifies track ID in `m_track` but perhaps there is a better place to update it?

### DIFF
--- a/src/samplereader/FragmentedSampleReader.cpp
+++ b/src/samplereader/FragmentedSampleReader.cpp
@@ -319,18 +319,24 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
                                                 AP4_Position mdat_payload_offset,
                                                 AP4_UI64 mdat_payload_size)
 {
-  // For prefixed initialization (usually ISM) we don't yet know the
-  // proper track id, let's find it now
-  if (m_track->GetId() == TRACKID_UNKNOWN)
+  AP4_MovieFragment fragment =
+      AP4_MovieFragment(AP4_DYNAMIC_CAST(AP4_ContainerAtom, moof->Clone()));
+  AP4_Array<AP4_UI32> ids;
+  fragment.GetTrackIds(ids);
+  if (ids.ItemCount() == 1)
   {
-    AP4_MovieFragment fragment =
-        AP4_MovieFragment(AP4_DYNAMIC_CAST(AP4_ContainerAtom, moof->Clone()));
-    AP4_Array<AP4_UI32> ids;
-    fragment.GetTrackIds(ids);
-    if (ids.ItemCount() == 1)
+    // For prefixed initialization (usually ISM) we don't yet know the
+    // proper track id, let's find it now
+    if (m_track->GetId() == TRACKID_UNKNOWN)
+    {
       m_track->SetId(ids[0]);
-    else
+      LOG::LogF(LOGDEBUG, "Track ID changed from UNKNOWN to %u", ids[0]);
+    }
+    else if (ids[0] != m_track->GetId())
+    {
+      LOG::LogF(LOGDEBUG, "Track ID does not match! Expected: %u Got: %u", m_track->GetId(), ids[0]);
       return AP4_ERROR_NO_SUCH_ITEM;
+    }
   }
 
   AP4_Result result;


glennguy:
Thanks for giving this issue some attention!

Just some info on stream changing - whenever IA wants to switch streams for example to change bitrate, it signals to Kodi that there should be a stream change. Kodi in turn asks IA to open the stream (`OpenStream()`) and here a new instance of `FragmentedSampleReader` would be created in this case. So any change to fix this should be inside FragmentedSampleReader.cpp  and no need to worry about what happens in other modules. 

So I guess the assumption of the existing code is that track IDs won't change once established. To confirm this, could you post here in a zip a few video segments from your stream? The last 2 downloaded in the log before crashing, and maybe one from several before that? I can inspect and see if the track ID is changing. Oh and also the log with debug logging enabled too would be great.

One more question on your patch - Is removing the `return` for no track IDs necessary for the fix?

kszaq:
Hi @glennguy, thank you for your detailed response!

I was under impression that if the bitrate change is decided by Representation Chooser, IA simply downloads next part with higher bitrate from the playlist and does not call `OpenStream()`.

In Kodi log, I can see that indeed a representation with different ID is chosen
```
AddOnLog: inputstream.adaptive: [Repr. chooser] Changed representation
Current ID 1 (Bandwidth: 9600000 bit/s, Resolution: 3840x2144)
Next ID 2 (Bandwidth: 15000000 bit/s, Resolution: 3840x2144)
```
and there is no `OpenStream()` later.

I am unable to collect a new log/download segments right now but I have a debug log from yesterday, please find it attached below (redacted to hide login data). It is from one of my tests with some logging added to compare fragment IDs, you can see in the last line that ID mismatch was detected right before the crash.
[01_KODI.zip](https://github.com/xbmc/inputstream.adaptive/files/10342672/01_KODI.zip)

Regarding `return` removal I thought that since the code is now called always, in situations where there is more than one ID in the stream, we should not error with `AP4_ERROR_NO_SUCH_ITEM` but simply continue.

Now I think that the correct way to fix this issue is to make Representation Chooser update the ID in the tracker somehow when the representation changes?

kszaq:
PR updated:
1. Re-added `return AP4_ERROR_NO_SUCH_ITEM` if track ID cannot be determined to prevent a crash.
2. Update Track ID only if it doesn't match.

At this point I consider this the easiest workaround for representation switching.

glennguy:
> IA simply downloads next part with higher bitrate from the playlist and does not call OpenStream().

OpenStream won't happen until the new segment/representation reaches the top of the queue in the buffer. You should see similar to this in the log when the stream actually changes:

```
2023-01-05 19:04:15.296 T:16692   debug <general>: AddOnLog: inputstream.adaptive: DEMUX_SPECIALID_STREAMCHANGE
2023-01-05 19:04:15.296 T:16692   debug <general>: AddOnLog: inputstream.adaptive: GetStreamIds()
2023-01-05 19:04:15.296 T:16692   debug <general>: AddOnLog: inputstream.adaptive: GetStream(2001)
2023-01-05 19:04:15.296 T:16692   debug <general>: AddOnLog: inputstream.adaptive: GetStream(2002)
2023-01-05 19:04:15.296 T:16692   debug <general>: AddOnLog: inputstream.adaptive: GetStream(2003)
2023-01-05 19:04:15.296 T:16692   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 2001 with codec_id 27
2023-01-05 19:04:15.296 T:16692   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 2002 with codec_id 27
2023-01-05 19:04:15.296 T:16692   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 2003 with codec_id 86018
2023-01-05 19:04:15.297 T:16692    info <general>: Opening stream: 2001 source: 256
2023-01-05 19:04:15.297 T:16692   debug <general>: AddOnLog: inputstream.adaptive: OpenStream(2001)
```

You can use the Stream Selection type of 'test' to simulate quick stream changes. Try a stream that doesn't crash and you can see what happens in the log. Rep chooser chooses a new representation/bitrate and then once that segment is ready to be used the stream change process happens. At this point the proper parsing of track IDs should be happening. 

> in situations where there is more than one ID in the stream, we should not error with AP4_ERROR_NO_SUCH_ITEM but simply continue.

Agreed. We don't support more than one track within a segment but it's something that should be supported in the future.

kszaq:
Thank you, I will try to use "test" stream selection.

However, you mentioned that when a representation changes, 
> a new instance of FragmentedSampleReader would be created in this case.
I am unable to find that piece of the code, can you please point me in right direction?

src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -319,17 +319,18 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
                                                 AP4_Position mdat_payload_offset,
                                                 AP4_UI64 mdat_payload_size)
 {
-  // For prefixed initialization (usually ISM) we don't yet know the
-  // proper track id, let's find it now
-  if (m_track->GetId() == TRACKID_UNKNOWN)
   {
```
CastagnaIT, line 327: remove uneeded brackets

src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -319,17 +319,18 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
...
     AP4_MovieFragment fragment =
         AP4_MovieFragment(AP4_DYNAMIC_CAST(AP4_ContainerAtom, moof->Clone()));
     AP4_Array<AP4_UI32> ids;
     fragment.GetTrackIds(ids);
-    if (ids.ItemCount() == 1)
+    // If track ID changed due to representation switch,
+    // update the tracker
```
CastagnaIT: if this is a workaround and there is the need to find a better solution
i think would better change this with some TODO comments

something like this, feel free to improve it
```suggestion
  //! @todo: To implement support for multiple tracks within a segment.
  //! @todo: When we play an adaptive stream and the stream quality change during playback the track ID change and we force update it, but may be need to find an appropriate solution to update the track.
```

kszaq:
Thank you for your review @CastagnaIT, I will update the PR once the investigation is done as there seems to be a better way to fix this...

I have done more testing and you are correct @glennguy that a new `FragmentedSampleReader` is created when the representation is switched. However, it looks like there is a sort of race condition that after a representation change, data from new representation goes to `ProcessMoof` with "old" `FragmentedSampleReader` from previous representation. Afterwards, a correct `FragmentedSampleReader` is used and there is no longer a need to replace track ID.

With the following patch:
```
diff --git a/src/main.cpp b/src/main.cpp
index 871b878..7756dde 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -375,6 +375,8 @@ bool CInputStreamAdaptive::OpenStream(int streamid)
         m_session->GetSingleSampleDecryptor(stream->m_adStream.getRepresentation()->pssh_set_);
     auto caps = m_session->GetDecrypterCaps(stream->m_adStream.getRepresentation()->pssh_set_);

+    LOG::Log(LOGDEBUG, "OpenStream: Create FragmentedSampleReader for track ID  %d", track->GetId());
+
     stream->SetReader(std::make_unique<CFragmentedSampleReader>(
         stream->GetAdByteStream(), movie, track, streamid, sampleDecrypter, caps));
   }
diff --git a/src/samplereader/FragmentedSampleReader.cpp b/src/samplereader/FragmentedSampleReader.cpp
index 5248316..eb44325 100644
--- a/src/samplereader/FragmentedSampleReader.cpp
+++ b/src/samplereader/FragmentedSampleReader.cpp
@@ -324,14 +324,15 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
         AP4_MovieFragment(AP4_DYNAMIC_CAST(AP4_ContainerAtom, moof->Clone()));
     AP4_Array<AP4_UI32> ids;
     fragment.GetTrackIds(ids);
-    // If track ID changed due to representation switch,
-    // update the tracker
-    if (ids.ItemCount() == 1 && ids[0] != m_track->GetId())
+    if (ids.ItemCount() == 1 && m_track->GetId() == TRACKID_UNKNOWN)
+    {
       m_track->SetId(ids[0]);
-    // For prefixed initialization (usually ISM) we don't yet know the
-    // proper track ID, if ID was not found in moof, we need to return now
-    else if (m_track->GetId() == TRACKID_UNKNOWN)
+    }
+    else if (ids[0] != m_track->GetId())
+    {
+      LOG::Log(LOGDEBUG, "ProcessMoof: TrackId does not match!");
       return AP4_ERROR_NO_SUCH_ITEM;
+    }
   }

   AP4_Result result;
```
The result is:
```
2023-01-06 00:03:20.916 T:30484   debug <general>: AddOnLog: inputstream.adaptive: DEMUX_SPECIALID_STREAMCHANGE
2023-01-06 00:03:20.916 T:30484   debug <general>: AddOnLog: inputstream.adaptive: GetStreamIds()
2023-01-06 00:03:20.916 T:30484   debug <general>: AddOnLog: inputstream.adaptive: GetStream(1001)
2023-01-06 00:03:20.916 T:7535    debug <general>: AddOnLog: inputstream.adaptive: ProcessMoof: TrackId does not match!
2023-01-06 00:03:20.916 T:30484   debug <general>: AddOnLog: inputstream.adaptive: GetStream(1002)
2023-01-06 00:03:20.916 T:30484   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 1001 with codec_id 86018
2023-01-06 00:03:20.916 T:30484   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 1002 with codec_id 173
2023-01-06 00:03:20.916 T:30484    info <general>: Opening stream: 1002 source: 256
2023-01-06 00:03:20.916 T:30484   debug <general>: AddOnLog: inputstream.adaptive: OpenStream(1002)
2023-01-06 00:03:21.223 T:30484   debug <general>: AddOnLog: inputstream.adaptive: OpenStream: Create FragmentedSampleReader for track ID 2
2023-01-06 00:03:21.223 T:30484   debug <general>: AddOnLog: inputstream.adaptive: UpdateSampleDescription: codec 1751479857
2023-01-06 00:03:21.223 T:30484   debug <general>: AddOnLog: inputstream.adaptive: GetStream(1002)
2023-01-06 00:03:21.223 T:30484   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 1002 with codec_id 173
2023-01-06 00:03:21.223 T:30484    info <general>: Opening stream: 1001 source: 256
2023-01-06 00:03:21.223 T:30484   debug <general>: AddOnLog: inputstream.adaptive: OpenStream(1001)
2023-01-06 00:03:21.223 T:30484   debug <general>: AddOnLog: inputstream.adaptive: GetStream(1001)
```
I can see that a `FragmentedSampleReader` for the new representation was created but data from the new representation was processed with previous `stream` instance. I hope this makes sense.

I will keep looking later this week...

kszaq:
PR updated: in current state it prevents a crash if data from "new" representation is passed to "old" representation's `FragmentedSampleReader`. I think is should be good to merge now, next step is prevent this from happening at all, in next PR *if* I can find out how to do this. 😄

kszaq:
Ping, please, any additional thoughts? I am aware that this patch does not solve the root cause which is incorrect `FragmentedSampleReader` used for the first fragment but it prevents a segmentation fault and would be good to have.

CastagnaIT:
I can't review please take patience, we are only two active devs on this project

kszaq:
Sorry if I sounded demanding, I guessed the last version got lost because of many comments and modifications down the way. Please take your time, I really appreciate the hard work you put into this project!

glennguy:
Hey @kszaq thanks for your patience.

I'm still trying to understand where you think 
>  in current state it prevents a crash if data from "new" representation is passed to "old" representation's FragmentedSampleReader

might be happening.

I think I've found the means to create a test stream that has the same conditions as in the player.pl content so I can debug locally. Stay tuned!

kszaq:
@glennguy Thank you for looking into this! 

> I'm still trying to understand where you think
> 
>>     in current state it prevents a crash if data from "new" representation is passed to "old" representation's FragmentedSampleReader
> 
> might be happening.

Please see [my comment](https://github.com/xbmc/inputstream.adaptive/pull/1092#issuecomment-1372918306) and the debug log at the end - only the first sample (fragment?) is passed to "old" representation's `FramgentedSampleReader`, the the new one is used.

glennguy:
I haven't forgotten this, will try to get to this week!

glennguy:
Ok for the record this is what is happening:

* demux packets are requested as normal by Kodi
* last demux packet for representation A is sent. `ReadSampleAsync()` is called (which will now start reading representation B
* `adaptiveStream::ensureSegment` is called. This recognises that the first segment buffer has now been consumed and rotates the buffer. The new first segment is from representation B. AdaptiveStream signals the stream has changed.
* A short time after the async read crashes because of the null sample table - the AP4 Linear Reader has track ID from rep A, but MOOF has track ID from rep B.
* Fix is to exit early from `ProcessMoof`. It doesn't matter that we've consumed the first part of this data. The reader and buffer position will be reset once the stream is reopened.

glennguy:
@kszaq could you please remove the temp commit and rebase? Happy to proceed with this, and thanks for your patience while I understood what was happening.

kszaq:
@glennguy Thank you for the detailed insight into what is happening. I removed the temp commit. Just to note, yesterday I added warning/debug log for better visibility on what is happening, I hope it's OK?

glennguy:
Thanks! For the log, I think it would be better to have both statements as `LOGDEBUG`, I don't think we want this to appear in a normal log.

kszaq:
Done.

glennguy (approved):
Thanks heaps @kszaq!

PR #1095, opened 2023-01-07
[FFmpeg] Use size_t size by default for AVPacketSideData

CastagnaIT:
Kodi 21 now use ffmpeg 5,
so we have to change size param of `AVPacketSideData` to `size_t` data type

There is no `Omega` branch yet
so we have to wait team complete repos transition to merge it to the right branch


CastagnaIT:
since now has been merged in xbmc master and Nexus
relative changes to webvtt demuxer this change is now outdated
so i close the PR

PR #1102, opened 2023-01-20
Configurable headers/parameters for manifests/streams

### LINKED ISSUE #1093: The Query params passed to the main manifest are stripped off when downloading the child manifest
### Describe the bug
Consider HLS livestream main manifest -> https://ip:port/manif.m3u8
Here is the structure of the manifest file:
* manif.m3u8
  - manif_1.m3u8
      - transport_1_1.ts
      - transport_1_2.ts
  - manif_2.m3u8
      - transport_2_1.ts
      - transport_2_2.ts

I need to pass token1 into header and token2 as a query params to be able to download any(main or the child) manifest from cdns.

[inputstream.adaptive](https://github.com/xbmc/inputstream.adaptive) is stripping off the query params from the url when downloading the child manifest (Ex: manif_1.m3u8).
The CDN Server would send back 403 or 405 errors as the url is missing the token2 query param. 

## Expected Behavior
The main manifest (Ex:  https://ip:port/manif.m3u8?token2param=token2) gets downloaded
The child manifest (Ex:  https://ip:port/manif_1.m3u8?token2param=token2) gets downloaded

## Actual Behavior
The main manifest (Ex:  https://ip:port/manif.m3u8?token2param=token2) gets downloaded
The child manifest (Ex:  https://ip:port/manif_1.m3u8) do not download(405E) as query param(token2param=token2) is stripped off.

## Possible Fix
Fix at InputStream.Adaptive :
1. Do not strip the query params passed to the main manifest URL, while downloading the subsequent child manifest. I mean you could add back the removed query param after all the base url processing is done.
2. There might already be a way to do this in InputStream.Adaptive. I might have missed it while reading the documentaition, please do suggest if this is the case :).
3. Give a new map property which can be set by the calling code. Ex: We have "inputstream.adaptive.stream_headers" to pass the headers. Similarly there could be a map property "inputstream.adaptive.stream_query_params" which we could pass from client code to populate into the child manifest and ts files.

Option 2 could be the best fix. (As I didnot get the exact reason why the query params are being removed [here](https://github.com/xbmc/inputstream.adaptive/blob/6a02e74d0845ebdc5b913d9fa95c9d2ced279ce2/src/parser/HLSTree.cpp#L478), but not added back later. Probably its important for the library)

Fix at client code:
1. Proxy Server: We could run an internal proxy server to attach query param and then forward the request to CDN server. This will add lag to the calls and delay downloads (though in milli seconds, it could have snow ball effect while live streaming)

### To Reproduce
I cannot give the exact tokens and urls(in log files) with which the issues are happening, but for testing purpose you can use the following url which I copied form (https://demos.jwplayer.com/live-streaming/) 
Test URL: https://wowzaec2demo.streamlock.net/live/bigbuckbunny/playlist.m3u8
You can see that it has a child manifest chunklist_w(some number).m3u8

If I use https://wowzaec2demo.streamlock.net/live/bigbuckbunny/playlist.m3u8?testParam=test as the callback url into InputStream.Adaptive then you will see that while downloading the chunklist_w(some number).m3u8 file it would have stripped off the 
?testParam=test query param.

### Debuglog

[kodi_issue_1093.log](https://github.com/xbmc/inputstream.adaptive/files/10357797/kodi_issue_1093.log)

### MPD/M3U8s/ISM
Test URL: https://wowzaec2demo.streamlock.net/live/bigbuckbunny/playlist.m3u8?testParam=test

### Your Environment
Used Operating system:

 - [X] Android
 - [X] iOS
 - [X] tvOS
 - [X] Linux
 - [X] OSX
 - [X] Windows
 - [X] Windows UWP

 - Kodi version: 19.x
 - inputstream.adaptive version: [19.0.7](https://github.com/xbmc/inputstream.adaptive/releases/tag/19.0.7-Matrix)

CastagnaIT:
As discussed on #1093
this PR has the purpose to introduce configurable url parameters for manifest and streams separately, by using kodi properties,
and also the url parameters sets will be propagated to all manifest childrens (e.g. HLS case)

Since to implement this i had to do some clean up/rework, i also revisited how the headers are handled,

I also found that in a recent PR i have partially broken the purpose of old behaviour (see #625),
with reference to this i have deprecated the possibility to inject headers on manifest url, by using `|` char e.g.
`http:\\www.media.com\manifest.m3u8|?token=example`
This was the first implementation before adding the possibility to use kodi properties to set headers,
_this method has been never documented_ and i think few addons use or are aware of this,
has no sense still keep it (url inject) imo better remove it on Kodi v21 to make easier code management
and is a minor change for addons, that will have to use `inputstream.adaptive.manifest_headers` property.

Others relative rework/cleanups related to this:

I added `AdaptiveTree::Configure` class method this to avoid store another kodiProps to tree class and to allow more easy cloning

On `AdaptiveStream` class the PrepareDownload/Download/write_data methods,
now no longer use class variable members to share data, now data are passed as reference,
which result fewer parameters for methods and better thread safe

What new and changed on kodi parameters:
- new `inputstream.adaptive.manifest_params` to apply params to manifest and children manifests
- new `inputstream.adaptive.manifest_headers` to apply headers to manifest and children manifests
- new `inputstream.adaptive.stream_params` to apply params to streams
- deprecated behaviour `inputstream.adaptive.stream_headers` for now still can be used to set headers to manifests AND streams, but this behaviour is now deprecated and will be removed on Kodi v21, so in future this property will be used to set headers to streams only.

Sidenote:
if you play a manifest url that already include parameters, and you also set `inputstream.adaptive.manifest_params` property,
the parameters of the url will have the precedence over the kodi property, but only for the master manifest, not for the childrens (like HLS)

Tested with live Dash/HLS with subtitles stream samples

### DIFF
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -10,7 +10,7 @@
     name="adaptive"
     extension=""
     tags="true"
-    listitemprops="license_type|license_key|license_data|license_flags|manifest_type|server_certificate|stream_headers|manifest_update_parameter|original_audio_language|max_bandwidth|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max"
+    listitemprops="license_type|license_key|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|max_bandwidth|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max"
     library_@PLATFORM@="@LIBRARY_FILENAME@"/>
   <extension point="xbmc.addon.metadata">
     <platform>@PLATFORM@</platform>
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -36,11 +36,9 @@ static const AP4_Track::Type TIDC[adaptive::AdaptiveTree::STREAM_TYPE_COUNT] = {
 
 CSession::CSession(const PROPERTIES::KodiProperties& kodiProps,
                    const std::string& manifestUrl,
-                   const std::map<std::string, std::string>& mediaHeaders,
                    const std::string& profilePath)
   : m_kodiProps(kodiProps),
     m_manifestUrl(manifestUrl),
-    m_mediaHeaders(mediaHeaders),
     m_KodiHost(std::make_unique<CKodiHost>()),
     m_reprChooser(CHOOSER::CreateRepresentationChooser(kodiProps))
 {
@@ -49,18 +47,20 @@ CSession::CSession(const PROPERTIES::KodiProperties& kodiProps,
   switch (kodiProps.m_manifestType)
   {
     case PROPERTIES::ManifestType::MPD:
-      m_adaptiveTree = new adaptive::DASHTree(kodiProps, m_reprChooser);
+      m_adaptiveTree = new adaptive::DASHTree(m_reprChooser);
       break;
     case PROPERTIES::ManifestType::ISM:
-      m_adaptiveTree = new adaptive::SmoothTree(kodiProps, m_reprChooser);
+      m_adaptiveTree = new adaptive::SmoothTree(m_reprChooser);
       break;
     case PROPERTIES::ManifestType::HLS:
-      m_adaptiveTree = new adaptive::HLSTree(kodiProps, m_reprChooser);
+      m_adaptiveTree = new adaptive::HLSTree(m_reprChooser);
       break;
     default:
       LOG::LogF(LOGFATAL, "Manifest type not handled");
       return;
   };
+
+  m_adaptiveTree->Configure(kodiProps);
 
   m_settingNoSecureDecoder = kodi::addon::GetSettingBoolean("NOSECUREDECODER");
   LOG::Log(LOGDEBUG, "Setting NOSECUREDECODER value: %d", m_settingNoSecureDecoder);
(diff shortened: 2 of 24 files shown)


matthuisman:
i see a v20 tag but I suspect this is going into v21 only?

CastagnaIT:
imo should be better have this improvements/fixes already on v20 (since just released)
another opinion from @glennguy would be also better

glennguy:
Code looks great and I'm happy with the fixes and cleanup, my concern would be introducing a breaking change without notice to addon devs - although, is this broken already? (you mention but I'm not 100% clear)

Although the `|` pipe method is not documented I imagine devs who have worked with just sending to ffmpeg in the past are probably the ones going to use this. I have in the past for example.

But on the other hand, the stream_headers property is documented clearly in the wiki, and we're broken at the moment for some cases where headers/params are to be passed on to children. 

It would be good to know the magnitude of impact we'd be giving but at least it would be a very simple fix for devs to implement.

@matthuisman as a developer of a crapload of add-ons, what are your thoughts?

CastagnaIT:
> (you mention but I'm not 100% clear)

here

original behaviour:
if property (`stream_headers`) set only, you will have same headers for manifest/streams.
if pipe set only,  you will have same headers for manifest/streams.
but if you set the property and also pipe, then you will have differents headers for manifest/streams (streams with property headers and manifest with pipe headers).

rework PR, broken a bit:
if property (`stream_headers`) set only, you will have same headers for manifest/streams.
if pipe set only,  you will have same headers for manifest/streams.
but if you set the property and also pipe, then you will have the headers set in to the property for manifest/streams and the pipe value is ignored.

this PR:
pipe way removed
headers can be set separately for manifest/streams by using relative properties

if you are concerned about removal of "pipe" way
I can see to reimplement it with a deprecation log warning, and will be full removed on v21

so just remains the breaking change of `inputstream.adaptive.stream_headers` property, that will be applied only to streams and not manifests

CastagnaIT:
now that i think i could add an additional behaviour to avoid an immediate breaking change in the v20, so:
if `stream_headers` prop is set and `manifest_headers` is not set, then automatically add same headers also to the manifest and print a log warning about breaking change of `stream_headers` for next v21.

What do you think? could be an acceptable solution?

EDIT: i applied the new changes to the PR, and updated PR description (no breaking changes for v20, but v21)

matthuisman:
I'm not concerned to add support for it. I only need to change code in my
common module :) so go ahead into 20 if you guys want :)

glennguy (approved):
Great!

matthuisman:
just to confirm, for headers to the license_key url, we still use the | ?

CastagnaIT:
i confirm, there are no changes for license

PR #1108, opened 2023-01-29
[HLSTree] Fix container type not set for video

CastagnaIT:
Following test stream cannot currently played
https://cdn-ue1-prod.tsv2.amagi.tv/avod/simplestream-lds-ldtimln/ODYPPCS001EP001/ODYPPCS001EP001.m3u8

this because the real video media link is encoded as url paramater
i have not found info about this way, could be a custom implementation
so in any case i think better fallback to TS when it is a video adaptationset


glennguy (approved):
Very sensible, seems to be .ts streams that are affected in every case I've seen.

PR #1113, opened 2023-02-01
Improved logging during capabilities probe

tmm1:
adds some debug logging during the decrypter probe

### DIFF
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -497,13 +497,17 @@ void WV_CencSingleSampleDecrypter::GetCapabilities(const uint8_t* key, uint32_t 
 {
   caps = { 0, hdcp_version_, hdcp_limit_ };
 
-  if (session_.empty())
+  if (session_.empty()) {
+    LOG::LogF(SSDDEBUG, "Session empty");
     return;
+  }
 
   caps.flags = SSD_DECRYPTER::SSD_CAPS::SSD_SUPPORTS_DECODING;
 
-  if (keys_.empty())
+  if (keys_.empty()) {
+    LOG::LogF(SSDDEBUG, "Keys empty");
     return;
+  }
 
   if (!caps.hdcpLimit)
     caps.hdcpLimit = resolution_limit_;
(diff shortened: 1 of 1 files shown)


wvdecrypter/wvdecrypter.cpp
```diff
@@ -497,13 +497,17 @@ void WV_CencSingleSampleDecrypter::GetCapabilities(const uint8_t* key, uint32_t
 {
   caps = { 0, hdcp_version_, hdcp_limit_ };
 
-  if (session_.empty())
+  if (session_.empty()) {
+    LOG::LogF(SSDDEBUG, "Widevine capabilities: session empty");
```
CastagnaIT: When you use LogF you print also the method name
so imo add also "Widevine capabilities:" is superfluos

```suggestion
    LOG::LogF(SSDDEBUG, "Session is empty");
```

(this and others below)

wvdecrypter/wvdecrypter.cpp
```diff
@@ -543,28 +547,34 @@ void WV_CencSingleSampleDecrypter::GetCapabilities(const uint8_t* key, uint32_t
...
       }
     }
-    catch (...) {
+    catch (const std::exception& e) {
+      LOG::LogF(SSDDEBUG, "Widevine capabilities: error, assuming secure path: %s", e.what());
       caps.flags |= (SSD_DECRYPTER::SSD_CAPS::SSD_SECURE_PATH | SSD_DECRYPTER::SSD_CAPS::SSD_ANNEXB_REQUIRED);
     }
     RemovePool(poolid);
+  } else {
+    LOG::LogF(SSDDEBUG, "Widevine capabilities: decoding not supported?");
```
CastagnaIT: i dont like questions on log,
so find another better formula
or remove it

CastagnaIT:
I also want to mention that adding many log calls will make code execution slower especially on low-end devices
therefore do not abuse Log methods, unless strictly necessary

wvdecrypter/wvdecrypter.cpp
tmm1: Thanks, I didn't notice the prefix was added already.

tmm1:
> I also want to mention that adding many log calls will make code execution slower especially on low-end devices therefore do not abuse Log methods, unless strictly necessary

That's fair, but this only happens once during setup so I think its okay. But you can decide to merge or close.

matthuisman:
this would come in handy :) its only debug level so shouldn't affect most users

CastagnaIT:
@tmm1 its ok, it is just to remind you if you decide to add more in the future

> its only debug level so shouldn't affect most users

declaring the debug level does not change the situation, since API calls are made anyway execution is affected, about this was explained to me by the team a lot time ago

CastagnaIT:
for the future, remember to add a prefix of the context to the commit and pr
this allow us to have a more clean git history to make easier search for something
in this case can become: `[wvdecrypter] add more debug logging during capabilities probe`

glennguy (approved):
Thanks @tmm1!

PR #1125, opened 2023-02-08
[DASHTree] Enable manifest update for representations with SegmentTimeline

### LINKED ISSUE #544: With Facebook live streams, sound is lost after some seconds

CastagnaIT:
## Description

Recent fixes has full fixed facebook live videos at least on Kodi 20,
the only problem is that parsing the manifest it do not set automatically the manifest update "full" param,
with the result that we request wrong segments and we have a semicorrupted playback

currently the "full" param is automatically set for AdaptationSet case but not in the Representations
so here i add the missing code

## Motivation and context

fix #544

## How has this been tested?

Open facebook website
Go to live section play a live video
catch the link and play it on ISA
i cannot attach here a link because after a bit it dead
so i attached a mpd example: [5856040494444186.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/10688441/5856040494444186.mpd.txt)

## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -363,6 +363,9 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
             {
               dash->current_representation_->flags_ |= DASHTree::Representation::TIMELINE;
               dash->currentNode_ |= MPDNODE_SEGMENTTIMELINE;
+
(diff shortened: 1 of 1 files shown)


glennguy (approved):
Makes sense, good find!

Issue #1128, opened 2023-02-09
Widevine AARCH64 support


tmm1:
We have aarch64 version of libwidevinecdm.so available now, as detailed here: https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/530

However to use it is still a challenge. I will document the requirements in this issue.

1. libnspr4.so dependency

    not sure if this is new, but the libnspr4 package must be installed because libwidevinecdm.so depends on libnspr4.so

2. glibc patch for `_dt_determine_tlsoffset`

    this patch is required: https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/devel/glibc/patches/arm/glibc-tls-libwidevinecdm.so-since-4.10.2252.0-has-TLS-with.patch

    without it, there will be a SEGV when running `dlopen()`. note that RPIOS already includes this patch in their `libc6==2.31-13+rpt2+rpi1+deb11u5` package

3. glibc patch for `GLIBC_ABI_DT_RELR` support (if using glibc >= 2.36)

    patch found here: https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/devel/glibc/patches/arm/glibc-HACK-Don-t-check-GLIBC_ABI_DT_RELR-support-for-Chrom.patch

    not currently required on ubuntu (using glibc 2.35) or debian/rpios (using glibc 2.31)

4. LSE atomic symbols present in .dynsym

    inside the .so file, there are references to two required symbols:

    ```
    $ readelf -Ws libwidevinecdm.so | grep -v GLIBC | grep -v WEAK | grep -v NOTYPE | grep UND
        25: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __aarch64_ldadd4_acq_rel
       132: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __aarch64_swp4_acq_rel
    ```

    it seems on chromeos they have modified glibc/gcc packages to provide these as dynamic symbols inside various libraries. on stock linux glibc systems, these are not present because they are only used as static/hidden methods provided by `libgcc.a` and embedded inside every library that uses them. we would expect that these definitions would be embedded inside the widevine.so, but instead they are undefined and expected to be provided by an external library

CastagnaIT:
have you some ideas when will be available a stable build?

tmm1:
I only know what was posted here: https://groups.google.com/a/chromium.org/g/chromium-os-dev/c/GdZ0mebutXw

[...]

PR #1130, opened 2023-02-10
[backport][widevine] add support for linux arm64

tmm1:
without these changes, the following error is observed:

    Initialize: Failed to load library: libwidevinecdm.so: undefined symbol: __aarch64_ldadd4_acq_rel
    Unable to load widevine shared library (libwidevinecdm.so)

cc #1128 https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/530

cc @matthuisman @glennguy @peak3d @knaerzche @jernejsk

for reference:
- https://github.com/devins2518/zig/blob/545c3117ff2f88ee9f29414429a06cedebc0d8c0/lib/compiler_rt/aarch64_outline_atomics.zig#L1421-L1461
- https://github.com/gcc-mirror/gcc/blob/fe4608efc15b881ac908a3f90d7322736495ae72/libgcc/config/aarch64/lse-init.c
- https://github.com/gcc-mirror/gcc/blob/fe4608efc15b881ac908a3f90d7322736495ae72/libgcc/config/aarch64/lse.S
- https://godbolt.org/z/z8W7z1cqx

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -140,6 +140,19 @@ void CSession::SetSupportedDecrypterURN(std::string& key_system)
       m_dllHelper = std::make_unique<kodi::tools::CDllHelper>();
       if (m_dllHelper->LoadDll(item.Path()))
       {
+#if defined(__linux__) && defined(__aarch64__) && !defined(ANDROID)
+        // On linux arm64, libwidevinecdm.so depends on two dynamic symbols:
+        //   __aarch64_ldadd4_acq_rel
+        //   __aarch64_swp4_acq_rel
+        // These are defined in libssd_wv.so, but to make them available in the main binary's PLT,
+        // we need RTLD_GLOBAL. LoadDll() above uses RTLD_LOCAL, so we use RTLD_NOLOAD here to
+        // switch the flags from LOCAL to GLOBAL.
+        void *hdl = dlopen(item.Path().c_str(), RTLD_NOLOAD | RTLD_GLOBAL | RTLD_LAZY);
+        if (!hdl)
+        {
+          LOG::Log(LOGERROR, "Failed to reload dll in global mode: %s", dlerror());
+        }
+#endif
         CreateDecryptorInstanceFunc startup;
         if (m_dllHelper->RegisterSymbol(startup, "CreateDecryptorInstance"))
         {
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -23,9 +23,6 @@
 #include <thread>
 #include <vector>
 
-#if defined(__linux__) && defined(__aarch64__) && !defined(ANDROID)
-#include <sys/auxv.h>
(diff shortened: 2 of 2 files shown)


CastagnaIT:
you can avoid ping peak3d it on threads is not on kodi team from many years
currently the only maintainers are @CastagnaIT and @glennguy

i think you have better knowledge of what is involved in loading the library where linux its not my area
so i do not have much to say

but i am a bit concerned about asm implementation for dynamic symbols
so how, or what we should do if something changes in the future? (maybe possible cases of different library versions?)
may this cause unhandled crashes on InputStreamAdaptive?
is this the only possible way?

i think its better ping also @phunkyfish

tmm1:
I updated the code to remove the ASM. Technically the ASM version will have increased performance because it can use the LSE extensions, but I don't think that's a big deal here. I left that implementation in the git history for reference in case it becomes an issue in the future.

CastagnaIT:
There is an additional problem to be taken in account
this implementation may be good for Nexus but for Omega we have plan to merge ssdwv project to the master project
there will no longer be an additional library builded, then `LoadDll` will be removed
I think you need to find an alternative

tmm1:
Options would be:

1. build separate .so with just the atomic functions which can be dlopen
2. add atomic functions to kodi itself so they are always present in the main binary and available to use

CastagnaIT:
@tmm1 can you create same PR for Omega branch?

matthuisman:
So for arm64 Kodi devices, they will need minimum whatever version IA that is released after this is merged? If so, I'll add a IA version check in my addons to give error if IA version too old for 64bit

glennguy (approved):
Thanks @tmm1 for finding the arm64 libraries and kicking this off

theofficialgman:
> Thanks @tmm1 for finding the arm64 libraries and kicking this off

where is my thanks 😆 , I am the one who noticed the widevine commit in chromium last fall, and that all the arm64 chromeOS CI was in place in January and asked in the google groups. But we really have Doug Anderson to thank.

glennguy:
I guess I was trying to thank tmm1 for bringing it into the space here. Thanks to you for the digging in the first place. In the future though if there's new CDM platform/archs that become available and you find them please share the info with us so we know about it, we would be very grateful. 
If a tree falls in a forest...

Schlumpf7:
so how i get widevine support to work on arm64? the inputstream adaptive helper still claims arm64 is not supported. 
How can i import it manually?

glennguy:
@Schlumpf7 Have a read through this: https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/530

amazingfate:
Hello, there seems to be a gcc fix from fedora: https://bugzilla.redhat.com/show_bug.cgi?id=1830472
https://src.fedoraproject.org/rpms/gcc/c/95507e8b681b5013b87bbbcb0514dd27941e648a
Can I make use of this libwidevine library system wide?

PR #1140, opened 2023-02-16
[AdaptiveTree] Implemented save manifest setting for debug

CastagnaIT:
## Description

This implement a way to automatically save downloaded manifests during playback,
with a configurable debug setting

The downloaded manifest files are saved in to the addon user data folder, under "manifests" folder, e.g.
`..\userdata\addon_data\inputstream.adaptive\manifests`

the filenames are created by using timestamp in the following format:
`manifest_[timestamp number][optional suffix info].txt`

suffix info is used for HLS case (see filenames on attached files below)

in the case where many files are saved too quickly, and the timestamp used have the same number,
the number of the duplicate file count, will be appended to the file name itself, for example:

manifest_1676539356.txt
manifest_1676539356_1.txt
manifest_1676539356_2.txt
...

this will keep also the order of sequential downloads

if the duplicate saved files (with same timestamp) are more than 10, an error will be raised to the log and it stop to save new files with same timestamp, this limit case however should be for some kind of bug to be solved

Each time you start play a new video, the addon delete all previously saved (old) manifests in the data folder.

Plus for each manifest saved:
- will be included in the txt file also the URL where has been downloaded, see attached file example below.
- will be printed in to the log `Manifest saved to: [path of the file]` to have a kind of reference of the filename

## Motivation and context

More times is happened that users are not able to provide the manifest

## How has this been tested?

with HLS live stream you can see all the downloaded manifest in the InputStreamAdaptive addon userfolder

example of saved files:
[manifest_1676561200_master.txt](https://github.com/xbmc/inputstream.adaptive/files/10757855/manifest_1676561200_master.txt)
[manifest_1676561202_child-audio.txt](https://github.com/xbmc/inputstream.adaptive/files/10757856/manifest_1676561202_child-audio.txt)
[manifest_1676561202_child-subtitle.txt](https://github.com/xbmc/inputstream.adaptive/files/10757857/manifest_1676561202_child-subtitle.txt)
[manifest_1676561202_child-video.txt](https://github.com/xbmc/inputstream.adaptive/files/10757859/manifest_1676561202_child-video.txt)

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -9,7 +9,9 @@
 #include "AdaptiveTree.h"
 #include "Chooser.h"
 
+#include "../utils/FileUtils.h"
 #include "../utils/UrlUtils.h"
+#include "../utils/Utils.h"
 #include "../utils/log.h"
 
 #ifndef INPUTSTREAM_TEST_BUILD
@@ -76,6 +78,13 @@ namespace adaptive
 
   void AdaptiveTree::Configure(const UTILS::PROPERTIES::KodiProperties& kodiProps)
   {
+    if (kodi::addon::GetSettingBoolean("debug.save.manifest"))
+    {
+      m_pathSaveManifest = FILESYS::PathCombine(FILESYS::GetAddonUserPath(), "manifests");
+      // Delete previously saved manifest files
+      FILESYS::RemoveDirectory(m_pathSaveManifest, false);
+    }
+
     m_manifestParams = kodiProps.m_manifestParams;
     m_manifestHeaders = kodiProps.m_manifestHeaders;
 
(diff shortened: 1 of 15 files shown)


src/parser/HLSTree.cpp
```diff
@@ -423,6 +419,27 @@ bool HLSTree::ParseManifest(std::stringstream& stream)
   return true;
 }
 
+void HLSTree::SaveManifest(AdaptationSet* adp, const std::stringstream& data, std::string_view info)
+{
+  if (m_pathSaveManifest.empty())
+    return;
+
+  std::string fileNameSuffix = "master";
```
glennguy, line 427: Does braces initialisation work here? `std::string fileNameSuffix{"master"};`

src/common/AdaptiveTree.cpp
```diff
@@ -526,4 +535,32 @@ namespace adaptive
...
+    if (!m_pathSaveManifest.empty())
+    {
+      // We create a filename based on current timestamp
+      // to allow files to be kept in download order useful for live streams
+      std::string filename = "manifest_" + std::to_string(UTILS::GetTimestamp());
+      if (!fileNameSuffix.empty())
+        filename += "_" + fileNameSuffix;
+
+      filename += ".txt";
+      std::string filePath = FILESYS::PathCombine(m_pathSaveManifest, filename);
```
glennguy, line 552: Braces initialisation?

glennguy:
Fantastic idea and implementation @CastagnaIT!!
I started to leave some style comments, but maybe should leave until out of draft sorry!

src/parser/HLSTree.cpp
CastagnaIT: i havent use {} brackets since recent discussions on slack,
so we allow use `=` within code block to make it more readable
but on class/struct vars initialization still use `{}`
see docs changes here: https://github.com/xbmc/xbmc/pull/22639

glennguy:
Perfect, thanks for the detail

glennguy (approved):

glennguy (approved):

PR #1151, opened 2023-02-19
[UrlUtils] Fixed methods to follow RFC 3986

### LINKED ISSUE #1148: non-root relative urls not handeled correctly
### Describe the bug

https://github.com/xbmc/inputstream.adaptive/blob/ba96563de6bce08ff97ec82f04b57dbc5328a7f6/src/utils/UrlUtils.cpp#L86-L89

only checks for relative urls with root path, but relative urls can be without root "foo/aes.key"

#
 - Kodi version: 21
 - inputstream.adaptive version: git

CastagnaIT:
## Description

The new URL utils methods were a bit wrong due to unclear usage of conditions that was checking for backslashes on the old code that manage BASE_URLS, which i later forgot to do a second check to cleanup whole things

now hope i full fixed and the code comply with RFC 3986 specifications
i add also a new unit test to verify results of most common use cases of `UTILS::URL::Join` method

## Motivation and context

fix #1148

## How has this been tested?

unit test and compared also with python `urllib.parse.urljoin` that follow same RFC 3986 specs

## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -101,12 +101,12 @@ static uint64_t ParseSegmentTemplate(const char** attr,
   if (!tpl.timescale) // if not specified timescale defaults to seconds
     tpl.timescale = 1;
 
-  if (!URL::IsUrlAbsolute(tpl.media))
+  if (URL::IsUrlRelative(tpl.media))
   {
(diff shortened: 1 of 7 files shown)


glennguy (approved):
Looks amazing, thanks for all of the tests.

PR #1164, opened 2023-02-24
[DASHTree] fix representation updates when two identical adaptation sets exist

tmm1:
## Description

fixes issue where MPD with multiple audio sets would not process updates if second audio set representation was used

## Motivation and context

given mpd as so:

```
    <Period start="PT6207H10M5.1481183S" id="1">
        <AdaptationSet mimeType="video/mp4" startWithSAP="1" segmentAlignment="true" par="16:9" />
        <AdaptationSet mimeType="audio/mp4" startWithSAP="1" lang="eng" segmentAlignment="true" />
        <AdaptationSet mimeType="audio/mp4" startWithSAP="1" lang="eng" segmentAlignment="true" />
```

observed that representations inside the last AdaptationSet were never updated because the code only processed the first match.

## How has this been tested?

compiled and ran against problem stream to verify fix

## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1787,19 +1787,14 @@ void DASHTree::RefreshLiveSegments()
           if (!updAdaptationSet)
             continue;
 
-          auto& adaptationSets = periods_[index]->adaptationSets_;
-          auto itAs = std::find_if(adaptationSets.begin(), adaptationSets.end(),
-                                   [&updAdaptationSet](const AdaptationSet* item)
-                                   {
-                                     return item->id_ == updAdaptationSet->id_ &&
-                                            item->group_ == updAdaptationSet->group_ &&
-                                            item->type_ == updAdaptationSet->type_ &&
-                                            item->mimeType_ == updAdaptationSet->mimeType_ &&
-                                            item->language_ == updAdaptationSet->language_;
-                                   });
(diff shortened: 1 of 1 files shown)


CastagnaIT:
a question, in your example there are no id fields to distinguish the AdaptationSet's
so in any case the for loop stop at first found

glennguy:
I had to stare at it for a couple of minutes, but makes sense now.
Previously we stop as soon as finding the matching Adp. Now it will go through all matches. Of course, only one of them will contain the matching representation.

glennguy (approved):
Thanks @tmm1!

glennguy:
All ok @CastagnaIT ?

CastagnaIT:
yes but needs backport

tmm1:
> yes but needs backport

do i need to make another PR?

CastagnaIT:
yes usually when a fix is introduced we do a backport to previous Kodi version (atm Nexus)
then yes you need to create a new PR with Nexus branch as base and cherry-pick the commits fixes from your Omega branch

When you create a PR backport, on the PR title (must be same or original PR) is needed add [backport] prefix
on backport PR description is not needed to copy all content of your original development PR (of Omega branch)
but just write "backport of #xxx" (xxx stand for 1164 in this case) to link the original PR

When you create commits try to be concise in the commit title, too long is cut off

PR #1121, opened 2023-02-07
[AdaptiveTree] Lock live segments in start_stream()

dobo90:
Sometimes I'm encountering a crash when starting a live dash stream with a following backtrace:

```
#0  0x910819d4 in adaptive::AdaptiveStream::start_stream() (this=<optimized out>) at /usr/src/debug/kodi-addon-inputstream-adaptive/inputstream.adaptive-20.3.2-Nexus/src/common/AdaptiveStream.cpp:606
#1  0x91069d0c in CInputStreamAdaptive::OpenStream(int) (this=0xaf51fa68, streamid=1006) at /usr/src/debug/kodi-addon-inputstream-adaptive/inputstream.adaptive-20.3.2-Nexus/src/main.cpp:309
#2  0x00e2f5d4 in CDVDDemuxClient::OpenStream(int) ()
#3  0x00e97ad4 in CVideoPlayer::OpenStream(CCurrentStream&, long long, int, int, bool) ()
#4  0x00e9e3c8 in CVideoPlayer::OpenDefaultStreams(bool) ()
#5  0x00ea3b18 in CVideoPlayer::Prepare() ()
#6  0x00ea5dc0 in CVideoPlayer::Process() ()
#7  0x011dcc6c in CThread::Action() ()
```

AdaptiveStream.cpp:606 points to a following line:
```
  currentPTSOffset_ = (next_segment->startPTS_ * current_rep_->timescale_ext_) /
    current_rep_->timescale_int_;
```

After taking a look on the assembly listing:
```
(gdb) disas /m _ZN8adaptive14AdaptiveStream12start_streamEv ...
   0x910819c8 <+1592>:  ldrne   r3, [sp, #28]
   0x910819d0 <+1600>:  ldr     r4, [r3, #324]  ; 0x144
=> 0x910819d4 <+1604>:  ldr     r0, [r8, #40]   ; 0x28
   0x910819d8 <+1608>:  ldr     r12, [r8, #44]  ; 0x2c
   0x910819dc <+1612>:  mov     r8, #0
   0x910819e4 <+1620>:  mov     r3, r8
   0x910819e8 <+1624>:  umull   r0, r1, r0, r4
   0x910819ec <+1628>:  mov     r2, r9
   0x910819f0 <+1632>:  mla     r1, r4, r12, r1
   0x910819f4 <+1636>:  bl      0x9105a6b4 <__aeabi_uldivmod@plt>
...
```

it appears that $r3 contains a valid representation:
```
[...]

PR #1175, opened 2023-02-27
[DASHTree] handle new Period better

tmm1:
## Description

i observed that on a playlist where new `<Period>` is added when going into commercial break, the player would stall out for 30s and then jump ahead in the stream.

the logs showed it as so:

```
2023-02-24 15:44:06.453 T:21159   debug <general>: CVideoPlayerVideo - Stillframe detected, switching to forced 29.970030 fps
2023-02-24 15:44:06.453 T:21146   debug <general>: Stream stalled, start buffering. Audio: 0 - Video: 0
2023-02-24 15:44:06.453 T:21159   debug <general>: CPtsTracker: pattern lost on diff 166833.335938, number of losses 1
2023-02-24 15:44:06.453 T:21146   debug <general>: CVideoPlayer::SetCaching - caching state 1
2023-02-24 15:44:06.453 T:21146   debug <general>: CDVDClock::SetSpeedAdjust - adjusted:0.000000
2023-02-24 15:44:06.453 T:21164   debug <general>: CDVDAudio::Pause - pausing audio stream
2023-02-24 15:44:07.244 T:2472    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
2023-02-24 15:44:08.248 T:2475     info <general>: Skipped 9 duplicate messages..
...
2023-02-24 15:44:41.121 T:2763    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
2023-02-24 15:44:41.808 T:21164    info <general>: Skipped 1 duplicate messages..
2023-02-24 15:44:41.808 T:21164    info <general>: CVideoPlayerAudio::Process - stream stalled
2023-02-24 15:44:41.808 T:21146   debug <general>: Stream stalled, start buffering. Audio: 0 - Video: 2
2023-02-24 15:44:41.809 T:21146   debug <general>: CVideoPlayer::SetCaching - caching state 1
2023-02-24 15:44:41.809 T:21164   debug <general>: CDVDAudio::Pause - pausing audio stream
2023-02-24 15:44:41.809 T:21146   debug <general>: CDVDClock::SetSpeedAdjust - adjusted:0.000000
2023-02-24 15:44:42.121 T:2768    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
2023-02-24 15:44:42.323 T:21159    info <general>: Skipped 2 duplicate messages..
2023-02-24 15:44:42.323 T:21159 warning <general>: OutputPicture - timeout waiting for buffer
2023-02-24 15:44:42.422 T:2769    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
2023-02-24 15:44:43.125 T:2771     info <general>: Skipped 6 duplicate messages..
```

## Motivation and context

The code would not previously deal with new `<Period>` appearing, and would only map Period entries based on their index. 

So if originally the MPD had one period, then all of a sudden it had two, only the first would be updated.

Eventually, if the old period was removed then all of a sudden it would resume playback but skip ahead because it would start adding segments from the new period into the old period again once the position (idx==0) matched again.

Now the code will properly insert new periods into the tree, and also notice when one period ends and switch into the next one.

This looks as so in the new logs:

```
RefreshLiveSegments: Inserting new Period (id=45124, start=23478067190)
ensureSegment: Switching to new Period (id=45124, start=23478067190)
ensureSegment: Switching to new AdaptationSet (startNumber=6014396)
```

## How has this been tested?

compile and runtime tested using a local news stream that keeps adding new Period and was not working previously

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -1115,7 +1115,7 @@ bool AdaptiveStream::seek_time(double seek_seconds, bool preceeding, bool& needR
 
 bool AdaptiveStream::waitingForSegment(bool checkTime) const
 {
-  if (tree_.HasUpdateThread())
+  if (tree_.HasUpdateThread() && state_ == RUNNING)
   {
     std::lock_guard<std::mutex> lckTree(tree_.GetTreeMutex());
     if (current_rep_ && (current_rep_->flags_ & AdaptiveTree::Representation::WAITFORSEGMENT) != 0)
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1002,6 +1002,7 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
       dash->periods_.push_back(dash->current_period_);
       dash->period_timelined_ = false;
       dash->current_period_->start_ = 0;
+      dash->current_period_->sequence_ = dash->current_sequence_++;
 
       for (; *attr;)
       {
@@ -1778,11 +1779,41 @@ void DASHTree::RefreshLiveSegments()
 
       for (size_t index{0}; index < updateTree->periods_.size(); index++)
       {
-        if (index == periods_.size()) // Exceeded periods
-          break;
         auto updPeriod = updateTree->periods_[index];
         if (!updPeriod)
           continue;
+
+        Period* period{nullptr};
+        // find matching period based on ID
+        auto itPd = std::find_if(periods_.begin(), periods_.end(),
+                                 [&updPeriod](const Period* item)
+                                 { return !item->id_.empty() && item->id_ == updPeriod->id_; });
+        if (itPd == periods_.end())
+        {
+          // not found, try matching period based on start
+          itPd = std::find_if(periods_.begin(), periods_.end(),
+                              [&updPeriod](const Period* item)
+                              { return item->start_ && item->start_ == updPeriod->start_; });
+        }
+        // found!
+        if (itPd != periods_.end())
+          period = *itPd;
+        if (!period && updPeriod->id_.empty() && updPeriod->start_ == 0)
+        {
+          // not found, fallback match based on position
+          if (index < periods_.size())
+            period = periods_[index];
+        }
+        // new period, insert it
+        if (!period)
+        {
+          LOG::LogF(LOGDEBUG, "Inserting new Period (id=%s, start=%ld)", updPeriod->id_.c_str(),
+                    updPeriod->start_);
+          updateTree->periods_[index] = nullptr; // remove to prevent delete; we take ownership
+          updPeriod->sequence_ = current_sequence_++;
+          periods_.push_back(updPeriod);
+          continue;
+        }
 
         for (auto updAdaptationSet : updPeriod->adaptationSets_)
         {
@@ -1790,7 +1821,7 @@ void DASHTree::RefreshLiveSegments()
           if (!updAdaptationSet)
             continue;
 
-          for (auto adaptationSet : periods_[index]->adaptationSets_)
+          for (auto adaptationSet : period->adaptationSets_)
           {
             if (!(adaptationSet->id_ == updAdaptationSet->id_ &&
                   adaptationSet->group_ == updAdaptationSet->group_ &&
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1542,7 +1542,10 @@ bool CSession::SeekChapter(int ch)
   if (ch >= 0 && ch < static_cast<int>(m_adaptiveTree->periods_.size()) &&
       m_adaptiveTree->periods_[ch] != m_adaptiveTree->current_period_)
   {
-    m_adaptiveTree->next_period_ = m_adaptiveTree->periods_[ch];
+    auto newPd = m_adaptiveTree->periods_[ch];
+    m_adaptiveTree->next_period_ = newPd;
+    LOG::LogF(LOGDEBUG, "Switching to new Period (id=%s, start=%ld, seq=%d)", newPd->id_.c_str(),
+              newPd->start_, newPd->sequence_);
     for (auto& stream : m_streams)
     {
       ISampleReader* sr{stream->GetReader()};
--- a/src/parser/DASHTree.h
+++ b/src/parser/DASHTree.h
@@ -50,6 +50,7 @@ public:
 
   uint64_t pts_helper_, timeline_time_;
   uint64_t firstStartNumber_;
+  uint32_t current_sequence_;
   std::string current_playready_wrmheader_;
   std::string mpd_url_;
 


tmm1:
This works great now:

```
RefreshLiveSegments: Inserting new Period (id=45211, start=23511749873)                
ensureSegment: Switching to new Period (id=45211, start=23511749873)
ensureSegment: Switching to new Period (id=45211, start=23511749873)                   
EnableStream(1001: false)   
EnableStream(1002: false)
EnableStream(1003: false)   
InitializePeriod: Period (id=45211 start=23511749873)
Reusing DRM psshSets for new period!
DEMUX_SPECIALID_STREAMCHANGE
GetStreamIds()
GetStream(1001)
EnableStream(1001: true)
OpenStream(1001)

**Comment by matthuisman** [comment]
i suspect DEMUX_SPECIALID_STREAMCHANGE is what we do in HLS for x-discontuity so this brings it in line with that?

**Comment by tmm1** [comment]
@matthuisman yea, also matches what happens with MPD VOD when it has multiple Periods. The STREAMCHANGE allows kodi to reinitialize playback and pick the proper streams from the newly available audio/video. It will also reinit drm so new period could use different keys or not have drm at all.

**Comment by matthuisman** [comment]
Awesome!

 Any clever ideas to stop frame rate switching on short periods? Users often find adverts trigger it and then it triggers again going back to main content. I can't think of any clean way to fix that

**Comment by CastagnaIT** [comment]
> Any clever ideas to stop frame rate switching on short periods?

suggest to not spam this PR for OP discussions
however the problem (on kodi side) should be similar of the variable frame rate mkv frame rate switch problem that i tried to fix time ago

File: src/common/AdaptiveStream.cpp [hunk 1]
```diff
@@ -909,8 +909,26 @@ bool AdaptiveStream::ensureSegment()
       std::this_thread::sleep_for(std::chrono::milliseconds(100));
       return false;
     }
+    else if (tree_.HasUpdateThread() && current_period_ != tree_.periods_.back())
+    {
+      auto itPer = std::find(tree_.periods_.begin(), tree_.periods_.end(), current_period_);
+      if (itPer == tree_.periods_.end())
+      {
+        LOG::LogF(LOGWARNING, "Current Period missing from tree!");
```
CastagnaIT: could be useful add a reference of what period ID is missing?

src/common/AdaptiveStream.cpp
```diff
@@ -909,8 +909,26 @@ bool AdaptiveStream::ensureSegment()
...
+        state_ = STOPPED;
+        return false;
+      }
+
+      auto pos = std::distance(tree_.periods_.begin(), itPer);
+      auto newPd = tree_.periods_[pos+1];
```
CastagnaIT: its safe `pos+1`? no need to check for possible out of range?

src/parser/DASHTree.cpp
```diff
@@ -1775,19 +1775,54 @@ void DASHTree::RefreshLiveSegments()
 
       for (size_t index{0}; index < updateTree->periods_.size(); index++)
       {
-        if (index == periods_.size()) // Exceeded periods
-          break;
         auto updPeriod = updateTree->periods_[index];
         if (!updPeriod)
           continue;
 
+        Period *period{nullptr};
```
CastagnaIT: every time you write/edit code always apply Clang format

on this project we dont have Jenkins integrations, just the `.clang-format` rules file
so you have to apply clang format manually so by using your editor shortcuts,
for example on visual studio select the code and CTRL+K+F

src/parser/DASHTree.cpp
```diff
@@ -1775,19 +1775,54 @@ void DASHTree::RefreshLiveSegments()
...
+        {
+          // not found, fallback match based on position
+          if (index < periods_.size())
+            period = periods_[index];
+        }
+        // new period, insert it
+        if (!period)
+        {
+          LOG::LogF(LOGDEBUG, "Inserting new Period (id=%s, start=%ld)", updPeriod->id_.c_str(), updPeriod->start_);
+          std::vector<Period*> tmp = periods_;
```
CastagnaIT: im not able to understand why you need to copy the vector of pointers
can you explain?

src/parser/DASHTree.cpp
```diff
@@ -1775,19 +1775,54 @@ void DASHTree::RefreshLiveSegments()
...
+          updateTree->periods_[index] = nullptr;
```
CastagnaIT: i fear that this will cause a memory leak
periods has been stored with `new` operator but you are replacing the pointer without deleting the object
usually should be done
```
delete vector[index]
current iterator = vector.erase(index)
```

src/common/AdaptiveStream.cpp
tmm1: Yes because the range check happens in the two enclosing conditionals. We know it's not at the back, and we know its in the range.

src/parser/DASHTree.cpp
tmm1: This is done on purpose, I don't want to delete it; I want to take ownership and insert it into the existing period list.

src/parser/DASHTree.cpp
tmm1: May not be required, I will test without it.

tmm1: The reason is because periods_ is used in a thread-unsafe manner (see comment at the top of the function).

So other threads from session/main may be iterating the vector, and if I alter the vector in-place then those iterators are invalidated and will cause exceptions.

So it is simpler to make a copy and then replace for now, until thread safety is added.

src/parser/DASHTree.cpp
tmm1: :+1: thanks

src/parser/DASHTree.cpp
CastagnaIT: yeah the comment on top is a my old comment as todo to check,
this method is called behind a `GetTreeMutex()` in this case the "tree" variables should be safe,
but, there may be old oversights, now long time passed and i dont remember anymore exaclty what was needed to check,
looking at PR #1121 can give an idea where is needed add missing mutex's

now i am not asking you to solve the problem on this PR
but at this point its better add a todo comment above the line
`std::vector<Period*> tmp`
similar to:
`//! @todo: this temporary variable can be safely removed when this method will be thread safe`
in this way will not be forgotten the reason of this when we have to cleanup

tmm1: I removed the tmp copy and it seems to work fine. I think before I was having some other memory corruption issue because the updateTree was deleting the period I copied. I added a comment to make it clear why we insert `nullptr` into the updateTree.

tmm1:
I ran this on a live stream for several hours today. There were 21 new periods and it worked seamlessly across them all. Ready to merge IMHO

glennguy:
I don't think that any changes should be needed to `ensureSegment` for this to work - HLS live multiperiod works correctly at the moment. When we come to the end of a stream the normal triggers should already end/close the stream and cause a streamchange packet to be sent.

Gut feeling is that stream IDs are causing you all the trouble.

Normally for VOD it would look like this in the log as you progress from first to second period:
```
EnableStream(1001: false)
EnableStream(1002: false)
DEMUX_SPECIALID_STREAMCHANGE
GetStreamIds()
GetStream(2001)
GetStream(2002)
OpenStream(2001)
OpenStream(2002)
```

Live HLS is similar but only first chapter is 100x, following chapters are the discontinuity sequence number * 1000 + stream number.

For DASH no `seqence_` or `initial_sequence` values are set on periods. 

I think that period ID for DASH can be alpha numeric so that doesn't help much though. So perhaps we just increment sequentially based on matching new periods to old periods?

tmm1:
> I don't think that any changes should be needed to `ensureSegment` for this to work

You're right. I removed the changes from `ensureSegment` and it still works fine.

tmm1:
@glennguy this should now work for live dash with multi-period exactly the same way it does for VOD. AFAIK VOD has no issues with the lack of distinct stream ids so I think this is good to merge.

glennguy:
Thanks for updating @tmm1 

I can't actually verify properly with the one from the dash live simulator, would you be able to have a look?
```
#KODIPROP:inputstream.adaptive.manifest_update_parameter=full
http://livesim.dashif.org/livesim/periods_20/testpic_2s/Manifest.mpd
```

VOD has distinct stream IDs normally e.g period 1 video is usually 1001, period 2 2001 etc, I'm not certain this is an issue but thought it may be.

Not trying to be a pain, just thinking that it ideally works on more than the 1 test stream. I know there's going to be plenty of cases (maybe yours) where there's just the main program content, and we only ever see going from 1 period to 2, then back to 1. But it would be great to know it can work well in most if not all cases.

If you want we can bring this in and fix other cases going forward, what do you think?

tmm1:
> I can't actually verify properly with the one from the dash live simulator, would you be able to have a look?

Seems in this one ISA is happen to continue streaming the Period it starts with. This is because `SegmentTimeline` is unbounded with no contents, and the pattern can continue forever. When a new period appears, it still follows the same exact template format.

In the real world MPDs I saw, the `SegmentTimeline` usually has `S` entries which tell it how many segments are available and ISA will eventually run out and that's what triggers it to move onto the new Period.



> I know there's going to be plenty of cases (maybe yours) where there's just the main program content, and we only ever see going from 1 period to 2, then back to 1. But it would be great to know it can work well in most if not all cases.

In mine I saw it go up to 3-4 periods sometimes and it still works correctly.



> If you want we can bring this in and fix other cases going forward, what do you think?

This would be my preference.

tmm1:
> VOD has distinct stream IDs normally e.g period 1 video is usually 1001, period 2 2001 etc, I'm not certain this is an issue but thought it may be.

I found a VOD to test and confirmed the stream ids do change there. So I will try to make it work the same way here.

tmm1:
> So I will try to make it work the same way here.

Done, confirmed it goes from 1001 to 2001 to 3001 on live stream.

glennguy:
Thanks @tmm1 for all the updates!

glennguy:
I take the point that the dash live simulator test stream is a very 'out there' case, and something needs to happen to allow the periods in that one to actually transition.

tmm1:
Thanks for merging!

> something needs to happen to allow the periods in that one to actually transition.

Open to suggestions. I'm not sure how to tell the period has ended. That stream still plays correctly so it didn't seem like a huge priority.

glennguy:
Yeah I had a couple of ideas but I think it's not a super huge priority either. I think the stream plays still only because the new period is still using the same segments and there's no actual discontinuity. What I found though is it stops once it gets to the 5:00 mark.
What I would expect with the eventual solution is that the 'now' time marker will move backwards as old periods are removed, and that some post parsing cleanup would need to happen to set the correct length of all periods that aren't the newest i.e if period 2 starts at 3:00 then go back to period 1 and make sure to truncate/finish it at 3:00.
But more questions come out of this:
* If we're playing in the old period near the 'far' edge and the period is removed, we need to then move or seek to the new oldest period, while making sure we don't crash because we've deleted current_segment_ or something like that...
* In this stream example there's always either 6 or 9 minutes of periods, plus the 'never-ending' newest period, yet we only show 5 minutes in the timeshift buffer window. I *think* (iirc) that this 5 minutes is starting at the oldest point and going forwards... I assume that this would be the preferred behaviour but not sure
* Do we truncate/finish the last/newest period according to timeshiftbuffer? Probably...

Anyway lets see what comes of this change and go from there.

PR #1191, opened 2023-03-16
Fix segfault when there is no STSD atom in MP4 stream

deGerard:
## Description

With this patch inputstream.adaptive not crash with a segfault anymore when a MP4 stream doesn't contain a STSD atom but just log an error message that there is no STSD in stream.

## Motivation and context

I have a copy of a BBC dash test stream on my local webserver and due a bug in my webserver I found that inputstream.adaptive causes a segfault. After some debugging I found out that when inputstream.adaptive process a stream without a STSD atom it try to use a null-pointer in the CFragmentedSampleReader constructor what causes the segfault.

## How has this been tested?

To exploit this I change in the initial MP4 file with a hex editor the stsd atom into something like stsq (just some unknown atom). Before apply the patch an segfault happen. After the patch there should be a 'No STSD in stream!' in the log.

## Types of change

## Checklist:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -358,6 +358,13 @@ bool CInputStreamAdaptive::OpenStream(int streamid)
       }
     }
 
+    if (!track->GetSampleDescription(0))
+    {
+      LOG::LogF(LOGERROR, "No STSD atom in stream");
+      m_session->EnableStream(stream, false);
+      return false;
+    }
+
     auto sampleDecrypter =
         m_session->GetSingleSampleDecryptor(stream->m_adStream.getRepresentation()->pssh_set_);
     auto caps = m_session->GetDecrypterCaps(stream->m_adStream.getRepresentation()->pssh_set_);


src/main.cpp
```diff
@@ -358,6 +358,13 @@ bool CInputStreamAdaptive::OpenStream(int streamid)
       }
     }
 
+    if (track->GetSampleDescription(0) == NULL)
```
CastagnaIT, line 361: use c++ nullptr
```suggestion
    if (track->GetSampleDescription(0) == nullptr)
```

CastagnaIT: compare NULL is not needed
```suggestion
    if (!track->GetSampleDescription(0))
```

src/main.cpp
```diff
@@ -358,6 +358,13 @@ bool CInputStreamAdaptive::OpenStream(int streamid)
...
+    {
+        LOG::Log(LOGERROR, "No STSD in stream!");
```
CastagnaIT:
```suggestion
        LOG::LogF(LOGERROR, "No STSD atom in stream");
```

CastagnaIT (commented):
thanks for your first contribution
just minor style fixes

CastagnaIT:
the commits must be squashed with the first one

deGerard:
Sorry, I made a mistake with the pull request. It's something new for me. But it guess it's fine now.

deGerard:
Last change I change the indent width to 2 spaces like the rest of the source.

glennguy (approved):
Thanks @deGerard

PR #1178, opened 2023-02-28
[backport] [DASHTree] handle new Period better

tmm1:
## Description

backport for #1175 

i observed that on a playlist where new `<Period>` is added when going into commercial break, the player would stall out for 30s and then jump ahead in the stream.

the logs showed it as so:

```
2023-02-24 15:44:06.453 T:21159   debug <general>: CVideoPlayerVideo - Stillframe detected, switching to forced 29.970030 fps
2023-02-24 15:44:06.453 T:21146   debug <general>: Stream stalled, start buffering. Audio: 0 - Video: 0
2023-02-24 15:44:06.453 T:21159   debug <general>: CPtsTracker: pattern lost on diff 166833.335938, number of losses 1
2023-02-24 15:44:06.453 T:21146   debug <general>: CVideoPlayer::SetCaching - caching state 1
2023-02-24 15:44:06.453 T:21146   debug <general>: CDVDClock::SetSpeedAdjust - adjusted:0.000000
2023-02-24 15:44:06.453 T:21164   debug <general>: CDVDAudio::Pause - pausing audio stream
2023-02-24 15:44:07.244 T:2472    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
2023-02-24 15:44:08.248 T:2475     info <general>: Skipped 9 duplicate messages..
...
2023-02-24 15:44:41.121 T:2763    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
2023-02-24 15:44:41.808 T:21164    info <general>: Skipped 1 duplicate messages..
2023-02-24 15:44:41.808 T:21164    info <general>: CVideoPlayerAudio::Process - stream stalled
2023-02-24 15:44:41.808 T:21146   debug <general>: Stream stalled, start buffering. Audio: 0 - Video: 2
2023-02-24 15:44:41.809 T:21146   debug <general>: CVideoPlayer::SetCaching - caching state 1
2023-02-24 15:44:41.809 T:21164   debug <general>: CDVDAudio::Pause - pausing audio stream
2023-02-24 15:44:41.809 T:21146   debug <general>: CDVDClock::SetSpeedAdjust - adjusted:0.000000
2023-02-24 15:44:42.121 T:2768    debug <general>: AddOnLog: inputstream.adaptive: ensureSegment: Begin WaitForSegment stream 1654049944917item-06item
[...]

PR #1201, opened 2023-03-30
[backport][DASH] fix live streams not playing, and duration updates.

### LINKED ISSUE #1152: Kodi Kayo addon [Aus] time issue for live sport
With the recent Kayo update for Kodi, when selecting a live sport you normally get the time of eg 0:00/1:23:53 meaning you are at start of live and it is 1:23 into live. As the game progresses say 30 mins and it is still live you get 30:00:00/1:53:53 so both clocks move.

Now what is happening is you see 30:00:00/1:23:53 so the end clock does not move even though the live sport is still on.

Then when you get to after this time you see for eg 1:30:00/1:23:53 instead of seeing 1:30:00/3:13:53.

Now when you skip say 10sec when the first clock is greater then the 2nd clock it goes back to the 1:23:53 time even when trying to skip forward. If you use the up button which is a 10min skip function forward this one works.

Was mentioned to me that it is potentially an inputstream.adaptive issue not Matt's excellent Kayo addon for Kodi.

Just wondering if anyone else is seeing this issue and can be rectified?

glennguy:
## Description

Two changes here:
* Updating the duration: some DASH configurations would never update the duration. Fixes for where the timeshiftbuffer duration changes on refresh (fixes #1152)
* Recent changes in #1178 brought a new variable to keep a sequence number on the current period. This lacked initialization causing strange behaviour

## Motivation and context

Fixes #1152 and unreported issue

## How has this been tested?

Tested on Kayo AU add-on. Before - no playback. Before #1178 - no stream duration updates. After both are fixed.

## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1027,7 +1027,7 @@ static void XMLCALL start(void* data, const char* el, const char** attr)
     const char *mpt(0), *tsbd(0);
 
     dash->firstStartNumber_ = 0;
-
+    dash->current_sequence_ = 0;
     dash->overallSeconds_ = 0;
     dash->stream_start_ = dash->GetNowTime();
     dash->mpd_url_ = dash->base_url_;
@@ -2001,6 +2001,7 @@ void DASHTree::RefreshLiveSegments()
                     LOG::LogF(LOGDEBUG,
                               "Full update with start number (repr. id: %s current start:%u)",
                               updRepr->id.c_str(), repr->startNumber_);
+                    overallSeconds_ = updateTree->overallSeconds_;
                   }
                 }
               }


CastagnaIT:
nexus branch?

glennguy:
Since your PR is very close I thought to not bother with Omega. But I can if you expect it to take a while to review and sort out bugs. I don't mind taking the approach of fixing forward with that one though. I'll let you decide, please say if you think I should PR against Omega too :)

CastagnaIT:
~OK then I'll link the PR and i add your changes to my PR, otherwise its difficult tell if it was included in Omega~
chenged idea better make separate PR less confusion

CastagnaIT (approved):
confirmed this has fixed also facebook live regression

Issue #1212, opened 2023-04-07
[Disney+] Subtitles appear 6 seconds early


MauriceW67:
### Describe the bug
Here is a clear and concise description of what the problem is:

When playing some titles with @matthuisman 's Disney+ addon, the subtitles appear exactly 6 seconds early.

Shows that have the issue: Loki, Marvel Studios LEGENDS, The Book of Boba Fett, National Treasure Edge of History

NOTE: The issue seems similar to https://github.com/xbmc/inputstream.adaptive/issues/993 but this was supposed to be fixed. Also, subtitle fixes from https://github.com/xbmc/inputstream.adaptive/issues/1090 were supposed to be part of the fix, but unfortunately there still is an issue and the time difference is now 6 seconds instead of 3.

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

The subtitles should appear at the correct time and not 6 seconds early.

## Actual Behavior

Subtitles appear 6 seconds early on some shows.

### To Reproduce

1. Start Kodi v21
2. Open the Disney+ addon
3. Browse a show that has the issue (I used National Treasure Edge of History)
4. Play the episode

### Debuglog

The debuglog can be found here:
Kodi 21 log: https://paste.kodi.tv/amogagefef.kodi
Kodi 20.1 log: https://paste.kodi.tv/hivojowoco.kodi

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:

### Screenshots 
Here are some links or screenshots to help explain the problem:

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

### Your Environment
Used Operating system:

 - [X] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Nvidia Shield Experience 9.1.0 (Android TV 11)
 - Kodi version: Kodi 21 Omega (kodi-20230406-42f35c38-master-arm64-v8a)
 - inputstream.adaptive version: 21.1.0

NOTE: Issue also occurs with Kodi 20.1 Nexus and IS:A 20.3.6

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

tsioukas:
[...]

PR #1219, opened 2023-04-09
Fix audio / subtitles regressions

CastagnaIT:
## Description

After reworking the parsers, it was not entirely clear to me how the original previous dash code around this worked it was quite a mess, now i fixed whole things hope in better way

`SetCodecInternalName` is not really used we could also drop it and get codec from representation but
i kept it to have the original codec string associated to the stream

in the dashparser i have better written the use subtitles type use cases before it was not intuitive at all

This restore also HLS audio codec workaround where now we have understand the reason why was there
ref: https://github.com/xbmc/inputstream.adaptive/issues/1212#issuecomment-1501309356

## Motivation and context

problem found while investigating for #1212

## How has this been tested?

webvtt ISOBMFF now works correctly
verified also
webvtt as file and webvtt segmented

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -878,10 +878,10 @@ void CSession::UpdateStream(CStream& stream)
                                extraData->size());
   }
 
-  stream.m_info.SetCodecInternalName(rep->GetFirstCodec()); //! @todo: to be verified
   stream.m_info.SetCodecFourCC(0);
   stream.m_info.SetBitRate(rep->GetBandwidth());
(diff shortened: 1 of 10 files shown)


glennguy (approved):
Thanks!

CastagnaIT:
@glennguy since i have changed some code in the subtiltesamplereader i included a couple of fixups

glennguy (approved):

PR #1221, opened 2023-04-12
[AdaptiveStream] Fix tree mutex deadlock

CastagnaIT:
## Description

commit https://github.com/xbmc/inputstream.adaptive/commit/181d600ef22527a0c8456dd8b4c99e86c2cd4fb8
has removed `lckTree.unlock();` from `AdaptiveStream::seek_time` method
what happens is:
1) AdaptiveStream::seek_time -> lock tree
2) after call ResetCurrentSegment that call StopWorker
3) meatime a download can be in progress
4) if a download in in progress call AdaptiveStream::write_data
5) AdaptiveStream::write_data call HLS `OnDataArrived` overrided method
6) deadlock

the code after the `lckTree.unlock();` that i have add seem not needs the tree lock but 
i could be wrong @glennguy can you try make a double check?

## Motivation and context

Video seek can cause a/v freeze

## How has this been tested?

akmai HLS sample stream

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -841,9 +841,9 @@ bool AdaptiveStream::ensureSegment()
     // wait until worker is ready for new segment
     std::unique_lock<std::mutex> lck(thread_data_->mutex_dl_);
     // lock live segment updates
-    std::lock_guard<std::mutex> lckTree(tree_.GetTreeMutex());
+    std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(tree_.GetTreeUpdMutex());
 
-    if (tree_.HasUpdateThread() && SecondsSinceUpdate() > 1)
+    if (tree_.HasManifestUpdates() && SecondsSinceUpdate() > 1)
     {
       tree_.RefreshSegments(current_period_, current_adp_, current_rep_, current_adp_->GetStreamType());
       lastUpdated_ = std::chrono::system_clock::now();
@@ -986,7 +986,7 @@ bool AdaptiveStream::ensureSegment()
         return false;
       }
     }
-    else if (tree_.HasUpdateThread() && current_period_ == tree_.m_periods.back().get())
+    else if (tree_.HasManifestUpdates() && current_period_ == tree_.m_periods.back().get())
     {
       if (!current_rep_->IsWaitForSegment())
       {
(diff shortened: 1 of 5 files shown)


glennguy:
Really good catch finding the old commit.

This makes sense to me. The last part of the seek function unprotected will allow for write_data to do its thing, but what happens if there's a tree update between there and needing to do the comparison of newSeg/old_seg, or setting `current_rep_->current_segment_ = old_seg`?

Again I'm no expert in this, would it be better to unlock only for `ResetCurrentSegment(newSeg)` (and relock afterwards) ? Would it be hacky to add a parameter to `StopWorker` to determine whether acquiring the lock is needed?

CastagnaIT:
> Again I'm no expert in this, would it be better to unlock only for ResetCurrentSegment(newSeg) (and relock afterwards) ?

if you unlock and relock there is no guarantiee that in the middle no tree updates happens
also because when you unlock it after that you need to wait download operation its stopped that may require some time

> Would it be hacky to add a parameter to StopWorker to determine whether acquiring the lock is needed?

the problem not is in the StopWorker but this lock:
https://github.com/xbmc/inputstream.adaptive/blob/Omega/src/parser/HLSTree.cpp#L668
we are acquiring two times the same tree mutex

i need to try thinking more

glennguy:
Ok I'm clearer now, thankyou.

So is the problem really that we should never be trying to do a tree refresh while segment download is occurring? Can we assume that if data is coming in, that we won't be needed do refresh tree yet and so make the tree update wait for a download finished signal?

CastagnaIT:
not exactly as is implemented the lock cover whole code in methods
i have a kind of idea to implement a kind of tree update "pause" reentrant i will let you know

CastagnaIT:
i made the changes not easy to find a good solution at least with the current code,
a kind of solution was using `std::recursive_mutex`, but recursive mutex dont allow to be used with differents threads
at same time, one excludes the other...so dont fit our use case

so the only way i have found is create a custom mutex embedded in the new `TreeUpdateThread` (this class is quite similar to `THREADDATA` in the CAdaptiveStream) where we can have a reentrant effect without depending on any thread
then the last lock_guard that unlock/"unpause" the updates will unblock the update thread

another note `RefreshSegments` method in the DASH/HLS parsers
was calling `RefreshUpdateThread` tree method, but i noticed that was noop... it was trying to force executing thread update but the thread update is executed only when a timeout occurs, so i have full removed related code

i tested dash/hls live seem to works as before
but better if you can try give a second round of tests

CastagnaIT:
now that i think maybe `RefreshUpdateThread` was used inappropriately as a reset for the CV timer?
so to give the right start time? i will take a look

EDIT:
i have reimplemented it in a more clear way
but by looking this seem that we are doing wrong things with manifest updates, or i understand wrong,
i thought the interval (in secs) was for updating the manifest every tot seconds, but here update intervals seem not respected

CastagnaIT:
temporary set wip, i noticed only now a weird delay on 2 dash tests, i will need to investigate

CastagnaIT:
fixed now you can review

glennguy (approved):
Great!

PR #1222, opened 2023-04-16
[AdaptiveStream] Cleanup download methods

CastagnaIT:
## Description

Add CUrl utils, this can be used to simpify our code and will be useful for future PR to cleanup  of`AdaptiveTree::download`
and also to cleanup code in future ssd merge to main project

to allow make better clear code to  `DownloadImpl` (previous `download` method) i had to change 
`segment_buffers_` to a raw pointer vector, that allow to:
- Get the SEGMENTBUFFER pointer from the vector in safe way and sempify the code (since we make use of `std::rotate` before was not possible get the pointer from the vector)
- Code execution seedup, e.g. we make extensive use of `std::rotate`, despite we are trying to "reuse" same SEGMENTBUFFER objects in the `segment_buffers_` vector until now it was a futile effort because `std::rotate` have high overhead to move objects so bad performances:
_Little example `std::rotate` over 100000 items:_
of class objects  6.77702 ms (our previous implementation)
vs
of class pointers 0.04552 ms (as this PR do)
Since our segment buffers are usually around 120 size we have however a little speed improvement,
instead testing dereferencing 100000 pointers vs get objects, has irrilevant timing difference we don't have to worry

I have also removed the check for header "connection":
https://github.com/xbmc/inputstream.adaptive/blob/21.1.0-Omega/src/common/AdaptiveStream.cpp#L227-L228
this has never been documented and there is nothing in the GIT history to explain the reason for doing weird thing
if anyone knows the reason, please leave feedback, but IMO if there is need to add `keep-alive` should be add like all other headers by using ISA headers properties already available

~PR depends from #1221 i would like avoid rebase each time~ edit: rebased

## Motivation and context

Make a semplified CUrl interface allow better future code maintaining
Download methods was a bit a mess

## How has this been tested?

Played some Dash/HLS

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -217,13 +217,11 @@ void AdaptiveStream::worker()
 
       lckdl.lock();
 
+      if (!isSegmentDownloaded)
       {
         std::lock_guard<std::mutex> lckrw(thread_data_->mutex_rw_);
-        if (!isSegmentDownloaded)
-        {
-          // Download cancelled or cannot download the file
-          state_ = STOPPED;
-        }
+        // Download cancelled or cannot download the file
+        state_ = STOPPED;
       }
 
       // Signal finished download
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -12,6 +12,7 @@
 #include "../WebmReader.h"
 #endif
 #include "../oscompat.h"
+#include "../utils/CurlUtils.h"
 #include "../utils/UrlUtils.h"
 #include "../utils/log.h"
 #include "Chooser.h"
@@ -22,9 +23,6 @@
 #include <iostream>
 
 #include <bento4/Ap4.h>
-#ifndef INPUTSTREAM_TEST_BUILD
-#include <kodi/Filesystem.h>
-#endif
 #include "kodi/tools/StringUtils.h"
 
 using namespace adaptive;
@@ -32,8 +30,6 @@ using namespace std::chrono_literals;
 using namespace kodi::tools;
 using namespace PLAYLIST;
 using namespace UTILS;
-
-const size_t AdaptiveStream::MAXSEGMENTBUFFER = 10;
 
 uint32_t AdaptiveStream::globalClsId = 0;
 
@@ -66,7 +62,6 @@ AdaptiveStream::AdaptiveStream(AdaptiveTree& tree,
     assured_buffer_length_(5),
     max_buffer_length_(10)
 {
-  segment_buffers_.resize(MAXSEGMENTBUFFER + 1);
   current_rep_->current_segment_ = nullptr;
 
   // Set the class id for debug purpose
@@ -81,6 +76,7 @@ AdaptiveStream::~AdaptiveStream()
   Stop();
   DisposeWorker();
   clear();
+  DeallocateSegmentBuffers();
 }
 
 void AdaptiveStream::Reset()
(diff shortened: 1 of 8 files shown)


src/common/AdaptiveStream.cpp
```diff
@@ -90,6 +86,244 @@ void AdaptiveStream::Reset()
   absolutePTSOffset_ = 0;
 }
 
+void adaptive::AdaptiveStream::AllocateSegmentBuffers(size_t size)
+{
+  size++; // TTHR
+
+  while (size-- > 0)
+  {
+    segment_buffers_.emplace_back(new SEGMENTBUFFER());
```
lrusak, line 95: Why not use a unique_ptr?

src/common/AdaptiveStream.h
```diff
@@ -89,50 +89,76 @@ class AdaptiveStream;
     bool StreamChanged() { return stream_changed_; }
 
   protected:
+    virtual bool parseIndexRange(PLAYLIST::CRepresentation* rep, const std::string& buffer);
+
+    virtual void SetLastUpdated(std::chrono::system_clock::time_point tm) {};
```
lrusak:
```suggestion
    virtual void SetLastUpdated(const std::chrono::system_clock::time_point tm) {};
```

src/common/AdaptiveStream.h
```diff
@@ -89,50 +89,76 @@ class AdaptiveStream;
...
+    // Segment download section
+
+    struct SEGMENTBUFFER
+    {
+      std::string buffer;
+      PLAYLIST::CSegment segment;
+      uint64_t segment_number{0}; // SEGMENT_NO_NUMBER is initialization segment!
+      PLAYLIST::CRepresentation* rep{nullptr};
+    };
+    std::vector<SEGMENTBUFFER*> segment_buffers_;
```
lrusak, line 113: unique_ptr

src/common/AdaptiveTree.cpp
```diff
@@ -21,6 +21,7 @@
 #endif
 
 #include <algorithm>
+// #include <cassert>
```
lrusak: Dead code?

src/common/AdaptiveTree.h
```diff
@@ -85,7 +86,7 @@ class ATTR_DLL_LOCAL AdaptiveTree
 
   AdaptiveTree(CHOOSER::IRepresentationChooser* reprChooser);
   AdaptiveTree(const AdaptiveTree& left);
-  virtual ~AdaptiveTree();
+  virtual ~AdaptiveTree() {}
```
lrusak:
```suggestion
  virtual ~AdaptiveTree() = default;
```

src/common/AdaptiveTree.h
```diff
@@ -186,6 +186,54 @@ class ATTR_DLL_LOCAL AdaptiveTree
 
   Settings m_settings;
 
+  class TreeUpdateThread
+  {
+  public:
+    TreeUpdateThread() {}
```
lrusak:
```suggestion
    TreeUpdateThread() = default;
```

src/common/AdaptiveTree.h
```diff
@@ -186,6 +186,54 @@ class ATTR_DLL_LOCAL AdaptiveTree
...
+
+    // \brief As "std::mutex" unlock, but resume the manifest updates (support std::lock_guard).
+    void unlock() { Resume(); }
+
+  private:
+    void Worker();
+    void Pause();
+    void Resume();
+
+    std::thread* m_updateThread{nullptr};
```
lrusak: Why is this a pointer?

src/utils/CurlUtils.h
```diff
@@ -0,0 +1,98 @@
...
+
+ /*!
+  * \brief Get the download speed in byte/s. To be called at the end of download.
+  */
+  double GetDownloadSpeed();
+
+ /*!
+  * \brief Get the total byte read of the download (total of chunks size).
+  */
+  size_t GetTotalByteRead() { return static_cast<size_t>(m_bytesRead); }
```
lrusak: Why not just return ssize_t ?

src/common/AdaptiveStream.cpp
CastagnaIT: because we need higher performance with `segment_buffers_` its what i mentioned in description above
if i use unique_ptr's need double indirections and std::rotate becomes very heavy to process, become even worse than before:
11.177 ms of uniqueptr instead of 0.04552 ms by using raw pointers
although it makes objects creation/destruction easier, it is not worth

src/common/AdaptiveStream.h
CastagnaIT: ~this is not part of the PR change
this PR need a rebase with the linked PR that need to be merged before this
so you have to exclude the first commit from the review~
just moved the method, the old code is unchanged but i will address it

src/common/AdaptiveStream.h
CastagnaIT: answered above

src/common/AdaptiveTree.cpp
CastagnaIT: intentional for debug, but is not part of this PR

src/common/AdaptiveTree.h
CastagnaIT: not part of this PR

src/utils/CurlUtils.h
CastagnaIT: `ssize_t` its a type used by kodi API interface only to return -1 value in case of internal errors on callback,
there is no reason to propagate this type on the ISA source code,
its not useful and make only more complex the code for nothing (bytes cannot be -1)

src/common/AdaptiveTree.h
CastagnaIT: please place this question on the other PR, its not part of this

src/common/AdaptiveTree.h
CastagnaIT: addressed on the right PR

src/common/AdaptiveTree.h
CastagnaIT: addressed on the right PR

src/utils/CurlUtils.h
CastagnaIT: now that i am thinking better change m_bytesRead to size_t
and move the cast to ReadChunk since this is the total bytes may increase a lot

src/common/AdaptiveStream.h
```diff
@@ -89,50 +89,76 @@ class AdaptiveStream;
...
+   /*!
+    * \brief Download a file, the representation chooser will be updated with current download speed.
+    * \param downloadInfo The info about the file to download
+    * \param data[OUT] The downloaded data
+    * \return Return true if success, otherwise false
+    */
+    virtual bool Download(const DownloadInfo& downloadInfo, std::string& data);
 
+   /*!
+    * \brief Download a segment file in chucks, the data could be also decrypted by the manifest parser,
```
glennguy: small typo - chucks -> chunks

src/common/AdaptiveStream.cpp
```diff
@@ -90,6 +86,244 @@ void AdaptiveStream::Reset()
   absolutePTSOffset_ = 0;
 }
 
+void adaptive::AdaptiveStream::AllocateSegmentBuffers(size_t size)
+{
+  size++; // TTHR
```
glennguy: I see this comment from the original GSOC rework survives - I still don't know what this means...

src/common/AdaptiveStream.cpp
```diff
@@ -90,6 +86,244 @@ void AdaptiveStream::Reset()
...
+      }
+    }
+
+    if (downloadStatus == CURL::ReadStatus::ERROR)
+    {
+      LOG::Log(LOGERROR, "[AS-%u] Download failed, cannot read chunk: %s", clsId, url.c_str());
+    }
+    else if (downloadStatus == CURL::ReadStatus::CHUNK_READ)
+    {
+      // Chunck reading operations have been stopped
```
glennguy: chunk

src/common/AdaptiveStream.cpp
```diff
@@ -90,6 +86,244 @@ void AdaptiveStream::Reset()
...
+        return false;
+      }
+
+      size_t totalBytesRead = curl.GetTotalByteRead();
+      double downloadSpeed = curl.GetDownloadSpeed();
+
+      // Set current download speed to repr. chooser (to update average).
+      // Small files are usually subtitles and their download speed are inaccurate
+      // by causing side effects in the average bandwidth so we ignore them.
+      static const size_t minSize{512 * 1024}; // 512 Kbyte
```
glennguy, line 218: I know this is not changed in this PR so feel free to take this as a comment only.
Depending on how you look at it, many areas of the world are 'lucky' to enjoy good internet speeds.

In my wonderful country many people have rotting copper connections or can only connect via congested 4G. I went on weekend vacation a month ago and 4G was able to give me 500kbps... sometimes. This situation is reflected by the streams that providers encode and supply:

image
720p is obviously reserved for those with 'elite' internet connections😆 

Could I suggest lowering this to 128K or 64K, which would still easily exclude most subtitle streams?

glennguy:
Just very minor things, looks awesome :)
Performance improvements are great too

src/common/AdaptiveStream.cpp
CastagnaIT: me too... (i will remove it)
are you know why we have to add +1?

src/common/AdaptiveStream.cpp
CastagnaIT: time ago i had thought about this, but for now i put in the roadmap,
the problem is that my previous tests with downloaded files less than 512kb we have completely wrong speed values,
then if you allow get speed values for less than 512 kb downloaded file in any case the "chooser" will work bad so we do not solve the problem but workaround it

mainly two additional checks are needed:
- Check how empty the buffer is, and inform the chooser to shift quality when we have critical empty buffer
- (and?) Detect VP stall could be another solution, then inform the chooser to shift quality

plus another one:
- Multi base url support, this allow to change endpoint in the case of problems, e.g. with n€tflix we have more addresses, and some times (rare cases, usually happens for congested network) need to manually change it to have a good buffering (happens to me only one time during Christmas festivities weeks)

this also requires some logic to cooperate with the download speed detected
so if the download speed is not reported there are other checks
this my theory put into practice obviously there is fun

src/common/AdaptiveStream.cpp
lrusak: Have you considered using std::deque instead of std::rotate and std::vector? If all you want is to move the front to the back that may be a better option (and much faster).

CastagnaIT: yes already thought about something similar
but our old code is quite mess and difficult to understand
nor me nor glennguy we are able to understand in full how works all things, so we are changing/improving the code bit by bit, there are still grey areas in the use of buffering variables so atm i dont touch it

glennguy (approved):

Issue #736, opened 2021-07-14
[To do] wvdecrypter Refactor


matthuisman:
We should move the wvdecrypter code into the main program as it's nearly always required.
This stops needing to load it up separately with dlopen.
It may speed it up, and makes it less complicated

Currently we have
ia.so > wvdecrypter.so > libwidevine.so
with above, we would have
ia.so > libwidevine.so

I think it started off separately as it needed to be shipped separate as the team didn't at that stage want the libwidevine integration code in the repo. that has now changed and currently wvdecrypter is shipped with all builds.

Maybe at the same time, the IA specific code for bento could be moved out of the bento files to allow for easier updating of the library.

glennguy:
FYI libewidevinecdm is opened through `OpenDRMSystem` here https://github.com/xbmc/inputstream.adaptive/blob/260384fbf87c426fe40457ba41e770442ad31006/src/main.cpp#L2305

>it started off separately as it needed to be shipped separate

Yes but it's not necessarily a bad thing to not have it coupled to the main program. From my point of view it allows a future libssd_pr.so or libssd_fp.so to be used.

Check the `GetSupportedDecrypterURN` function in main.cpp and you can see that it's flexible to whichever key system you want to use. A decryptor can also support one or many key systems.

Another way to think of it - is this a solution in search of a problem?

Would love to hear @peak3d and/or @liberty-developer thoughts on this

matthuisman:
@glennguy 
libwidevine is actually opened from the native_library files.
eg. https://github.com/xbmc/inputstream.adaptive/blob/Matrix/wvdecrypter/cdm/base/native_library_posix.cc#L25

i just think it'll simplify it a lot
No more needing to search for the decrypter for starters.
Surely it'll speed up loading it.
Might make debugging easier too?

If more decrypters are added, then just the same thing - have them built-in.
Not sure why they need to be separate binaries..
Especially when they are all shipped together anyway.

I'd just have them sit in a new "decrypter" directory in the inputstream.adaptive/src
eg. inputstream.adaptive/src/decrypter/widevine

I'm not too bothered either way, just seems like it would be a "nice to have"

CastagnaIT:
if in future will be more type of derypters files, the loading time (and then playback starts/stop) could be worse?
probably have separate binaries is to limit the overall size of the main file, but we don't have such large sizes...

glennguy:
@matthuisman I think this is still a 'to-do'.
[...]

PR #1236, opened 2023-04-24
link ssd_vw statically to avoid dlopen

lrusak:
## Description

This allows avoiding the shared library ssd_wv.so. We can link it statically to avoid the dlopen and symbol lookup. I'm not sure why it was originally like this (likely to allow loading some other library).

This also simplifies packaging as we don't have to search for ssd_wv.so when launching inputstream.adaptive.

When looking at the diff I recommend disabling whitespace changes as there is a few large shifts in indentation. https://github.com/xbmc/inputstream.adaptive/pull/1236/files?w=1

## Motivation and context

less complexity == better

## How has this been tested?

built and runtime tested (I didn't build or runtime test android, I'll rely on jenkins for building android).

## Types of change

## Checklist:


src/Session.cpp
```diff
@@ -117,69 +115,13 @@ void CSession::SetSupportedDecrypterURN(std::string& key_system)
...
-      m_dllHelper = std::make_unique<kodi::tools::CDllHelper>();
-      if (m_dllHelper->LoadDll(item.Path()))
-      {
-#if defined(__linux__) && defined(__aarch64__) && !defined(ANDROID)
-        // On linux arm64, libwidevinecdm.so depends on two dynamic symbols:
-        //   __aarch64_ldadd4_acq_rel
-        //   __aarch64_swp4_acq_rel
-        // These are defined in libssd_wv.so, but to make them available in the main binary's PLT,
-        // we need RTLD_GLOBAL. LoadDll() above uses RTLD_LOCAL, so we use RTLD_NOLOAD here to
-        // switch the flags from LOCAL to GLOBAL.
```
tmm1, line 151: @lrusak did you follow any of this work earlier?

Going forward here will mean aarch64 widevine breaks, which is something [we knew would happen](https://github.com/xbmc/inputstream.adaptive/pull/1130#issuecomment-1428293770)

Do you think it's feasible we could get [these two symbols](https://github.com/xbmc/inputstream.adaptive/pull/1130/files#diff-cb49b4f5a6e2a38d51a696be771430404f8a6c7f7012b92d5d27da461825f790) patched into the kodi build itself?

lrusak: why can't we just have them in `inputsream.adaptive.so` ?

tmm1: when we `dlopen(libwidevinecdm.so)` from this addon, `ld` searches the main binary's PLT (linker table) to resolve the missing symbols. so it expects these two symbols to be there.

however when kodi `dlopen(inputstream.adaptive.so)` like it does for any addon, it uses `RT_LOCAL` (which is the default on linux anyway) so any symbols defined in this addon will stay confined to the add-on and not make their way to kodi's PLT

the trick i was using here was that `RT_NOLOAD | RT_GLOBAL` will force export entries up. we could continue to use that trick, but that would involve finding the path to `inputstream.adaptive.so` itself and calling dlopen on that now.

tmm1: Our choices are basically:

- do nothing and expect the user/OS to LD_PRELOAD some library that defines the syms
- define these syms inside kodi source/binary itself
- define inside isa.so and use dlopen reload hack
- ship another .so with isa that contains only these symbols, that we then load with dlopen(RT_GLOBAL)

lrusak: is there an issue with using `RT_NOLOAD | RT_GLOBAL` when loading the add-on shared library from Kodi?

tmm1: using `RT_GLOBAL` for an add-on system would not be ideal, because it could mean functions defined in different addons could conflict and interact with each other.

on macOS `RT_GLOBAL` is the default, which is something [I had to change](https://github.com/xbmc/xbmc/pull/20219) because it affected how different versions of an add-on written in golang would conflict with each other when upgrading at runtime

lrusak: For the time being LD_PRELOAD seems like a simple solution. Maybe we can document a simple source code example, compilation instructions and usage?

CastagnaIT: as said if you remove it you are breaking widevine on all Kodi OS (LE/CE etc...), RPI OS and Linux distributions,
reading on forum threads the use of LD_PRELOAD is not a so easy solution for all OS
because they need to build a selfmade library to do a similar thing etc...

it is also complicated for users who do not know what this is to have to teach them to install a kind of preloadon their OS's to make Kodi works with widevine...

doing such a thing you must first discuss on slack at least with the LibreElec and CoreElec team, and also get more informations for RPI OS and in general for Linux OS

I don't know if this cleanup is worth doing it
given the expectations of removing this ssd library

the objective is to move all ssd code content to main project,
and keep this shared library only to contains the widevine aarch64 fix

where here i have attached a example scheme
https://github.com/xbmc/inputstream.adaptive/issues/736#issuecomment-1445342957

CastagnaIT:
closed since decrypters rework PR has been opened

PR #1238, opened 2023-04-25
[AdaptiveTree] Fix buffer data for goolge tests

CastagnaIT:
## Description

Fix buffer data for tests with related cleanups to `OnDataArrived`

the root of problem was that on `TestAdaptiveStream::DownloadSegment`
was missing the buffer `resize` then memcpy was failing
now i have semplified a bit

this fix the problem of the CI being blocked for hours

but another bug still exists that sometimes cause segmentation fault when running tests
the problem should always related to `AdaptiveStream::read` operations
but at the moment i have not found it

## Motivation and context

Sometimes google tests would get stuck in an endless loop
in the `AdaptiveStream::read` because the data was corrupted and can never arrive

## How has this been tested?

run google test dozens of times
## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -177,14 +177,12 @@ bool adaptive::AdaptiveStream::DownloadImpl(const DownloadInfo& downloadInfo,
             if (state_ == STOPPED)
               break;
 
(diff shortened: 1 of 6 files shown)


glennguy (approved):
Thanks for finding this, I thought all this time the issue was a problem in adaptivestream

PR #1239, opened 2023-04-26
CInputStreamAdaptive: use std::map for m_IncludedStreams

lrusak:
I'm not sure why this was an array but it made no sense because the index was an enum

## Types of change

## Checklist:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,7 +29,6 @@ using namespace UTILS;
 CInputStreamAdaptive::CInputStreamAdaptive(const kodi::addon::IInstanceInfo& instance)
   : CInstanceInputStream(instance)
 {
-  memset(m_IncludedStreams, 0, sizeof(m_IncludedStreams));
 }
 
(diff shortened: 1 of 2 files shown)


glennguy:
> I'm not sure why this was an array

We have no idea either unfortunately, but thank you for the improvement in readability.

glennguy (approved):

glennguy:
As a note, this needs to be reworked as there's a limitation on 1 of each type of stream being 'included streams'. I've seen examples of .ts files that have 1 video, 3 audio and 3 subtitles streams. In this scenario last one in wins. I don't think it would be a straight forward task though.

CastagnaIT:
seem have sense but i havent a sample to test it

lrusak:
Merge?

CastagnaIT:
it would be appropriate to do at least one test
but atm i dont have a sample stream for this use case, does anyone have one?

glennguy:
The SBS Akamai sample stream is an example of this - tested working fine here :)

PR #1241, opened 2023-04-29
[ADTS] - Initialise pts

glennguy:
## Description

Fixes starting ADTS streams beginning on 0 pts. Streams in this scenario will play fine if they are resumed from non-zero, in this case the first segment passed to the reader has a non-zero pts. But when starting from 0 we end up using the uninitialised value for pts.

## Motivation and context

Attempting to fix this issue here: https://github.com/aussieaddons/plugin.video.abc_iview/issues/3091
The cause of that issue is separate (malformed master playlist that has combined a/v .ts stream, but specifies non-existent separate audio stream) however an alternate is available which has separate a/v streams. Switching to the alternate stream revealed this issue.

## How has this been tested?

Tested against several problematic streams from the above provider

## Types of change

## Checklist:

### DIFF
--- a/src/ADTSReader.h
+++ b/src/ADTSReader.h
@@ -86,5 +86,6 @@ private:
   AP4_ByteStream *m_stream;
   ID3TAG m_id3TagParser;
   ADTSFrame m_frameParser;
-  uint64_t m_basePts, m_pts;
+  uint64_t m_basePts{0};
+  uint64_t m_pts;
 };


src/ADTSReader.cpp
```diff
@@ -132,8 +132,7 @@ bool ADTSFrame::parse(AP4_ByteStream *stream)
 /**********************************************************************************************************************************/
 
 
-ADTSReader::ADTSReader(AP4_ByteStream *stream)
-  : m_stream(stream)
+ADTSReader::ADTSReader(AP4_ByteStream* stream) : m_stream{stream}, m_basePts{0}
```
CastagnaIT: nice small fixes are the most obscure
but instead add here, please initialize vars by using his class source
https://github.com/xbmc/inputstream.adaptive/blob/72a4cc3610574b908738dc84c204ab4bc8400e66/src/ADTSReader.h#L89

glennguy: Done :)

CastagnaIT (approved):

PR #1244, opened 2023-04-29
[DASH] parse encryption scheme value

glennguy:
## Description

After the revert to single widevine session changes, it exposes a problem with the CDM that was the reason I originally went that way. It wasn't picked up in the change because I tested on a different source.

When setting up decryption and `GetCapabilities` is called in wvdecrypter the encryption mode in the CDM is set to cenc by default. This isn't an issue of itself when testing the license against dummy data but the CDM seems to not be able to handle changing modes once this first decryption is done. Of note, for some reason the CDM is fine with this scenario if there are 2 sessions??? Starting off at cenc and then changing to cbcs in the case of my old test stream will currently end up with decryption failing. When testing the single session PR, I used a live CMAF stream that was HLS - HLS is already set up to determine cenc/cbcs based on the `METHOD` value in the key. This PR gives the same capability for DASH streams. 

Example tag:
`<ContentProtection value="cbcs" schemeIdUri="urn:mpeg:dash:mp4protection:2011" cenc:default_KID="30303030-3030-3030-3030-303030303030"/>`

## Motivation and context

Fixes broken playback in DASH cbcs streams

## How has this been tested?

```
#KODIPROP:inputstream.adaptive.license_key=https://proxy.uat.widevine.com/proxy?provider=widevine_test||R{SSM}|
https://storage.googleapis.com/wvmedia/cbcs/h264/tears/tears_aes_cbcs.mpd
```

## Types of change

- [] **Bug fix** (non-breaking change which fixes an issue)
- [] **New feature** (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -1344,6 +1344,13 @@ bool adaptive::CDashTree::ParseTagContentProtection(pugi::xml_node nodeParent,
       xml_attribute attrKID = XML::FirstAttributeNoPrefix(nodeCP, "default_KID");
       if (attrKID)
         defaultKID = attrKID.value();
+
+      // get crypto mode if available
+      std::string_view protectionValue = XML::GetAttrib(nodeCP, "value");
+      if (protectionValue == "cenc")
+        m_cryptoMode = CryptoMode::AES_CTR;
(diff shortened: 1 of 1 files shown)


CastagnaIT (approved):
good catch

PR #1250, opened 2023-05-04
[PropertiesUtils] Add streams_config property

CastagnaIT:
## Description

This PR add a new property: `inputstream.adaptive.streams_config`
the name a bit generic to allow in future possibile new extensions of these configurations relative to streams without add new properties (if you want suggest a better naming say so)

the values are handled in similar way of a pair values (or something similar of python dict) as follow:
`key=value;`
more values are so concatenated:
`key=value;key2=value2;key3=value3;`

### `inputstream.adaptive.streams_config`
Supported configuration keys:
- `audio_langcode_default`: Set the "default" flag to the audio streams with specified language code (need to match the manifest)
- `audio_langcode_original`: Set the "original" flag to the audio streams with specified language code (need to match the manifest)
- `audio_prefer_stereo`: Can be set to `true` when you want prefeer stereo track as default instead of multichannels (it depends from audio_langcode_default and/or audio_langcode_original)
- `audio_prefer_type`: Can be set to `original` to prefer set default streams with original language, or `impaired` to prefer set default streams for impaired, otherwise left empty value for default behaviour.
- `subtitles_langcode_default`: Set the "default" flag to the subtitle streams with specified language code (need to match the manifest)

Example:
`audio_langcode_default=pt_br;audio_prefer_stereo=true`

## Motivation and context

Completed fix for #422 for DASH and in more general way,
after the recent parser rework i have improved the behaviour of  `AdaptationSet` tag ISA custom attributes:
`default`, `impaired`, `forced`, `original`
where now can be add and used to full override default manifest stream attributes,
the reason is that Kodi stream flags have not always have the same meaning of manifests attributes
and/or there are video services that dont follow exactly the specs and lead to wrong flags set to the audio/subtitle tracks

ofc add these custom ISA attributes needs to implement a proxy on a video add-on not so easy to do for novice devs but also expensive for low end devices, therefore this PR add two new properties that allow to set (and so override) the default stream language and original stream language in easy way

### HLS manifest note
This manifest type not always have appropriate language codes and also not always provide right number of audio channels in th metadata, therefore above configurations may fails, strictly depends on the service provider

### WIP - i will evaluate this part in the coming days:
i will need to make some more tests
use json for property values

## How has this been tested?

Tested with a sample dash

## Types of change

## Checklist:

### DIFF
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -10,7 +10,7 @@
     name="adaptive"
     extension=""
     tags="true"
-    listitemprops="license_type|license_key|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max|live_delay"
+    listitemprops="license_type|license_key|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max|live_delay|streams_config"
     library_@PLATFORM@="@LIBRARY_FILENAME@"/>
   <extension point="xbmc.addon.metadata">
     <platform>@PLATFORM@</platform>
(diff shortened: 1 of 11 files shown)


kino0924:
@CastagnaIT 

I have been looking for a way for this and looks great feature.
However, its not merged for long time.
Is there reason that its not merged or not closed?

I really need this feature since I have to set every dash stream that I have to different audio track since the track is not default/original

CastagnaIT:
this is marked Draft so WIP,
before this PR there was a need to make other code changes
this code also needs to be revisited because there were points that were not really clear

now i am busy finishing a more important change that has priority over this one,
but i have little free time in these months

anyway meantime for original flag you can try use https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptiveoriginal_audio_language

kino0924:
> this is marked Draft so WIP, before this PR there was a need to make other code changes this code also needs to be revisited because there were points that were not really clear
> 
> now i am busy finishing a more important change that has priority over this one, but i have little free time in these months
> 
> anyway meantime for original flag you can try use https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptiveoriginal_audio_language

Thank you for the reply.
And yes, I already tried original_audio_language and sadly it didnt work.
I really hope this feature gets implemented someday!

Thanks again.

CastagnaIT:
closed because superseeded by #2040

PR #1256, opened 2023-05-11
[Nexus][AdaptiveTree] Fix memory leak

glennguy:
Periods (and their children) are not being removed, this fixes that. Also observed has been some crashing which happens on DASH manifest refreshes when cloning the tree - may be related to the leaking...

## Description

Resolves memory leak present which is most noticeable on large/long live DASH manifests

## How has this been tested?

Tested on Kayo stream for some time

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -523,7 +523,11 @@ public:
 
   AdaptiveTree(CHOOSER::IRepresentationChooser* reprChooser);
   AdaptiveTree(const AdaptiveTree& left);
-  virtual ~AdaptiveTree() = default;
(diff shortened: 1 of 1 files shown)


CastagnaIT (approved):

matthuisman:
Great speedy work. Thanks

matthuisman:
Does the need to go into Omega as well?

glennguy:
> Does the need to go into Omega as well?

No I don't think so - in the rework periods are stored in AdaptiveTree as smart pointers so should be deleted automatically: `std::vector<std::unique_ptr<PLAYLIST::CPeriod>> m_periods;`

PR #1262, opened 2023-05-17
Don't start update thread on DASH updates

glennguy:
## Description

~~Quick test for recent issues regarding buffering/picture corruption~~ Users are reporting improvements over 20.3.8... would prefer to see with my own eyes but I think something is better than nothing, and this corrects something that shouldn't be happening regardless.

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveTree.h
+++ b/src/common/AdaptiveTree.h
@@ -536,7 +536,7 @@ public:
   virtual void Configure(const UTILS::PROPERTIES::KodiProperties& kodiProps);
 
   virtual bool open(const std::string& url) = 0;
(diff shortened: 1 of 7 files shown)


CastagnaIT (commented):
we can avoid interface changes
and just move the call of `StartUpdateThread` as last operation of `PostOpen`:
https://github.com/xbmc/inputstream.adaptive/blob/Omega/src/common/AdaptiveTree.cpp#L67-L76
other manifest type will be noop because dont update the required variables to allow update thread to start (HLS has his different implementation)

i remember you that this pr should be a backport, miss the Omega one

PR #1264, opened 2023-05-22
[FragmentedSampleReader] Fix memory leak in AP4_LinearReader::Process…

deGerard:
## Description
Fix a memory leak in CFragmentedSampleReader::ProcessMoof where there is a decrypter object is constructed via AP4_CencSampleDecrypter::Create but it is never deconstructed.

## Motivation and context
If I'm right the m_singleSampleDecryptor is always pointing to something at this point and after checking what the AP4_CencSampleDecrypter::Create actual does when m_singleSampleDecryptor exist is checking the IV length corresponding to the given cipher type and then construct the decrypter object what is not used.
I removed this checks because this should already been done for m_singleSampleDecryptor and the decrypter object isn't constructed anymore.
I guess this patch will not break anything. But if there is change that AP4_CencSingleSampleDecrypter::Create was called with m_singleSampleDecryptor == NULL then it look like it running into a down-casting of single_sample_decrypter in the Bento4 code before this patch.

## How has this been tested?
Running the 'make test' and check if I can still watch some content with widevine DRM.

## Types of change

## Checklist:

### DIFF
--- a/src/samplereader/FragmentedSampleReader.cpp
+++ b/src/samplereader/FragmentedSampleReader.cpp
@@ -421,13 +421,9 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
         // we assume unencrypted fragment here
         goto SUCCESS;
 
-      AP4_CencSampleDecrypter* decrypter = nullptr;
-      if (AP4_FAILED(result = AP4_CencSampleDecrypter::Create(sample_table, algorithm_id, 0, 0, 0,
-                                                              reset_iv, m_singleSampleDecryptor,
-                                                              decrypter)))
-      {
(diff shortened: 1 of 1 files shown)


glennguy:
Thanks for picking this up @deGerard, I think you're correct in all of the above.

I completely missed that the block being removed is totally redundant now - as you say we're bypassing the IV checks at https://github.com/axiomatic-systems/Bento4/blob/8c08fa19f841cde24b990bbe05177bfcda732e56/Source/C%2B%2B/Core/Ap4CommonEncryption.cpp#L2041-L2090 but I don't think it matters. In wvdecrypter we assume an IV size of 16 anyway.

glennguy (approved):

PR #1267, opened 2023-05-28
[AdaptiveTree] Rework to autodetect manifest type

CastagnaIT:
## Description

Time ago i have started to do this rework to allow ISA autodetect the manifest type this will allow in future to full remove the current mandatory property `inputstream.adaptive.manifest_type`

The changes has been based by taking inspiration from exoplayer so detection should have a high success rate

Summary changes:
* Autodetect manifest type
* `open` method on parsers now just "open" the data for parsing, then download methods has been completely removed from adaptive tree and moved as helper method in curl utils
* As related code, this also fix problem where the representation chooser speed is updated also when we download manifest updates during playback or when HLS parser use AdaptiveTree::Download to download KID data
* As related code, has been deprecated the hack/custom `$START_NUMBER$` placeholder in manifest url, used for DASH only seem for youtube addon only, the `manifest_update_parameter` kodi property do the same, so there no reasons to kept both ways, we simplify sourcecode, addons have a very little change to do (ps. this placeholder support was not documented in wiki)

SIDENOTE:
- the `CDashTree::RefreshLiveSegments` method looks a lot changed due to indentation change, my changes are limited until to `updateTree->Open` call
- with `inputstream.adaptive.manifest_update_parameter` its possible, for now, set parameters to the url for manifest update request, this is a temporary way for a followup PR, the goal is decouple `manifest_update_parameter` with an appropriate property to set parameters for the manifest update, and a property to force enable the manifest update, i would like in future remove the "full" behaviour parameter, we should fix our broken Live dash use cases

## Motivation and context

Detect the manifest type should be automatic and not a task of addons

## How has this been tested?

tested STRM with dash, hls, and ism with utf16
tested with addon that use proxy to manipulate manifest
tested with dash on am@zon vod

## Types of change

## Checklist:


src/parser/HLSTree.h
```diff
@@ -52,6 +57,22 @@ class ATTR_DLL_LOCAL CHLSTree : public AdaptiveTree
                                PLAYLIST::StreamType type) override;
 
 protected:
+  /*!
+   * \brief Download KID from media initialization section, overridable method for test project
+   */
+  virtual bool DownloadKID(std::string_view url,
```
glennguy: I don't know if DownloadKID (Key ID) is the correct name for downloading AES keys, maybe DownloadKey?

src/common/AdaptiveTreeFactory.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "../parser/HLSTree.h"
+#include "../parser/SmoothTree.h"
+
+using namespace PLAYLIST;
+using namespace UTILS;
+
+namespace
+{
+
+PROPERTIES::ManifestType InferManifestType(std::string_view url,
```
glennguy: Just a comment for later - we should introduce some tests for this.

glennguy (changes requested):

src/common/AdaptiveTreeFactory.cpp
CastagnaIT: i will try take a look to include it here

src/parser/HLSTree.h
CastagnaIT: done

src/common/AdaptiveTreeFactory.cpp
CastagnaIT: tests added

glennguy: Awesome, thanks!

PR #1270, opened 2023-05-29
[ADTS] add support for AC3 and E-AC3

### LINKED ISSUE #1085: HLS: .ec3 file extension for audio doesn't work (plays ok in ffmpeg)
more info to come....

image

glennguy:
## Description

Adding support for parsing ADTS encapsulated AC3 and E-AC3 audio streams. The Bento4 library already has built in tools for parsing all of these formats however the current implementation in IA has rolled it's own support for AAC. I removed that code and extended the Bento4 implementation to support all 3.

Why is extending Bento4 needed?
Bento4 implementation assumes that we are reading an entire file and doesn't work with the segmented streaming approach we take. The frame parsers want to read the entire frame plus the header of the next frame. When doing this we will end up reading the start of the next segment in IA. At this point we can't rewind the stream head to fill the buffer with the data, and end up losing the last data frame of each segment.
The extended classes just omit this part of the Bento4 implementation - we're only interested in parsing the header itself so we can populate the metadata for passing on to ffmpeg.

## Motivation and context

When we have these files (usually .ac3 / .ec3 file extensions) our reader doesn't know what to do and is only supporting the AAC format.
Fixes #1085 

## How has this been tested?

Tested with:
```
https://storage.googleapis.com/shaka-demo-assets/apple-advanced-stream-ts/master.m3u8
```
which contains aac, ac3 and eac3 audio streams.
Audio group IDs:  aud1 == AAC, aud2 == AC3, aud3 == EAC3

IA will always choose EC3 for this playlist however I made small changes to HLSTree.cpp for testing:

https://github.com/xbmc/inputstream.adaptive/blob/f8c43ab5f6746dfbd883a24151a78a92bc8706d9/src/parser/HLSTree.cpp#L829-L830 becomes
```
      if (attribs["TYPE"] == "AUDIO") 
      {
        if (attribs["GROUP-ID"] == "aud2" || attribs["GROUP-ID"] == "aud3")
        {
          continue;
        }
        streamType = StreamType::AUDIO;

      }
```
This will test AAC. To AC3, replace `aud2` with `aud1` .

## Types of change

## Checklist:


glennguy:
Will need Bento4 release on xbmc and update here before merge.

glennguy:
Another note - AC4 support can be easily added however this isn't in ffmpeg yet.

src/ADTSReader.cpp
```diff
@@ -73,62 +78,141 @@ ID3TAG::PARSECODE ID3TAG::parse(AP4_ByteStream *stream)
...
+  AP4_Ac3Frame frame;
+  AP4_Size sz = buffer.GetDataSize();
+  parser.Feed(buffer.GetData(), &sz);
+  AP4_Result result = parser.FindFrameHeader(frame);
+  if (!AP4_SUCCEEDED(result))
+    return false;
+
+  m_totalSize = frame.m_Info.m_FrameSize;
+  m_sampleRate = frame.m_Info.m_SampleRate;
+  m_channelConfig = frame.m_Info.m_ChannelCount;
```
CastagnaIT: look like `m_channelConfig` is a wrong name,
has sense rename to `m_channelCount`?

src/utils/CodecParser.h
```diff
@@ -0,0 +1,54 @@
...
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include <bento4/Ap4Ac3Parser.h>
+#include <bento4/Ap4AdtsParser.h>
+#include <bento4/Ap4Eac3Parser.h>
+
```
CastagnaIT: what do you think to move these new files (CodecParser.*) to `/parser` folder instead of utils?
and also suggest to wrap all header code with an appropriate namespace

src/utils/CodecParser.h
```diff
@@ -0,0 +1,54 @@
...
+#define AP4_ADTS_HEADER_SIZE 7
+#define AP4_ADTS_SYNC_MASK 0xFFF6
+#define AP4_ADTS_SYNC_PATTERN 0xFFF0
+#define AP4_AC3_SYNC_PATTERN 0x0B77
+#define AP4_AC4_SYNC_MASK 0xFFF0
+#define AP4_AC4_SYNC_PATTERN 0x0AC40
```
CastagnaIT: new code should not use defines, but for example
`constexpr AP4_UI32 AP4_ADTS_SYNC_PATTERN = 0xFFF0;`

src/utils/CodecParser.h
```diff
@@ -0,0 +1,54 @@
...
+enum class AdtsType
+{
+  NONE,
+  AAC,
+  AC3,
+  EAC3,
+  AC4
+};
+
+class CAdaptiveAdtsParser : public AP4_AdtsParser
```
CastagnaIT: on all classes miss `ATTR_DLL_LOCAL` or is intentional?

src/utils/CodecParser.cpp
```diff
@@ -0,0 +1,249 @@
...
+  if (AP4_FAILED(result))
+    return result;
+
+  /* parse the header */
+  AP4_AdtsHeader adts_header(raw_header);
+
+  /* check the header */
+  result = adts_header.Check();
+  if (AP4_FAILED(result))
+    goto fail;
```
CastagnaIT: single `goto` can be removed

CastagnaIT: also the others below

CastagnaIT (commented):
nice thing, for now not tested just look at code

please do bento4 bump to a separate commit

glennguy:
Thanks for the review! I've made all suggested changes except for the bento4 bump - I'll fix that before merge since I need to commit the changes in bento4 repo first.

src/parser/CodecParser.cpp
```diff
@@ -0,0 +1,233 @@
...
+  }
+  AP4_Position currentPos;
+  stream->Tell(currentPos);
+  stream->Seek(currentPos - (AP4_EAC3_HEADER_SIZE));
+  return adtsType;
+}
+
+AP4_Result CAdaptiveAdtsParser::FindFrameHeader(AP4_AacFrame& frame)
+{
+  unsigned int available;
```
CastagnaIT: this var `unsigned int available;`
is declared but unused also on other method below, please delete it

all other changes ok
just remember to make a separate commit for bento ver bump

CastagnaIT (approved):

Issue #1197, opened 2023-03-20
Choppy 4K HDR10(+) video on Android TV in Amazon VOD addon


MauriceW67:
### Describe the bug
Here is a clear and concise description of what the problem is:

When playing 4K HDR10(+) video from the Amazon VOD addon, the video is very choppy. If I turn off the HDR10+ option in the addon, the 4K version plays fine in SDR HEVC. Tested with various episodes of The Boys Season 3 and Reacher Season 1.

Note 1: 4K Dolby Vision plays fine (tested with Lord of the Rings: The Rings of Power).

Note 2: This does not only happen on my Philips TV, but also on the Nvidia Shield Pro 2019. And it also happens on Kodi 20.1 Nexus, so not just on Kodi 21 master.

Note 3: @CastagnaIT asked if the debug info shows any corrections. The answer is no. During the choppy video the skip and audio correction counters don't increase. The audio does have some dropouts, but not at every video skip though.

Note 4: This issue was first created in the Amazon VOD GitHub, but @glennguy has requested me to create it here as well.

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

4K HDR10+ videos playing smoothly in the Amazon VOD addon.

## Actual Behavior

4K HDR10+ videos play very choppily.

### To Reproduce

1. Enable the HDR10+ option in the Amazon VOD addon
2. Play a TV show episode that is available in HDR10+ such as The Boys season 3 or Reacher season 1
3. Notice the choppy video

### Debuglog

The debuglog can be found here:
Debug log of 4K HDR10+ video playing: https://paste.kodi.tv/azeviduwov.kodi (part 1) and https://paste.kodi.tv/wojudozeko.kodi (part 2)
Debug log of same episode playing in 4K SDR: https://paste.kodi.tv/bajihivozu.kodi

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
Manifest files were too big for Kodi pastebin, so I zipped them up here:
https://drive.google.com/file/d/1KhVRliJaModP5-fGibdxO42N107iMBO1/view?usp=share_link

[manifest_1679331240_hdr10.txt](https://github.com/xbmc/inputstream.adaptive/files/11108925/manifest_1679331240_hdr10.txt)
[manifest_1679334649_sdr.txt](https://github.com/xbmc/inputstream.adaptive/files/11108928/manifest_1679334649_sdr.txt)

### Screenshots 
[...]

PR #1279, opened 2023-06-05
[AdaptiveTree] Fix pointer move and reintroduced audioTrackId

CastagnaIT:
## Description

With recent reworks in the AdaptiveTree::SortTree the std::move used to move all representations from an adpSet to another one was working correctly but since the "starting" adpSet will be at the end deleted, the representations still have members that point to the deleted adpSet, and this was causing the crash

I have add a "todo" note to AdaptiveTree::SortTree because i am a bit concerned about this "merge", i suspect that this is wrong thing to do since adpsets may have different data in childrens (e.g. dash xml) and on others manifests may be not useful to do,
if you have some details please specify them because the purpose of the merge is not completely clear and git history dont provide clear details

For dash case i have reintroduced `audioTrackId`, this is specific am@zon case only, since we "merge" adpSets the only way to distinguish limit cases are by using the `audioTrackId`, in case we remove or rework the adpSet "merge" behaviour, we can also consider to remove again the `audioTrackId` parsing.
An example is on the attached mpd where there are two audio streams with same "lang" code `es`, but declared with different region code on audioTrackId.

## Motivation and context

Final fixes for #1197
## How has this been tested?

Played am@zon 4k hdr streams

Example mpd for audioTrackId: [manifest_1685902687.txt](https://github.com/xbmc/inputstream.adaptive/files/11649211/manifest_1685902687.txt)

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveTree.cpp
+++ b/src/common/AdaptiveTree.cpp
@@ -229,6 +229,13 @@ namespace adaptive
       auto& periodAdpSets = period->GetAdaptationSets();
 
       // Merge VIDEO & AUDIO adaptation sets
+      //! @todo: seem that merge adpsets is this not safe thing to do, adpsets may have different encryptions
+      //!        and relative different child data (e.g. dash xml child tags)
+      //!        it is needed to investigate if we really need do this,
+      //!        if so maybe limit for some use cases or improve it in some way.
+      //!        audio merging has been impl years ago without give any details of the reasons or for what manifest types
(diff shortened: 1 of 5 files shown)


glennguy (approved):
Thanks @CastagnaIT 
Off the top of my head, I think that the merge function is to clean up HLS manifests where there are multiple pairs of audio/video due to different audio codecs...

PR #1292, opened 2023-06-14
[AdaptiveTree] First rework to cleanup segment code

### LINKED ISSUE #1129: Live DASH stream fails to play
### Describe the bug

When attempting to play certain live DASH streams in Kodi Nexus, IA fails to play, and Kodi crashes. The same manifest plays fine in other players, e.g. VLC.

## Expected Behavior

Play the stream.

### To Reproduce

Try playing https://ve-dash-uk.live.cf.md.bbci.co.uk/pool_901/live/uk/silver_stream_02/silver_stream_02.isml/pc_hd_abr_v3_dash.mpd
Attention: Stream is not permanently available and geolocked to the UK.

### Debuglog

### MPD/M3U8s/ISM

https://ve-dash-uk.live.cf.md.bbci.co.uk/pool_901/live/uk/silver_stream_02/silver_stream_02.isml/pc_hd_abr_v3_dash.mpd

[pc_hd_abr_v3_dash.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/10760417/pc_hd_abr_v3_dash.mpd.txt)

### Screenshots 

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [x] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Debian testing
 - Kodi version: 20.0.0 (20230116)
 - inputstream.adaptive version: 20.3.2

CastagnaIT:
## Description

This is a first PR to start cleanup all code related to segments which is all so messy and barely readable

Some main changes:
- The `CSegment` `range_begin_` `range_end_` now contains the exact data, and is no longer crazily mixed with PTS and numbering
- Has been removed the `segNum` code hack in the `AdaptiveStream::PrepareDownload`
- Has been removed the hackish way of `SEGMENTBUFFER::segment_number` to detect initialization segment by using `~`
- The weird use of `CRepresentation::initialization_` with separate statement `m_hasInitialization` has been removed
- Now the `AdaptiveStream::PrepareDownload` is responsable to build the url where download the segment and there is no more scattered code with parsers
- `CSegmentTemplate` now have the old called "placeholder replace" in its place with appropriate implementation to format template identifiers

Note 1: i have add `m_time` and `m_number` to `CSegment` in temporary way, the reason is that the existing old code couples the timeline with the segments, this should usually be separated, to do so will require many more changes to the code and will not be addressed here

Note 2: `CSegment` `m_time` looks almost equal to `startPTS_`, maybe is possible merge both members (separate PR?)

Note 3: Most of Dash tests in the TestDASHTree.cpp has been readapted for the new changes, since the tests wants know the final url from segment template we need to run "stream test" to generate and get the segments urls used for downloads

~Note 4: There is some HLS code stuff in the HLSTree.cpp that needs a double check by someonelse, looks to changed/commented code~ restored propagation across new periods of init segment  

## Motivation and context

A first attempt to solve some old code hacks
and fix #1129

## How has this been tested?

Tests

DASH: tested couple of LIVE samples, and some VOD samples
HLS: tested akmai local test (chapters), LIVE RAI, LA7

## Types of change

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,7 @@ set(ADP_SOURCES
 	src/common/Representation.cpp
 	src/common/ReprSelector.cpp
 	src/common/Segment.cpp
+	src/common/SegmentBase.cpp
 	src/common/SegmentList.cpp
 	src/common/SegTemplate.cpp
 	src/parser/CodecParser.cpp
@@ -103,6 +104,7 @@ set(ADP_HEADERS
 	src/common/Representation.h
 	src/common/ReprSelector.h
 	src/common/Segment.h
+	src/common/SegmentBase.h
 	src/common/SegmentList.h
 	src/common/SegTemplate.h
 	src/parser/CodecParser.h
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1004,7 +1004,7 @@ AP4_Movie* CSession::PrepareStream(CStream* stream, bool& needRefetch)
   }
 
   if (repr->GetContainerType() == ContainerType::MP4 && !repr->HasInitPrefixed() &&
-      !repr->HasInitialization())
+      !repr->HasInitSegment())
   {
     //We'll create a Movie out of the things we got from manifest file
     //note: movie will be deleted in destructor of stream->input_file_
--- a/src/common/AdaptationSet.cpp
+++ b/src/common/AdaptationSet.cpp
@@ -64,6 +64,7 @@ void PLAYLIST::CAdaptationSet::CopyHLSData(const CAdaptationSet* other)
     m_representations.push_back(std::move(rep));
   }
 
+  m_baseUrl = other->m_baseUrl;
   m_streamType = other->m_streamType;
   m_duration = other->m_duration;
   m_startPts = other->m_startPts;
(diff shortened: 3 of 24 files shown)


CastagnaIT:
@glennguy when you will do the review please pay attention for "Note 4" above, maybe you can check if it is ok, i will remove the commented code just after the review

glennguy:
Hi Stefano
On note 4 it looks good to me. Previous code is just to keep track of the current init segment across periods. HLS spec says that we should keep the same init seg on next period unless another or none is specified however I have *never* seen an example where the packager didn't explicitly tell the init segment for the next period if it was the same. You'd only expect to have that situation of same init segment for example having a DAI stream that didn't have ads inserted inbetween periods (e.g akamai test stream - but that's TS not mp4 anyway).

On note 2 - happy to look at later if that's the way you want to go.

glennguy (approved):

CastagnaIT:
now understand so for new period is intended EXT-X-DISCONTINUITY then after this tag we could parse a new EXT-X-MAP with a changed init segment, mmh i think that despite its a uncommon use worth not remove this behaviour
i can use `CopyHLSData` to copy the previous repr init segment to semplify things

edit: changes done

PR #1293, opened 2023-06-14
[DASHTree] Dont force rounding segment count

CastagnaIT:
## Description

the calculation of segment count was forced rounded by always add +1
attached sample stream show that we are generating a surplus segment that on playback try download without success because not existent

looking at similar calculation of exoplayer their use ceil to round the result value
https://github.com/google/ExoPlayer/blob/ac9d5337b2b51a855f3c33f3b126d9ef921ebc69/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java#L459-L463

## Motivation and context

i was trying play following sample stream
https://ftp.itec.aau.at/datasets/mmsys18/testing/Beauty_4sec/multi-codec.mpd
that show different playback weirdness that make me think there are more little bugs to check, eg.:
- fails to download last segment but playback result broken, at differents points of playback!? I would expect to playback until the last downloaded segment despite the last segment is broken (could be due to the AdaptiveStream worker that stop operations at first broken segment)
- fails to download last segment but sometime reach the end of duration, but playback continues without stop...
- after this PR fix, playback works but when reach the end of video duration, the kodi GUI progress time bar go to "0" then stop

## How has this been tested?

https://ftp.itec.aau.at/datasets/mmsys18/testing/Beauty_4sec/multi-codec.mpd
## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -19,6 +19,7 @@
 #include "pugixml.hpp"
 
 #include <algorithm> // max
+#include <cmath>
 #include <cstdio> // sscanf
 #include <numeric> // accumulate
 #include <string>
@@ -1037,7 +1038,7 @@ void adaptive::CDashTree::ParseTagRepresentation(pugi::xml_node nodeRepr,
         double lengthSecs =
             static_cast<double>(segTemplate->GetDuration()) / segTemplate->GetTimescale();
         segmentsCount =
-            static_cast<size_t>(static_cast<double>(reprTotalTimeSecs) / lengthSecs + 1);
+            static_cast<size_t>(std::ceil(static_cast<double>(reprTotalTimeSecs) / lengthSecs));
       }
 
       if (segmentsCount < 65536) // SIDX atom is limited to 65535 references (fragments)
--- a/src/test/TestDASHTree.cpp
+++ b/src/test/TestDASHTree.cpp
@@ -244,9 +244,9 @@ TEST_F(DASHTreeTest, CalculateCorrectSegmentNumbersFromSegmentTemplateWithPTO)
 
   auto& segments = tree->m_periods[0]->GetAdaptationSets()[0]->GetRepresentations()[0]->SegmentTimeline();
 
-  EXPECT_EQ(segments.GetSize(), 451);
+  EXPECT_EQ(segments.GetSize(), 450);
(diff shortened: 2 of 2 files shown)


CastagnaIT:
`CalculateCorrectSegmentNumbersFromSegmentTimelineWithPTO` test fails
i wonder if the value set in the test was purely a copy-paste of `segments.GetSize()` result, it would be better to check with a live stream

following sample looks similar to the failed test: https://livesim.dashif.org/livesim/periods_20/testpic_2s/Manifest.mpd
and seem to works correctly

but shown another problem not related to this, so we are downloading many times the manifest update
should be due to last mdp update code changes that i have done with recent PR, 
I think it revealed a lack in ISA where miss appropriate mpd update implementation for DASH,
there are differents update behaviours that need to be considered by using `minimumUpdatePeriod`
https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#timing-mpd-mup
in fact comment out lines https://github.com/xbmc/inputstream.adaptive/blob/Omega/src/common/AdaptiveStream.cpp#L811-L812 "solve" the situation, i will try take a look afterwards

glennguy:
> i wonder if the value set in the test was purely a copy-paste of segments.GetSize()

I can't remember exactly but I think that the test was added to make sure that the value didn't deviate when changing code, rather than making sure it was right in the first place.

Looking at the TSB length of 1800 and segment length of 4 though it should be 450, so I think the changes here are correct.

> think it revealed a lack in ISA where miss appropriate mpd update implementation for DASH

Yeah I never understood that code - came from GSOC student/peak3d. Seems maybe to be a way to ensure it always updates mpd on each segment? 
Agree that we should follow the DASH specs to avoid unrequired refreshes.

glennguy (approved):

glennguy:
Happy for you to merge but maybe best to change the expected test results?

CastagnaIT:
> Yeah I never understood that code - came from GSOC student/peak3d. Seems maybe to be a way to ensure it always updates mpd on each segment?
> Agree that we should follow the DASH specs to avoid unrequired refreshes.

it seem so and it is fine i think, but we need to implement a way to use or disable it according to the manifest configuration/use cases

i will fix the failed test case later

PR #1302, opened 2023-06-28
Fix ways to update manifest and intervals

CastagnaIT:
## Description

After recent rework/cleanup with PR #1288 critical issues were highlighted in the current implementation of the manifest update code, where mainly our existing code use bad timing intervals, dont manage correctly manifest updates (for DASH use cases) and also can mixes update types resulting download too much times same file

This is a fast fixup to allow at least find a kind of solution with existing code without introduce a big expensive rework

whats done:
`AdaptiveTree::m_updateInterval` has three meaning as dash specs:
> A nonzero value for MPD@minimumUpdatePeriod defines the [MPD validity duration](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#mpd-validity-duration) of the present snapshot of the [MPD](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#mpd), starting from the moment its download was initiated. This allows the service to provide regular updates to the [MPD](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#mpd) while limiting the refresh interval to avoid overload.
> 
> The value 0 for MPD@minimumUpdatePeriod indicates that the [MPD](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#mpd) has no validity after the moment it is retrieved. In such a situation, the client SHALL acquire a new [MPD](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#mpd) whenever it wants to make new [media segments](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#media-segment) available (no "natural" state changes will occur due to passage of time).
> 
> Absence of the MPD@minimumUpdatePeriod attribute indicates an infinite validity (the [MPD](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#mpd) will never be updated).

this allow distinguish what type of manifest update is needed then and then use or not the update thread

what missing for an appropriate manifest update (require future reworks):
- Timing interval used in the update thread should be dynamic, so we have the need to implement a method that allow to change the interval value, two reasons:
   - for HLS case the best should be take the last segment duration time and use that "time" value to download the next manifest update
   - for HLS case in the case the downloaded manifest dont contains changes on segments, take the last segment duration time divide time in half, and use that time to download the next manifest update (after restore the right timing)
- implement for DASH `UTCTiming` tags support is required to synchronise manifest updates

i will add a link to what said above to project roadmap

## How has this been tested?

Live tests only

- Dash
LA7
Facebook
https://livesim.dashif.org/livesim/segtimeline_1/testpic_2s/Manifest.mpd
https://livesim.dashif.org/livesim/ato_10/testpic_2s/Manifest.mpd

- HLS
RAI
dailymotion

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -782,7 +782,7 @@ bool AdaptiveStream::ensureSegment()
     // lock live segment updates
     std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(tree_.GetTreeUpdMutex());
 
-    if (tree_.HasManifestUpdates() && SecondsSinceUpdate() > 1)
+    if (tree_.HasManifestUpdatesSegs() && SecondsSinceUpdate() > 1)
     {
       tree_.RefreshSegments(current_period_, current_adp_, current_rep_, current_adp_->GetStreamType());
       lastUpdated_ = std::chrono::system_clock::now();
@@ -926,7 +926,8 @@ bool AdaptiveStream::ensureSegment()
         return false;
       }
     }
-    else if (tree_.HasManifestUpdates() && current_period_ == tree_.m_periods.back().get())
+    else if ((tree_.HasManifestUpdates() || tree_.HasManifestUpdatesSegs()) &&
+             current_period_ == tree_.m_periods.back().get())
     {
       if (!current_rep_->IsWaitForSegment())
       {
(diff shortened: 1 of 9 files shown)


CastagnaIT:
@glennguy i have not found weirdness on playback but if you have some other live streams to test its better

glennguy (approved):
All makes sense to me. Thanks @CastagnaIT!

PR #1323, opened 2023-07-11
Fixes for buffer and recent regressions

### LINKED ISSUE #1295: Since ISA v21.1.1, some high bitrate content has abnormal memory usage that can cause Kodi to crash
### Describe the bug

Stream some new episodes(?) such as THE DAY, Black Mirror Season 6 using the Netflix add-on. Playback suffers severe frame drops for 1-2 minutes and eventually freezes.
It's worth noting that the problem seems to only occur at the highest bitrate. Maybe it is necessary to reach a certain bit rate. So we need to manually select the highest bit rate video stream when playing.
ISA v21.1.2 didn't change anything, even worse I think.

## Expected Behavior

Smooth video playback

## Actual Behavior

Severely dropped frames that don't recover over time and eventually freeze.

## Possible Fix

Apparently this is a problem introduced since v21.1.1, because if I use v21.1.0, this problem doesn't happen.

### To Reproduce

1.Open Netflix add-on
2.Play specific content, such as Black Mirror Season 6
3.Choose 4K, **best bitrat**e to play.

### Debuglog

https://paste.kodi.tv/nufepuyara.kodi

### MPD/M3U8s/ISM

https://paste.kodi.tv/uruwirupiy

### Screenshots 

### Your Environment
Used Operating system:Shield TV Pro 2019

 - [X] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name:9.1.1(Android TV 11)
 - Kodi version:Kodi 21 20230613-2307e4d6
 - inputstream.adaptive version:v21.1.1

CastagnaIT:
## Description

Fixes found from slack discussion

One of the problems of high memory occupation
is due to
https://github.com/xbmc/inputstream.adaptive/blob/21.1.2-Omega/src/common/AdaptiveStream.cpp#L617-L638
where we set:
assured_buffer_length_ to 60
max_buffer_length_ to 120

in the old code:
https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/common/AdaptiveStream.cpp#L541-L550
instead set:
assured_buffer_length_ to 4
max_buffer_length_ to 8

this implies we have a very big buffer that require much memory so increase a lot until the end

i think it is better to completely disable this bad code that we have been dragging around for years
now we have almost stable memory occuped, ~but we can see about 100 mb increased on start~ (seem regular its debug vs release build)
there is a little memory that increase of about 1mb each time after some seconds that could require other investigations but could be a problem at kodi side (its present also on old isa version)

this bug also reveals another problem on buffering
when we put the playback on pause, we continue download segments where should be paused and resumed when playback will be resumed, this can be improved later, until now this problem was hidden because the buffer is short

Commit: Fallback to ADTS sample reader when TS initialization fails
fix the audio with youtube HLS manifests
where we have wrong container, use file extension is not reliable,
tested with HLS version of https://www.youtube.com/watch?v=Hf5enZVznC4

## Motivation and context

fix #1295

## Types of change

## Checklist:

### DIFF
--- a/inputstream.adaptive/resources/settings.xml.in
+++ b/inputstream.adaptive/resources/settings.xml.in
@@ -174,11 +174,13 @@
         <setting id="ASSUREDBUFFERDURATION" type="integer" label="30200" help="30167">
           <level>1</level>
           <default>60</default>
+          <visible>false</visible> <!-- Working code disabled, rework needed -->
           <control type="edit" format="integer" />
         </setting>
         <setting id="MAXBUFFERDURATION" type="integer" label="30201" help="30167">
           <level>1</level>
           <default>120</default>
+          <visible>false</visible> <!-- Working code disabled, rework needed -->
           <control type="edit" format="integer" />
         </setting>
         <setting id="MEDIATYPE" type="integer" label="30112">
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -998,14 +998,14 @@ bool AdaptiveStream::seek(uint64_t const pos)
   // we seek only in the current segment
   if (state_ != STOPPED && pos >= absolute_position_ - segment_read_pos_)
   {
-    segment_read_pos_ = static_cast<uint32_t>(pos - (absolute_position_ - segment_read_pos_));
+    segment_read_pos_ = static_cast<size_t>(pos - (absolute_position_ - segment_read_pos_));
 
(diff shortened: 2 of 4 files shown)


glennguy (approved):
Thanks for pursuing this until we found an answer!

PR #1325, opened 2023-07-13
[SmoothTree] Regression fixes and better logging

### LINKED ISSUE #1324: Live ism streams are broken since version 21.1.1
### Describe the bug

The live streams which use ism manifests in the [plugin.video.orange.spain](https://github.com/Paco8/plugin.video.orange.spain) addon are not working since inputstream adaptive 21.1.1.

With some tv channels there's only audio, no video. With other channels there's audio only for some seconds, then there's video for some seconds (no audio) and then playback stops.

It works correctly with inputstream adaptive 21.1.0.

## Expected Behavior

The tv channels should play correctly, with video and audio.

## Actual Behavior

Only audio or audio for some seconds, then video for some seconds, then stops.

### To Reproduce

I can't provide a strm to test, because the token expires after 2 minutes.

### Debuglog

https://paste.kodi.tv/erivezazej.kodi

In that log I play two tv channels, the first one plays audio only, the second one plays audio for some seconds, then video for some seconds, then stops.

### MPD/M3U8s/ISM

[manifests.zip](https://github.com/xbmc/inputstream.adaptive/files/12015066/manifests.zip)

### Screenshots 

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Android 12
 - Kodi version: 21.0-ALPHA2
 - inputstream.adaptive version: 21.1.2

CastagnaIT:
## Description

Fixes:
- basetime was not set correctly with the results that lead to have wrong start PTS, results in a broken video
- ProtectionHeader can have SystemID wrapped by curly brackets `{ -- }` i did not find this in the specs

Other:
- Some know unsupported features now will be skipped and printed to the log
- Impaired subtitle tracks flag is now considered
- Some better log prints
- Add new smoothstreaming  test `CheckAsyncTimelineStartPTS` to test live manifest with different start timestamp
- `CSegmentTemplate::FormatUrl` has been reworked in order to keep possible single `$` chars along the url address, tests has been added on test project

Last changes:
imo `CSession::CheckFragmentDuration` method was a bit confusionary because was mixing implemetation between all manifest types and was not clear its functioning so i have full reworked it, now the method has been renamed to `OnSegmentChangedRead` that clear explain when its executed, and `InsertLiveSegment` method now require to be implemented on each manifest type, if any/supported

`SampleReader::GetFragmentInfo` called by `CSession::OnSegmentChangedRead` take info only from fmp4 sample table "stbl"->"stts" for the timestamp duration and the implementation in SR is no longer messed with anything else.

On this last thing its not clear to me because our code read the "stbl"->"stts" each time for each new segment, i could not find any documentation that says that the atom "stbl" its present on each segment, it should be only exist on first one inside "moov" atom? or maybe im wrong? because in case this method could be changed and retrieve the duration value only one time by removing some uneeded code, maybe this value should be updated only when we change "track" atom, could be on period change? i have not many info on my head around this

For SmoothStreaming case:
 `InsertLiveSegment` is called automatically by using callbacks from samplereader
i think this is the best way since we need to add more subsequential values,
the previous implementation was also wrong on this, because the TFRF atom was wrongly implemented with lack of versioning support and lack of subsequent fragments parsing.
About this there is a problem unsolved:
Despite i have implemented the subsequent fragments parsing, currently still not working,
because the code that i have rollback on commit 6871e8664d829fdcbacd3b442bedc62481a74c2a prevent to add subsequent segments done with more callbacks, change this code cause weird buffering with mixed frames, where i havent found a good solution

Future todo:
while reworking this i found that missing also a proper implementation of `DVRWindowLength` in the SmoothStreaming (i add a todo in the code) use this value technically should allow us to get rid of `expired_segments_` variable and add a proper implementation to delete "expired" segments from the timeline itself, thus avoiding adding objects to the timeline vectors indefinitely without but deleting them

## Motivation and context

fix #1324 
## How has this been tested?

Only unencrypted DRM live:
`https://gstv-tnz-gsmediastreaming.preview-jpea.channel.media.azure.net/dfd06b62-e9d1-4a7f-bcbb-89d2ecbc82ee/preview.ism/manifest`
DRM protected live streams has been tested by the user of the issue https://github.com/xbmc/inputstream.adaptive/issues/1324#issuecomment-1636305191

## Types of change

## Checklist:


glennguy (approved):
Amazing, thank you for this huge effort!

glennguy:
Sample table boxes I believe are only within MOOV boxes (in the init data)... but in the case of 'prefixed initialization' then maybe there's a MOOV box on every segment? In that case I would assume that the box is the same every time. If there was a change in the init data then surely there would be a new period to signal the discontinuity and need for player/decoder reset.

PR #1342, opened 2023-07-27
[HLSTree] Reworked EXT-X-STREAM-INF parsing

### LINKED ISSUE #1259: Inputstream adaptive can not play hls without video
### Describe the bug
Inpustream Adaptive close when trying to play hls file that ocntain audio only (no video)

## Expected Behavior
Inputstream adaptive should play the audio even there is no video

## Actual Behavior

Inputstream adaptive closes without any error message

### To Reproduce
Try to
 play this free link: https://stream.rtl.lu/live/hls/radio/rtllx

### Debuglog
[kodi.txt](https://github.com/xbmc/inputstream.adaptive/files/11469136/kodi.txt)

### MPD/M3U8s/ISM

Available at  [hls stream](https://stream.rtl.lu/live/hls/radio/rtllx)

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [x] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Kodi version: 20.1 Nexus
 - inputstream.adaptive version: 20.3.6-1~kinetic

CastagnaIT:
## Description

While testing i found over time old HLS parsing problems not solved yet:
- On manifest with multi-codec video streams the representations mixes all codecs together, this was causing problems with the "Ask quality" stream selector (show mixed codecs) and adaptive streaming, because each adaptation set must have a single codec type as dash specs
- With manifests that have EXT-X-STREAM-INF variants with video streams but also variants with only audio (without video), these variants was wrongly added as a video stream causing broken playback on all sense, this can be confirmed by using "Ask quality" stream selector where wrong video streams are shown with lack of metadata

Related to this will be fixed #1259 since now EXT-X-STREAM-INF variants with only audio are now correctly handled

Has been full removed the `ExtGroup` struct because media groups are now parsed when required by variants itself

~Since merge adaptation sets with a separate method should not needed anymore, has been removed in full the merging code on `AdaptiveTree::SortTree`
This also revert/remove the code to `urn:mpeg:dash:adaptation-set-switching:2016` introduced by PR #694~
EDIT: After some more investigations i choosed to reintroduce "adaptation sets merging" method but for dash only, the reason behind is due to amazon manifests that appears to have uncommon adaptation sets with same data, for this reason i have add a new test case `AdaptionSetMerge` so to not forgot in future this peculiarity.
`AdaptiveTree::SortTree` method now does exactly what it describes and does not merge adpsets,
the adaptation set merging code that include also `urn:mpeg:dash:adaptation-set-switching:2016` introduced by PR #694 has been moved to a separate method in dash parser, but i am somewhat puzzled as to how it was implemented because:
- wipe out adaptation sets data
- for example on this manifest [SwitchingManifestExample.txt](https://github.com/xbmc/inputstream.adaptive/files/12194881/SwitchingManifestExample.txt) there are different protection levels i may be wrong but it would be good to keep this and other possible different data

Moving this code i have also limited the merging code to adaptation sets with same codec, currently mixing more codecs to the same adaptation set can cause broken playback, because we have no way to determine supported codecs in advance and also currently there is no way to fallback if playback dont works with a specific codec

## How has this been tested?

https://demo.unified-streaming.com/k8s/features/stable/video/tears-of-steel/tears-of-steel-hoh-subs.ism/.m3u8

and HLS audio stream from
https://stream.rtl.lu/live/hls/radio/rtllx

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptationSet.cpp
+++ b/src/common/AdaptationSet.cpp
@@ -10,6 +10,7 @@
 
 #include "Representation.h"
 #include "../utils/StringUtils.h"
+#include "../utils/Utils.h"
 
 #include <algorithm> // any_of
 
@@ -155,4 +156,28 @@ bool PLAYLIST::CAdaptationSet::Compare(const std::unique_ptr<CAdaptationSet>& le
   }
 
   return false;
+}
+
+PLAYLIST::CAdaptationSet* PLAYLIST::CAdaptationSet::FindByCodec(
+    std::vector<std::unique_ptr<CAdaptationSet>>& adpSets, std::string codec)
+{
+  auto itAdpSet = std::find_if(adpSets.cbegin(), adpSets.cend(),
+                               [&codec](const std::unique_ptr<CAdaptationSet>& item)
+                               { return CODEC::Contains(item->GetCodecs(), codec); });
+  if (itAdpSet != adpSets.cend())
+    return (*itAdpSet).get();
+
+  return nullptr;
+}
+
+CAdaptationSet* PLAYLIST::CAdaptationSet::FindMergeable(
+    std::vector<std::unique_ptr<CAdaptationSet>>& adpSets, CAdaptationSet* adpSet)
+{
+  auto itAdpSet = std::find_if(adpSets.cbegin(), adpSets.cend(),
+                               [&adpSet](const std::unique_ptr<CAdaptationSet>& item)
+                               { return item->IsMergeable(adpSet); });
+  if (itAdpSet != adpSets.cend())
+    return (*itAdpSet).get();
+
+  return nullptr;
 }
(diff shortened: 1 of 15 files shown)


CastagnaIT:
i found a problem with manifest like this
https://storage.googleapis.com/shaka-demo-assets/apple-advanced-stream-ts/master.m3u8
here EXT-X-MEDIA are declared after and not before EXT-X-STREAM-INF this now cause a wrong parsing

this because we have a linear text parsing by line,
instead we should have an additional middle step where we store data grouped by tags or a middle struct, after that parse the data, in this way doesnt matter if some tag is declared before or after another one

this require a full rework of `CHLSTree::ParseManifest`

PR #1344, opened 2023-07-29
Decrypter refactor

### LINKED ISSUE #736: [To do] wvdecrypter Refactor

glennguy:
## Description

Refactor of the single sample decrypter project
wvdecrypter remains a separate cmake project however:
* Separate projects for Android/other platforms
* Built as a static library and not a dll/shared library - part of inputstream.adaptive.so/dll
* Separate header/source files and separate files for each class (was previously one 'super file')

Factory pattern has been implemented for loading the decrypter, keeping a loose coupling inside main IA code.

Also begin work on tidying up - renaming classes and variables, small improvements here and there. I've drawn a line where I got to though as this will take forever if I fix everything.

## Motivation and context

This code has largely been untouched for some time, and the current way of loading a separate completely uncoupled dll does not make it easy to introduce new decrypter types. I can only imagine that the current setup was planned so that others can develop their own decrypter libraries separately however this concept never took off.

fix #736

## How has this been tested?

Tested on runtime in Windows and Android.

## Types of change

## Checklist:


glennguy:
@CastagnaIT  I think this is somewhat ready for a review. As mentioned I don't want to get too bogged down in the tidy up part of things. I've done some renaming and formatting but there's definitely room for improvement - in a separate PR though I think.

glennguy:
Still todo though - arm64 Linux.

From the suggestions that tmm1 made, I think the easiest will be to make a small dll file that does the reload business before opening the cdm. Haven't looked into it much though.

CastagnaIT:
yeah now that i finished my PR's i can start take a look here

CastagnaIT:
i have taken a fast look at PR code change,
in general its a copy/paste with some variations so imo looks ok for this big refactor
i will do the review after you add the changes i sent to you on slack ofc if you agree
(i have some little changes to request not big problems)
to not become crazyyou can make the PR with a single commit

i want only remember you that for next time when you add some comments in code regards changes to do or to check in future,
to always add `//! @todo: ...` so that for the review and for the future code cleanups its more easy do a "todo" search

> Still todo though - arm64 Linux.

for arm64 yes we need add a separate project dll to load with dlopen
that contains only the two arm methods
better a separate commit just for this

src/main.cpp
```diff
@@ -637,6 +637,10 @@ bool CVideoCodecAdaptive::Open(const kodi::addon::VideoCodecInitdata& initData)
 {
   if (!m_session || !m_session->GetDecrypter())
     return false;
+  
+  // need to set host instance in decrypter here
```
CastagnaIT: this is a todo?
if so please add `//! @todo:`

src/main.cpp
```diff
@@ -637,6 +637,10 @@ bool CVideoCodecAdaptive::Open(const kodi::addon::VideoCodecInitdata& initData)
...
+
+  //std::this_thread::sleep_for(std::chrono::seconds(600));
```
CastagnaIT: Can this commented code be removed?

TheDaChicken:
This is already such a big improvement:D

CastagnaIT (approved):
for me its ok as is,
widevine aarch64 fix can be reimplemented separately since require someone that can test it

feel free to merge it
just please rebase with master branch,
and i suggest to squash all commits in a single one

glennguy:
Thanks for all of your help @CastagnaIT !

PR #1346, opened 2023-07-31
[wvdecrypter] Fix memory heap corruption

CastagnaIT:
## Description

After PR #1310 when you play a Widevine encrypted content on a debug build kodi crash with a memory heap corruption
immagine

the reason seem to be on changed code that looks bad and unsafe,
the code take the mem pointer from data_out and store to packet_out, this to manually copy data by using `memcpy` seem to avoid use the bento4 object method `AppendData`

the size of buffer internally of the data_out object is assumed to be big enough to contains written data, but if not so?
weird thing can happens also because at the end is called `data_out.SetDataSize` to update the wrong object data size,
from what i understand SetDataSize not only update the container data size but can also reallocate the object buffer,
if this happens the reallocation copy the buffer data but based on initial data size and not on what we have written with `memcpy`, that could exceed the size...

I do not know if I have explained myself correctly anyway dont crash anymore

maybe in the past has been done so to try limit `SetDataSize` uses in the
`while (packet_in < packet_in_e)` loop, at least with h264 i get at max 4 loops only not big things

a test confermation would be appreciated

## How has this been tested?

played n€tflix on windows, debug build
played n€tflix on android, but release build

## Types of change

## Checklist:

### DIFF
--- a/wvdecrypter/wvdecrypter.cpp
+++ b/wvdecrypter/wvdecrypter.cpp
@@ -1162,11 +1162,8 @@ AP4_Result WV_CencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 pool_id,
 
     if (fragInfo.nal_length_size_ && (!iv || bytes_of_cleartext_data[0] > 0))
     {
-      //Note that we assume that there is enough data in data_out to hold everything without reallocating.
-
       //check NAL / subsample
       const AP4_Byte *packet_in(data_in.GetData()), *packet_in_e(data_in.GetData() + data_in.GetDataSize());
-      AP4_Byte *packet_out(data_out.UseData() + data_out.GetDataSize());
       AP4_UI16 *clrb_out(iv ? reinterpret_cast<AP4_UI16*>(data_out.UseData() + sizeof(subsample_count)):nullptr);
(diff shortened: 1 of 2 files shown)


glennguy:
Hi
I don't have an opportunity tonight to test it... but I think it needs to be tested on Windows with a non-secure path sample stream e.g the bitmoving parkour one - if you get a chance before I do:
```
#KODIPROP:inputstream.adaptive.license_key=https://cwip-shaka-proxy.appspot.com/no_auth|Content-Type=|R{SSM}|
https://bitmovin-a.akamaihd.net/content/art-of-motion_drm/mpds/11331.mpd
```

I was testing on this in the WIP PR and had crashing..,. maybe I messed up though and it's running secure path when it doesn't need to?

CastagnaIT:
thanks just tested, played without crashes
https://paste.kodi.tv/ofomicepan.kodi
but i would like a confermation on a system different of mine

glennguy:
My bad here - I have secure path currently hardcoded in my branch. Playback fine here with the changes, thanks heaps for finding this so quickly!

glennguy (approved):

PR #1351, opened 2023-08-05
Some cleanup to AdaptiveStream and Session

CastagnaIT:
## Description

Some cleanup to AdaptiveStream and Session should not functionality change some details on commits
the only exception is that the Choosers now can get callbacks for stream different than VIDEO (audio/subs) that however are already handled

## Motivation and context

i was investigating to improve HLS parser and im stuck on some of these to do future cleanups

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -52,8 +52,6 @@ AdaptiveStream::AdaptiveStream(AdaptiveTree& tree,
     m_fixateInitialization(false),
     m_segmentFileOffset(0),
     play_timeshift_buffer_(kodiProps.m_playTimeshiftBuffer),
-    rep_counter_(1),
-    prev_rep_(0),
     last_rep_(0)
 {
   current_rep_->current_segment_ = nullptr;
@@ -820,14 +818,6 @@ bool AdaptiveStream::ensureSegment()
     }
     else
       nextSegment = current_rep_->get_next_segment(current_rep_->current_segment_);
-
-    if(prev_rep_== current_rep_)
-      rep_counter_++;
-    else
-    {
-      rep_counter_=1;
-      prev_rep_=current_rep_;
-    }
 
     if (nextSegment)
     {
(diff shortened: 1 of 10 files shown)


CastagnaIT:
i dont know reason but dash dont works anymore, i will need to investigate

CastagnaIT:
problem fixed
was because i had removed the call to InitializeDRM from CSession::InitializePeriod for unencrypted streams case,
i thought it would be more intuitive to call InitializeDRM only when you have encrypted content, but do this requires more cleanup to the code than i thought so i revert the change

src/common/AdaptiveStream.cpp
```diff
@@ -840,37 +841,23 @@ bool AdaptiveStream::ensureSegment()
...
-              current_adp_, segment_buffers_[available_segment_buffers_ - 1]->rep);
-        }
+        // The representation from the last added segment buffer
+        CRepresentation* prevRep = segment_buffers_[available_segment_buffers_ - 1]->rep;
+
+        bool isLastSegment = nextsegmentPosold + available_segment_buffers_ ==
+                             current_rep_->SegmentTimeline().GetSize() - 1;
+        // Dont change representation if it is the last segment of a period
+        if (isLastSegment)
+          newRep = prevRep;
```
glennguy: I forgot about this.

I don't think my existing code was correct because I noticed a crash on this once. It was a long time ago so I can't remember the exact details - hoping that your change from using valid_segment_buffers_ to available_segment_buffers_ is where my error was.

FYI the reason behind adding this condition was that when it comes time to play the last segment in a period, AdaptiveStream wasn't able to insert the initialization segment (in the case of fMP4) and you would get corrupted or blank video for the last segment.

glennguy (approved):

src/common/AdaptiveStream.cpp
CastagnaIT: hmm better if i add this as comment in the code

PR #1358, opened 2023-08-19
[AdaptiveStream] Fix sidx/init parsing

CastagnaIT:
## Description

When i made the big parsers rework the `AdaptiveStream::ResolveSegmentBase` method confused me a lot
and also in how the `AdaptiveStream::parseIndexRange` was a bit mess implemented

Use case: on DASH parser when a representation has no SegmentTemplate, no SegmentBase, no Timeline and then only the BaseUrl
we was creating a fake SegmentBase just to have access to `AdaptiveStream::ResolveSegmentBase` method,
this (at least to me) was not clear from the old (<=Nexus) code flow and there should not be this odd workaround

so this PR rework `AdaptiveStream::parseIndexRange` in a more clear way, its funcitonality is unchanged, but just fixed
the creation of the initialization segment that i have broken with the parsers rework

`ResolveSegmentBase` method now has been now renamed with a more appropriate name

~**TODO:**
representation `HasInitPrefixed` code condition is causing a mess due to its weird unclear conditions
and wipe out the original MP4 stream MOOV atom because create a fake MOOV
https://github.com/xbmc/inputstream.adaptive/blob/21.1.3-Omega/src/Session.cpp#L1033-L1036~

UPDATE: "init prefixed" variable has been in the past implemented for SmoothStreaming manifest case, because with that protocol the MP4 provided dont contains the initialization segment (Movie (MOOV) atom), that is needed for bento4 to demux MP4 fragments, so look like that variable was used as workaround to avoid create the Movie atom with other manifest cases.
Now that condition has been full removed, reworked code to works for SmoothStreaming case only, so that not interfere with regular MP4 cases.

## Motivation and context

i found that n€tflix has broken audio on windows without passthru i finally understand what happens

## How has this been tested?

Dash: N€tflix trailers, audio working tested on Windows without passthru and Android
Dash: https://dash.akamaized.net/dash264/TestCases/4b/qualcomm/1/ED_OnDemand_5SecSeg_Subtitles.mpd
Smoothstreaming to test "init prefixed" and Movie MOOV atom creation

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1013,7 +1013,7 @@ void CSession::UpdateStream(CStream& stream)
   stream.m_info.SetCodecInternalName(codecStr);
 }
 
-AP4_Movie* CSession::PrepareStream(CStream* stream, bool& needRefetch)
+void CSession::PrepareStream(CStream* stream, bool& needRefetch)
 {
   needRefetch = false;
   CRepresentation* repr = stream->m_adStream.getRepresentation();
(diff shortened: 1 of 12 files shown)


CastagnaIT:
if you want to review okay, but let me possibility to merge by myself
i would like to have some time to check one thing

CastagnaIT:
I have done, i fixed also another bug related to wrong calculated segments ranges
i add the link to the DASH manifest use case in the description
where on Kodi 19/20 (and 21) cannot be currently played

glennguy (approved):
Looks good!

Issue #268, opened 2019-05-23
PSSH not found


matthuisman:
Hi @peak3d,

Working on my next add-on and struck this error when trying to playback content.

Could not extract license from video stream (PSSH not found)

I think in this case, the PSSH is returned in the license request json response.
But, this fails before every trying to get the license.

Attached is the debug log, mpd file, init.mp4
 
[master_h264-8adc3a0a.mpd.txt](https://github.com/peak3d/inputstream.adaptive/files/3210235/master_h264-8adc3a0a.mpd.txt)
[kodi.log](https://github.com/peak3d/inputstream.adaptive/files/3210237/kodi.log)
[init.mp4.txt](https://github.com/peak3d/inputstream.adaptive/files/3210239/init.mp4.txt)

Here is a screenshot of the android app raw payload to the license server
**https://lic.drmtoday.com/license-proxy-widevine/cenc/**
![image](https://user-images.githubusercontent.com/6225961/58225815-04b48e00-7d78-11e9-81f7-67fd97ac617a.png)

And here is the json response:
![image](https://user-images.githubusercontent.com/6225961/58225952-7a205e80-7d78-11e9-9d51-5edade423160.png)

matthuisman:
Here are the exact files the above Android app requests are for
[master_h264-5327795.mpd.txt](https://github.com/peak3d/inputstream.adaptive/files/3210264/master_h264-5327795.mpd.txt)
[init.mp4.txt](https://github.com/peak3d/inputstream.adaptive/files/3210265/init.mp4.txt)
[seg-1.m4f.txt](https://github.com/peak3d/inputstream.adaptive/files/3210266/seg-1.m4f.txt)

peak3d:
@matthuisman 
1.) the manifest does not support widevine DRM, I wonder that inputstream.adaptive doesn't stop earlier.
2.) pssh is necessary to get the license request, install the eme logger (chrome plugin) and look what pssh data is sent. I assume its a combination from key_id and content_id. You can pass this with inputstream.adaptive.license_data property.

matthuisman:
Is KID that same as PSSH?
https://bitmovin.com/docs/encoding/tutorials/how-to-create-widevine-drm-protected-content

I found the json response that returns the license server has a keyId: ad8debca0f7b4d57aa520ce6007d39ec as well as the url and custom-data.
The same key is also called 'protectionKey' elsewhere.

Would that maybe be the KID?

[...]

PR #1373, opened 2023-09-04
[DASHTree] Fixed license_data property use case

### LINKED ISSUE #1372: [Dash - Omega - Regression] - SESSION::CSession::InitializePeriod: Unhandled encrypted stream.
Kodi 21 fails very quickly
LOG:
[kodi.log](https://github.com/xbmc/inputstream.adaptive/files/12511649/kodi.log)

MANIFEST:
[androidtv.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/12511653/androidtv.mpd.txt)

I suspect its erroring as cant find Widevine.
This addon passes the license data using: inputstream.adaptive.license_data

Works OK Kodi 20
Kodi 20 working log:
[working.log](https://github.com/xbmc/inputstream.adaptive/files/12511752/working.log)

CastagnaIT:
## Description

after the parsers refactor when `inputstream.adaptive.license_data` is set is not more considered for two problems
- On CSession::InitializeDRM the `else if (!sessionPsshset.defaultKID_.empty())` condition was never true because the KID was not parsed anymore, due to no match with ContentProtection's tags and supportedKeySystem
- After rework the dash parser dont take in account of this use case anymore, because since dont find a match for supportedKeySystem, set period as EncryptionState::ENCRYPTED, and not as EncryptionState::ENCRYPTED_SUPPORTED. Then fails to initialize the DRM

With this PR has been also removed the support to inject the kid by using `{KID}` placeholder in the custom PSSH data provided with the license_data property. The main reasons are, its a very specific use case that sniff the kid from PlayReady ContentProtection tag only (if found), its use its so difficult and prone to not working, his implementation its hacky and will probably cause difficulties in the future to introduce support for new ContentProtection cases currently not implemented.

Additional thing, while investiganting i also found that ContentProtection for PlayReady with following format is not working
because we taking in account of `mspr:pro` tag but not `pro` child tag, so it has been added
```
<ContentProtection schemeIdUri="urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95">
	<cenc:pssh>...</cenc:pssh>
	<pro>...</pro>
</ContentProtection>
```

## Motivation and context

~try~ fix #1372
## How has this been tested?

tested by users in issue
## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -480,10 +480,10 @@ bool CSession::InitializeDRM(bool addDefaultKID /* = false */)
 
       if (sessionPsshset.pssh_ == PSSH_FROM_FILE)
       {
-        LOG::Log(LOGDEBUG, "Searching PSSH data in FILE");
-
         if (m_kodiProps.m_licenseData.empty())
         {
+          LOG::Log(LOGDEBUG, "Searching for PSSH data in the stream file");
+
           auto initialRepr{m_reprChooser->GetRepresentation(sessionPsshset.adaptation_set_)};
 
           CStream stream{*m_adaptiveTree, sessionPsshset.adaptation_set_, initialRepr, m_kodiProps};
@@ -554,17 +554,16 @@ bool CSession::InitializeDRM(bool addDefaultKID /* = false */)
           }
           stream.Disable();
         }
-        else if (!sessionPsshset.defaultKID_.empty())
+        else
         {
-          std::string licenseData = BASE64::Decode(m_kodiProps.m_licenseData);
-          // Replace KID placeholder, if any
-          STRING::ReplaceFirst(licenseData, "{KID}", sessionPsshset.defaultKID_);
+          // This can allow to initialize a DRM that could be also not specified
+          // as supported in the manifest (e.g. on DASH ContentProtection tags)
+          LOG::Log(LOGDEBUG, "Set init PSSH data provided by the license data property");
 
+          std::string licenseData = BASE64::Decode(m_kodiProps.m_licenseData);
           init_data.SetData(reinterpret_cast<const AP4_Byte*>(licenseData.c_str()),
                             static_cast<AP4_Size>(licenseData.size()));
         }
-        else
-          return false;
       }
       else
       {
--- a/src/common/AdaptiveUtils.h
+++ b/src/common/AdaptiveUtils.h
@@ -27,7 +27,7 @@ namespace PLAYLIST
 constexpr uint16_t PSSHSET_POS_DEFAULT = 0;
 // Marker for not valid psshset position
 constexpr uint16_t PSSHSET_POS_INVALID = std::numeric_limits<uint16_t>::max();
-// Marker to extract the PSSH from the file
+// Marker to try extract the PSSH from the file, or to try use a custom PSSH license data provided
 constexpr std::string_view PSSH_FROM_FILE = "FILE";
 // Marker for not set/not found segment position
 constexpr size_t SEGMENT_NO_POS = std::numeric_limits<size_t>::max();
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -88,6 +88,16 @@ std::string DetectCodecFromMimeType(std::string_view mimeType)
 
 adaptive::CDashTree::CDashTree(const CDashTree& left) : AdaptiveTree(left)
 {
+  m_isCustomInitPssh = left.m_isCustomInitPssh;
+}
+
+void adaptive::CDashTree::Configure(const UTILS::PROPERTIES::KodiProperties& kodiProps,
+                                    CHOOSER::IRepresentationChooser* reprChooser,
+                                    std::string_view supportedKeySystem,
+                                    std::string_view manifestUpdParams)
+{
+  AdaptiveTree::Configure(kodiProps, reprChooser, supportedKeySystem, manifestUpdParams);
+  m_isCustomInitPssh = !kodiProps.m_licenseData.empty();
 }
 
 bool adaptive::CDashTree::Open(std::string_view url,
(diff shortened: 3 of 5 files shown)


matthuisman:
im using windows-x86_64 and looks like thats the only build that failed :(
Are you able to try kick it off again?

matthuisman:
jenkins build again

CastagnaIT:
@glennguy i know atm you dont have time, but i leave here a comment for next days/week,
so you can give me your opinion on these changes

by inspecting again the old Nexus dash parser, seem that it when dont find a match for ContentProtection vs keySystem
always set current pssh to `"FILE"` (that is PSSH_FROM_FILE on Omega) so that in any case seem to allow extract pssh from stream file, but the old code flwo confuse me a bit

now i have modified `ParseTagContentProtection` method, so that its return value is ignored,
then will always execute `InsertPsshSet` (also when dont find a keySystem match, in this case set PSSH_FROM_FILE)
can have sense?

glennguy:
The old flow only worked when the tags appeared in a certain order (maybe it didn't work also because of some other missing attribute in one of the tags... I'm sure this is buried in an old issue somewhere). There was an issue with the VTM GO addon I think where they had to use a proxy to force the FILE way to happen because of this. But this was an edge case IMO.

Not sure if setting pssh from file makes sense if we don't have a keysystem match though, unless we expect the pssh box to reveal a keysystem supported that's not in the manifest (unlikely right?) e.g try to play DASH file that only supports fairplay drm. Is this change fixing the problem that Matt has though?

CastagnaIT:
yes the change fix the matt addon problem, that however is weird use
his manifest provide ContentProtection tag, but for playready only

```
    <AdaptationSet mimeType="video/mp4" scanType="progressive" segmentAlignment="true" startWithSAP="1" group="0">
      <ContentProtection schemeIdUri="urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95" value="PlayReady">
        <mspr:pro>xgIAAAEAAQC8AjwAVwBSAE0ASABFAEEARABFAFIAIAB4AG0AbABuAHMAPQAiAGgAdAB0AHAAOgAvAC8AcwBjAGgAZQBtAGEAcwAuAG0AaQBjAHIAbwBzAG8AZgB0AC4AYwBvAG0ALwBEAFIATQAvADIAMAAwADcALwAwADMALwBQAGwAYQB5AFIAZQBhAGQAeQBIAGUAYQBkAGUAcgAiACAAdgBlAHIAcwBpAG8AbgA9ACIANAAuADAALgAwAC4AMAAiAD4APABEAEEAVABBAD4APABQAFIATwBUAEUAQwBUAEkATgBGAE8APgA8AEsARQBZAEwARQBOAD4AMQA2ADwALwBLAEUAWQBMAEUATgA+ADwAQQBMAEcASQBEAD4AQQBFAFMAQwBUAFIAPAAvAEEATABHAEkARAA+ADwALwBQAFIATwBUAEUAQwBUAEkATgBGAE8APgA8AEwAQQBfAFUAUgBMAD4AaAB0AHQAcABzADoALwAvAHAAbABhAHkAcgBlAGEAZAB5AC4AZQBuAHQAaQB0AGwAZQBtAGUAbgB0AC4AbgBlAHgAdAAuAHQAaABlAHAAbABhAHQAZgBvAHIAbQAuAGMAbwBtAC8AcABsAGEAeQByAGUAYQBkAHkALwByAGkAZwBoAHQAcwBtAGEAbgBhAGcAZQByAC4AYQBzAG0AeAA8AC8ATABBAF8AVQBSAEwAPgA8AEsASQBEAD4AYgB0ADMAeQB6AFEAMQAxAFEAawBPADYAbABBAHEAUQBHAHAAUQA2ADAAUQA9AD0APAAvAEsASQBEAD4APABDAEgARQBDAEsAUwBVAE0APgBTAHEAVABRAGMAUQA2AGcAOABEAGMAPQA8AC8AQwBIAEUAQwBLAFMAVQBNAD4APAAvAEQAQQBUAEEAPgA8AC8AVwBSAE0ASABFAEEARABFAFIAPgA=</mspr:pro>
      </ContentProtection>
```

looks like that addon use `license_data` property https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptivelicense_data to play it with widevine
and ok seem to works, in short we avoid parse pssh and we use one signed with PSSH_FROM_FILE to allow Session::InitializeDRM to set the custom license data.

but current `ParseTagContentProtection` method is a bit hacked because `license_data` property allow also to inject the KID to custom license data, and we have to read the kid somewhere in some way
https://github.com/xbmc/inputstream.adaptive/blob/Omega/src/Session.cpp#L559-L561
by looking at nexus sourcecode the kid is sniffed from a ContentProtection with playready header, if found.
this is a very specific feature that come from initial ISA commits and there are no info about it,
moreover i have broken it with the rework and then not works anymore, so its possible that there are no addons that use it...
i would like to remove this particular Kid injection, because currently result more an hack, what do you think?
PS. in the current PR commit i add possibility in some way to parse kid to inject, but however i dont like too much this thing

@matthuisman do you have any memory of addons that use `license_data` property with `{kid}` placeholder injection?

CastagnaIT:
i found the discussion to https://github.com/xbmc/inputstream.adaptive/issues/268#issuecomment-495148309
seem peak3d has added this for a @matthuisman case, @matthuisman do you still need this "kid" placeholder injection?

matthuisman:
i didnt request that, think peak just added thinking I may need it.

I've searched my addons and I dont use {KID} placeholder.
In fact, only this single STAN addon uses license_data (no data in mpd) and it does not use {KID}

CastagnaIT:
fine then i will remove this placeholder and simplify code

matthuisman:
great. let me know when you want me to test :)

CastagnaIT:
i think i have done a mistake somewhere, i need to check

CastagnaIT:
done now its ready to be tested

@glennguy i have restored the previous behaviour, so the return value of ParseTagContentProtection is taken in account again,
just cleanup a bit more code, PR ready for review

matthuisman:
tested https://jenkins.kodi.tv/job/xbmc/job/inputstream.adaptive/job/PR-1373/6/artifact/cmake/addons/build/zips/inputstream.adaptive+windows-x86_64/inputstream.adaptive-21.2.0.zip

TESTED OK
Thank you

tekker:
Tested PR-1373-6 Armv7-Android works fine :)

OSX-Arm64 works fine :)

CastagnaIT:
@glennguy i merge PR now, i need to make a new ISA release for the new Kodi release that happens in next few hours
in case you want propose some code changes leave comments and i will do at a later time

PR #1387, opened 2023-09-26
Improvements/fixes to code misuses

### LINKED ISSUE #660: base64 encoded key in m3u8 does not work
I have encountered a HLS stream that uses base64 to encode the key directly into a m3u8 playlist

`#EXT-X-KEY:METHOD=AES-128,URI="data:;base64,LF6VBuLEMHdzuFSX3dHuIg==",IV=0x00000000000000000000000000001156`

This works in browsers but fails to play back in nearly every player I've tried including inputstream.adaptive which tries to fetch it from the server

CastagnaIT:
## Description

Changes:
 **`CPeriod::PSSHSet` `pssh_` changed from std::string to vector of uint8_t**
For better data integrity pssh data type has been changed to vector of uint8_t, but this highlighted some variable hacks, some i already knew others discovered now. So `pssh_` variable content type was inconsistent, because can have a mix of types, like string of bytes or base64 string or a placeholder (to extract protection data from stream) this will drive one crazy to understand how and why a value is different and in what cases, because may lead to hidden bad behaviours.
To recap this, ISM set pssh as string of bytes, DASH/HLS as base64 string, and DASH can also set it by using the placeholder `PSSH_FROM_FILE`, and HLS also have a bad hack that set to pssh variable the KID url (see CHLSTree::ProcessEncryption for AES-128 case).

To solve this situation, `pssh_` variable now contains only bytes as uint8_t, and no base64 string allowed, to do so:
Pssh data are decoded while parsing manifest, then the data will be always bytes.
The placeholder `PSSH_FROM_FILE` has been removed, and replaced by add a `HasNoData` method in the `CPeriod::PSSHSet`.
For the HLS case with KID to PSSH hack, i have decoupled the KID url from pssh.

**HLS: implemented appropriate URI parsing**
URI attribute parsing of EXT-X-KEY tag lead to wrong data (as #660 issue), because we are assuming that URI always start with `URI="data:text/plain;base64,`, now i add a new `CHLSTree::ConvertDataUriToBytes` method that allow a better safe data extracting, this way has be converted from https://github.com/video-dev/hls.js/blob/v1.4.12/src/loader/level-key.ts#L120,
and could be useful in future to implement playready support

**HLS: implemented appropriate PSSH parsing**
i avoided use Bento4 because too much bento4 data types related
the old code: https://github.com/xbmc/inputstream.adaptive/blob/21.3.0-Omega/src/parser/HLSTree.cpp#L858-L864
its very limited and ignore versioning, ignore pssh that can containing keys+data
on git history there is nothing explained about this, if someone has some info please leave a comment

**Add CHLSTree todo for `RenewLicense` method call**
this not wiki documented, i dont know how test it atm, there are some implementation questions that i havent found answers, if someone has some info please leave a comment

`CCharArrayParser` has been changed to `uint8_t`, since all uses are of this type

## Motivation and context

initially i was investigating for #660 and i thought of handling pssh data as uint8_t "bytes" instead of std::string its more appropriate, but while doing this data type conversion i got stuck on some code misuses to be fixed

fix #660

## How has this been tested?

HLS/DASH widevine
HLS AES128

## Types of change

## Checklist:

### DIFF
--- a/src/Iaes_decrypter.h
+++ b/src/Iaes_decrypter.h
@@ -19,7 +19,7 @@ class IAESDecrypter
 public:
   virtual ~IAESDecrypter() {};
 
-  virtual void decrypt(const AP4_UI08* aes_key,
+  virtual void decrypt(const std::vector<uint8_t>& aes_key,
                        const AP4_UI08* aes_iv,
                        const AP4_UI08* src,
                        std::vector<uint8_t>& dst,
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -322,16 +322,16 @@ bool CSession::PreInitializeDRM(std::string& challengeB64,
   const char* optionalKeyParameter{nullptr};
 
   // Set the provided PSSH
-  std::string decPssh{BASE64::Decode(psshData)};
+  const std::vector<uint8_t> decPssh = BASE64::DecodeStrToUint8(psshData);
   init_data.SetData(reinterpret_cast<const AP4_Byte*>(decPssh.data()),
                     static_cast<AP4_Size>(decPssh.size()));
 
   // Decode the provided KID
-  std::string decKid{BASE64::Decode(kidData)};
+  const std::vector<uint8_t> decKid = BASE64::DecodeStrToUint8(kidData);
 
   CCdmSession& session(m_cdmSessions[1]);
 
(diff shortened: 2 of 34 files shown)


CastagnaIT:
for now i close it since i need to think more about uint8_t changes

PR #1388, opened 2023-09-27
Improvements/fixes to code misuses

### LINKED ISSUE #660: base64 encoded key in m3u8 does not work

CastagnaIT:
## Description

Changes:
 **`CPeriod::PSSHSet` `pssh_` changed from std::string to vector of uint8_t**
For better data integrity pssh data type has been changed to vector of uint8_t, but this highlighted some variable hacks, some i already knew others discovered now. So `pssh_` variable content type was inconsistent, because can have a mix of types, like string of bytes or base64 string or a placeholder (to extract protection data from stream) this will drive one crazy to understand how and why a value is different and in what cases, because may lead to hidden bad behaviours.
To recap this, ISM set pssh as string of bytes, DASH/HLS as base64 string, and DASH can also set it by using the placeholder `PSSH_FROM_FILE`, and HLS also have a bad hack that set to pssh variable the KID url (see CHLSTree::ProcessEncryption for AES-128 case).

To solve this situation, `pssh_` variable now contains only bytes as uint8_t, and no base64 string allowed, to do so:
Pssh data are decoded while parsing manifest, then the data will be always bytes.
The placeholder `PSSH_FROM_FILE` has been removed and reordered conditions that set the license init PSSH in `CSession::InitializeDRM` so to fallback to extract the pssh as last resort.
For the HLS case with KID url to PSSH hack, i have decoupled the KID url from pssh.

**HLS: implemented appropriate URI parsing for data format**
URI attribute parsing of EXT-X-KEY tag lead to wrong data (as #660 issue), because we are assuming that URI always start with `URI="data:text/plain;base64,`, now i add a new `CHLSTree::GetUriByteData` method that allow a better safe data extracting, this way has be converted from https://github.com/video-dev/hls.js/blob/v1.4.12/src/loader/level-key.ts#L120,
and could be useful in future to implement playready support

**HLS: for AES-128, add support to read KID from EXT-X-KEY URI with base64 data format**
for example
```
#EXT-X-KEY:METHOD=AES-128,URI="data:;base64,dKbarMNMRSmPPEXQ3dDCNA=="
#EXT-X-KEY:METHOD=AES-128,URI="data:application/octet-stream;base64,xtfhsRCnE3+t2bVKn+gLEQ==",IV=0x50f881b1f9598945878e60ccb139bb9c
```
but i havent found a workable sample stream to test this use case

**HLS: implemented appropriate PSSH parsing**
i avoided use Bento4 because too much bento4 data types related
the old code: https://github.com/xbmc/inputstream.adaptive/blob/21.3.0-Omega/src/parser/HLSTree.cpp#L858-L864
its very limited and ignore versioning, ignore pssh that can containing keys+data
on git history there is nothing explained about this, if someone has some info please leave a comment

**Add CHLSTree todo for `RenewLicense` method call**
this not wiki documented, i dont know how test it atm, there are some implementation questions that i havent found answers, if someone has some info please leave a comment

`CCharArrayParser` has been changed to `uint8_t`, since all uses are of this type

**Code changes to `CSession::InitializeDRM` method to set license init PSSH data**
To remove the placeholder `PSSH_FROM_FILE` and keep a similar behaviour to set the initial pssh, i have reordered all these conditions
https://github.com/xbmc/inputstream.adaptive/blob/21.3.0-Omega/src/Session.cpp#L418-L533
hoping there is no oversights the behaviour is similar, with these differences:
- `license_data` ISA prop if set, now can always override an existing pssh
- The `drmOptionalKeyParam` (previously named `optionalKeyParameter`) has been limited to "com.microsoft.playready" license type, this because its PlayReady specific DRM parameter and should be not set to others DRM's
- Since `PSSH_FROM_FILE` dont exist anymore, we try to extract the protection data from the stream as last resort
- I have add some _todo_'s comments, around `license_data` behaviour because that ISA property have multiple mixed purposes that i find messy and should be reworked/changed

## Motivation and context

initially i was investigating for #660 and i thought of handling pssh data as uint8_t "bytes" instead of std::string its more appropriate, but while doing this data type conversion i got stuck on some code misuses to be fixed

fix #660

## How has this been tested?

HLS/DASH widevine
HLS AES128

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -66,7 +66,7 @@ CSession::CSession(const PROPERTIES::KodiProperties& kodiProps,
 
   if (!kodiProps.m_serverCertificate.empty())
   {
-    std::string decCert{BASE64::Decode(kodiProps.m_serverCertificate)};
+    std::string decCert{BASE64::DecodeToStr(kodiProps.m_serverCertificate)};
     m_serverCertificate.SetData(reinterpret_cast<const AP4_Byte*>(decCert.data()),
                                 static_cast<AP4_Size>(decCert.size()));
   }
@@ -320,15 +320,13 @@ bool CSession::PreInitializeDRM(std::string& challengeB64,
 
   AP4_DataBuffer init_data;
   init_data.SetBufferSize(1024);
-  const char* optionalKeyParameter{nullptr};
 
   // Set the provided PSSH
-  std::string decPssh{BASE64::Decode(psshData)};
-  init_data.SetData(reinterpret_cast<const AP4_Byte*>(decPssh.data()),
-                    static_cast<AP4_Size>(decPssh.size()));
+  std::vector<uint8_t> decPsshData = BASE64::Decode(psshData);
+  init_data.SetData(decPsshData.data(), static_cast<AP4_Size>(decPsshData.size()));
 
   // Decode the provided KID
-  std::string decKid{BASE64::Decode(kidData)};
+  std::string decKid{BASE64::DecodeToStr(kidData)};
 
   CCdmSession& session(m_cdmSessions[1]);
 
@@ -337,7 +335,7 @@ bool CSession::PreInitializeDRM(std::string& challengeB64,
 
   if (m_decrypter && init_data.GetDataSize() >= 4 &&
       (session.m_cencSingleSampleDecrypter = m_decrypter->CreateSingleSampleDecrypter(
-           init_data, optionalKeyParameter, decKid, true, CryptoMode::AES_CTR)) != 0)
+           init_data, "", decKid, true, CryptoMode::AES_CTR)) != 0)
   {
     session.m_cdmSessionStr = session.m_cencSingleSampleDecrypter->GetSessionId();
     sessionId = session.m_cdmSessionStr;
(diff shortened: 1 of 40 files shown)


CastagnaIT:
@glennguy i think is ok now
yeah there are a lot of sparse changes, i could not make separate commits because changing pssh type from string to uint8_t
required many adaptations to the code and fix the code misuses found around it

however what matters most to me is that you do a good check of changes done in to `CSession::InitializeDRM` method,
theoretically there should be no regressions

~ps. i havent tested on android yet, only windows~
android tested

src/Session.cpp
```diff
@@ -1584,3 +1523,74 @@ AP4_Movie* CSession::CreateMovieAtom(CStream* stream)
   }
   return nullptr;
 }
+
+bool CSession::ExtractStreamProtectionData(PLAYLIST::CPeriod::PSSHSet& sessionPsshset,
```
CastagnaIT, line 1527: @glennguy for protection data extraction, do you think worth adding a check to allow extraction on representation with MP4 ContainerType only?

CastagnaIT:
~i havent time to check yet, but android build once installed on kodi dont load the ISA binary
so weird i suspect could be the new cast for jhbyteArray but idk~

update: fixed

glennguy:
> HLS: implemented appropriate PSSH parsing
i avoided use Bento4 because too much bento4 data types related
the old code: https://github.com/xbmc/inputstream.adaptive/blob/21.3.0-Omega/src/parser/HLSTree.cpp#L858-L864
its very limited and ignore versioning, ignore pssh that can containing keys+data
on git history there is nothing explained about this, if someone has some info please leave a comment

This would go back to the original implementation of HLS/Widevine. Nothing fancy here, just using magic numbers to pull the PSSH out of the text.

src/samplereader/FragmentedSampleReader.h
```diff
@@ -81,7 +81,7 @@ class ATTR_DLL_LOCAL CFragmentedSampleReader : public ISampleReader, public AP4_
   AP4_DataBuffer m_encrypted;
   AP4_DataBuffer m_sampleData;
   CodecHandler* m_codecHandler{nullptr};
-  const AP4_UI08* m_defaultKey{nullptr};
+  std::vector<uint8_t> m_defaultKey;
```
glennguy, line 84: Just checking - is this ok to now be uninitialised?

src/Session.cpp
glennguy: I'm not sure this even applies for other containers... not enough knowledge on that sorry!

glennguy (approved):
Great work, just minor comments

src/samplereader/FragmentedSampleReader.h
CastagnaIT: before was nullptr so we have to do nothing more for vector
seem that we check for empty key here
https://github.com/xbmc/inputstream.adaptive/pull/1388/files#diff-c4b54d32a3ee2ca0e116448b56e9276218c27b93c81fce3abf402083263572abR894
at least for non android case

src/Session.cpp
CastagnaIT: i don't either, bento4 should be sufficient to stop parsing non-conforming data, shouldn't be a big problem
i was just thinking to try to limit the execution of this method when not needed
it can be considered later

mediaminister:
I get a green screen with this PR:

STRM-file or use the VRT MAX Kodi add-on
```
#EXTM3U

#EXTINF:-1,VRT 1
#KODIPROP:inputstream.adaptive.license_key=https://widevine-license.vudrm.tech/proxy|x-vudrm-token=vrt%7C2023-10-20T09%3A26%3A40Z%7Cv2%7CE0LPBIWNzqhemk9kk7hyra2r3YJnxcIbOBBqI5IuhXN%2Bvhrtqo%2B%2FgEXSLPJLX3mgehjW0%2FSk9V4h%2BoAsfA4dsxz6wDlUWIPTsEeF2O4HIuw%3D%7C7635173e09c693193d90b34d831d08d1b109c73471943613696cc0ff42831f29|R{SSM}|
https://live.vrtcdn.be/groupc/live/8edf3bdf-7db3-41c3-a318-72cb7f82de66/live.isml/.mpd?filter=%28%21%28type%3D%3D%22audio%22%26%26FourCC%21%3D%22AACL%22%29%29

```

CastagnaIT:
thanks for the sample i missed this use case
i will investigate soon

Issue #1195, opened 2023-03-18
Headers set by Server aren't forwarded


completafurdc:
### Describe the bug
Here is a clear and concise description of what the problem is:
When a channel is played, It redirects the URL to another location. The new URL is response sets some `cookie` header which are needed to load the children. As I can see **403 Error by debugging**, It looks like this is probably the cause as the server sent headers aren't set.

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

The headers set by server should have also been set and sent to children. Like `Cookie`

## Actual Behavior

The headers set by server aren't sent. Like `Cookie`

### To Reproduce

1. Played a channel where server is setting some sort of Header, `Cookie` in my case

### Debuglog

The debuglog can be found here:
NA

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
NA

### Screenshots 
Here are some links or screenshots to help explain the problem:

NA

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

NA

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Windows 11
 - Kodi version: 20.1
 - inputstream.adaptive version: 20.3.5

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

CastagnaIT:
i assume that you are talking of HLS manifest (?) and that the cookie is provided by the master playlist for the children?
there is already opened issue for this #184

this is a custom requirement for some servers, its possible that we will implement this support in future with a property to enable it

as workaround, some users use a custom addon service that make requestwith right cookies by using the addon service and provide the manifests to ISA
https://github.com/xbmc/inputstream.adaptive/wiki/How-to-provide-custom-manifest-and-license
but i dont know if it is easily possible with HLS

completafurdc:
Hi! Thanks for your response. I am taking about both HLS and DASH Manifest. Sorry didn't looked to so old issues.

Yes, actually a lot more providers these days require this as a part of basic authentication. Custom solution to implement a server won't be possible on scale and can bring in several unforseen edge cases. 

[...]

PR #1390, opened 2023-10-02
Implement service broker singleton

CastagnaIT:
## Description

The main idea is to provide a way to have shared resources without being forced to make copies / transport them through the methods of classes

I thought of implementing this while i was trying to find a solution to #184, #1195 where is needed to share cookies from manifest response to all subsequent http requests, such as child manifests (hls) or manifest update, segments, kid url (hls), etc...
being that we have no way to share resources, imo this could be a solution that could be expanded also for other uses in future

The current idea for this implementation is:
- [x] Add to service broker the `KodiProperties` where can be easily read, so removing all `KodiProperties` uses from methods/classes
- [x] Add to service broker a way to read/store add-on settings as common way, this avoids using the kodi addon interface e.g. `kodi::addon::GetSettingBoolean` directly in the code by dispersing it, thus having a common place to easily maintain it
- [ ] Allow service broker to store cookies to be shared (as fix for #184, #1195)

~WIP - I opened PR in advance for possible discussions~

This is intended as code refactor without change existing code behaviours,
~with the exception for cookies sharing that however should not result in a big change, i will do in a separate commit~

UPDATE: The cookie implementation change i will do in a separate PR

## How has this been tested?

play random stream
## Types of change

## Checklist:


glennguy (approved):
Love it, well done!

src/SrvBroker.h
```diff
@@ -0,0 +1,73 @@
...
+  static CSrvBroker* GetInstance()
+  {
+    static CSrvBroker instance;
+    return &instance;
+  }
+
+  /*
+   * \brief Get Kodi properties component, to read Kodi (Listitem / playlist files "STRM") properties.
+   */
+  static std::shared_ptr<ADP::KODI_PROPS::CCompKodiProps> GetKodiProps()
```
lrusak, line 54: Why not a unique_ptr and return a reference? Do you actually need a shared_ptr?

CastagnaIT: yeah i think we can also do in that way

CastagnaIT: i tried but looks like its not possible to do
seem that unique_ptr require class to be declared and not foward declared as done here

lrusak: > i tried but looks like its not possible to do
> 
> seem that unique_ptr require class to be declared and not foward declared as done here

You can forward declare a class used in a unique_ptr. If you have compilation issues it may be because you are missing a constructor/destructor which may need to be added to the .cpp file (even if they are just declared default).

CastagnaIT: i have tried as you said by creating constructor/destructor in to cpp for each class but compiler produce same errors
`error C2338: static_assert failed: 'can't delete an incomplete type'`
at this point better use raw pointers instead of smartpointer and create and delete them manually

PR #1403, opened 2023-10-20
[Session] Fix regression to extract PSSH/KID

CastagnaIT:
## Description

There were two bugs
1) On Session i used STRING::ToHexadecimal to convert systemId, where instead was to be used the hex bytes conversion
2) On Dash parser when ContentProtection tag dont provide a pssh, before the PR #1388 instead of an empty pssh was set the placeholder PSSH_FROM_FILE value, after that PR, since there is an empty value cause AdaptiveTree::InsertPsshSet to fails. Now i have changed to allow always insert an empty pssh, because Session is able to use `ExtractStreamProtectionData` method to extract it when missing.

i have cleanup a bit also related code and i noticed that PSSHSet::m_usageCount usage is exclusive of a single HLS parser line for widevine case only:
https://github.com/xbmc/inputstream.adaptive/blob/51b565977c9a6921144bc44235c1bb25a15b4ae2/src/parser/HLSTree.cpp#L266C1-L266C1
that ask to myself if this could be maybe improved in future to remove in some way that count state

## Motivation and context

Bug feedback comment https://github.com/xbmc/inputstream.adaptive/pull/1388#issuecomment-1772223023

## How has this been tested?

Unfurnately the provided sample by user seem to be expired after short time
so i need a feedback by user itself

## Types of change

## Checklist:

### DIFF
--- a/src/common/Period.h
+++ b/src/common/Period.h
@@ -112,7 +112,7 @@ public:
 
   uint16_t InsertPSSHSet(const PSSHSet& pssh);
   void RemovePSSHSet(uint16_t pssh_set);
-  void DecrasePSSHSetUsageCount(uint16_t pssh_set);
+  void DecreasePSSHSetUsageCount(uint16_t pssh_set);
   std::vector<PSSHSet>& GetPSSHSets() { return m_psshSets; }
 
   // Make use of PLAYLIST::StreamType flags
--- a/src/common/Period.h
+++ b/src/common/Period.h
@@ -94,8 +94,8 @@ public:
     // Custom comparator for std::find
     bool operator==(const PSSHSet& other) const
     {
-      return m_usageCount == 0 || (media_ == other.media_ && pssh_ == other.pssh_ &&
-                                   defaultKID_ == other.defaultKID_ && iv == other.iv);
+      return media_ == other.media_ && pssh_ == other.pssh_ && defaultKID_ == other.defaultKID_ &&
+             iv == other.iv;
     }
 
     //! @todo: create getter/setters
@@ -110,10 +110,9 @@ public:
     CAdaptationSet* adaptation_set_{nullptr};
   };
 
-  uint16_t InsertPSSHSet(PSSHSet* pssh);
-  void InsertPSSHSet(uint16_t pssh_set) { m_psshSets[pssh_set].m_usageCount++; }
+  uint16_t InsertPSSHSet(const PSSHSet& pssh);
   void RemovePSSHSet(uint16_t pssh_set);
-  void DecrasePSSHSetUsageCount(uint16_t pssh_set) { m_psshSets[pssh_set].m_usageCount--; }
+  void DecrasePSSHSetUsageCount(uint16_t pssh_set);
   std::vector<PSSHSet>& GetPSSHSets() { return m_psshSets; }
 
   // Make use of PLAYLIST::StreamType flags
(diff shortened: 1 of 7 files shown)


CastagnaIT:
@mediaminister please give a feedback by using testbuilds of this PR
your provided sample stream i think its expired because dont work anymore

mediaminister:
Yes, samples expire after a few hours.
Here is a new one:
```
#EXTM3U

#EXTINF:-1,VRT 1
#KODIPROP:inputstream.adaptive.license_key=https://widevine-license.vudrm.tech/proxy|x-vudrm-token=vrt%7C2023-10-20T14%3A17%3A53Z%7Cv2%7C4zNQGpNteOCHiNYO5dnOcQXKUtPlwFU5xD%2BgyoGrM%2FJ6XdI1H0YIQEZKE9fvBCxQtc7ozsMOiPEQiWuVNgFvH2BtszYCGuD80KUmu44Ng5E%3D%7C8457ae4b96699d183eaa5285792ba9cb591240b30b8694b06d3bc3c59e882bb6|R{SSM}|
https://live.vrtcdn.be/groupc/live/8edf3bdf-7db3-41c3-a318-72cb7f82de66/live.isml/.mpd?filter=%28%21%28type%3D%3D%22audio%22%26%26FourCC%21%3D%22AACL%22%29%29
```
I'll test this at a later moment.

CastagnaIT:
thanks new sample work good

glennguy:
> i noticed that PSSHSet::m_usageCount usage is exclusive of a single HLS parser line for widevine case only

This goes back a couple of years so I'm having a hard time remembering. I think the idea might be to avoid requesting a new DRM license when changing streams if it's using the same key.
This value was also used in the HLS AES-128 case when the same key is used multiple times. If the key rotates and then drops off the playlist (live) then the use count drops to zero and that PSSH can be removed. Not sure if it's still like this.

src/common/Period.h
```diff
@@ -110,10 +110,9 @@ class ATTR_DLL_LOCAL CPeriod : public CCommonSegAttribs
     CAdaptationSet* adaptation_set_{nullptr};
   };
 
-  uint16_t InsertPSSHSet(PSSHSet* pssh);
-  void InsertPSSHSet(uint16_t pssh_set) { m_psshSets[pssh_set].m_usageCount++; }
+  uint16_t InsertPSSHSet(const PSSHSet& pssh);
   void RemovePSSHSet(uint16_t pssh_set);
-  void DecrasePSSHSetUsageCount(uint16_t pssh_set) { m_psshSets[pssh_set].m_usageCount--; }
+  void DecrasePSSHSetUsageCount(uint16_t pssh_set);
```
glennguy: Just a typo - missing an 'e' in Decrease

glennguy (approved):

CastagnaIT:
> This goes back a couple of years so I'm having a hard time remembering. I think the idea might be to avoid requesting a new DRM license when changing streams if it's using the same key.

yeah, i think that could be more appropriate move this out from the parser,
would be to investigate whether it is possible to know (from the decrypter?) whether the license for the key used has already been requested, so that this count can be removed

> This value was also used in the HLS AES-128 case when the same key is used multiple times. If the key rotates and then drops off the playlist (live) then the use count drops to zero and that PSSH can be removed. Not sure if it's still like this.

i suspect is changed

PR #1411, opened 2023-10-31
[cleanups] Cleanups to includes and minor things

CastagnaIT:
## Description

removed some unused includes, foward some other to try to lighten compiler to save some building secs

## Types of change

## Checklist:

### DIFF
--- a/lib/webm_parser/src/master_parser.h
+++ b/lib/webm_parser/src/master_parser.h
@@ -151,13 +151,10 @@ class MasterParser : public ElementParser {
 
   // Hash functor for hashing Id enums for storage in std::unordered_map.
   struct IdHash : StdHashId {
-    // Type aliases for conforming to the std::hash interface.
-    using argument_type = Id;
-    using result_type = StdHashId::result_type;
-
     // Returns the hash of the given id.
-    result_type operator()(argument_type id) const {
-      return StdHashId::operator()(static_cast<StdHashId::argument_type>(id));
+    size_t operator()(Id id) const
+    {
+      return StdHashId::operator()(static_cast<std::underlying_type<Id>::type>(id));
     }
   };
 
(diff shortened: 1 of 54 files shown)


CastagnaIT:
@lrusak i have removed shared_ptr on last commit of this pr

lrusak:
> @lrusak i have removed shared_ptr on last commit of this pr

@CastagnaIT see this commit -> https://github.com/lrusak/inputstream.adaptive/commit/ad3b05e8882e503926aa4db0cdf03a2c8ef4292a

I only build tested and briefly runtime tested it.

There seems to be no nullptr checks so I assume that Init is called before anything else is available.

CastagnaIT:
thanks for the example, i had done similar but only for the deconstructor and was not working
i will copy this way but i prefer have pointers than references, yeah its not intended to be allowed have nullptr's
is builded before doing any other addon operation so on "Open"

glennguy (approved):

lrusak:
> thanks for the example, i had done similar but only for the deconstructor and was not working
> 
> i will copy this way but i prefer have pointers than references, yeah its not intended to be allowed have nullptr's

You should likely be using references here. It doesn't make sense to use pointers if they can never be null anyways. We have quite a few examples of this in core Kodi now. You'll see our CServiceBroker returns references. The places that it returns pointers are likely because the pointer can be null (like windowing) and should always be checked before use.

> is builded before doing any other addon operation so on "Open"

Perfect 👍

CastagnaIT:
> You should likely be using references here. It doesn't make sense to use pointers if they can never be null anyways. We have quite a few examples of this in core Kodi now. You'll see our CServiceBroker returns references. The places that it returns pointers are likely because the pointer can be null (like windowing) and should always be checked before use.

oh i understood ok i will change it

PR #1412, opened 2023-10-31
Workaround to keep in memory cookies between http sessions

### LINKED ISSUE #184: Missing Cookie On Some Requests
inputstream.adaptive is not sending cookies for some requests resulting in error response.

See attached log the request to https://live01.video.globo.com/d/s/dash-univ24ha/dash-univ24ha_audio-377579.mp4 is not sending the required cookies resulting in 400 response.
[kodi.log](https://github.com/peak3d/inputstream.adaptive/files/2346577/kodi.log)

Here is the mpd manifest:
[manifest.txt](https://github.com/peak3d/inputstream.adaptive/files/2346579/manifest.txt)

### LINKED ISSUE #1195: Headers set by Server aren't forwarded

CastagnaIT:
## Description

Cookies management here its an hack,
currently there are no way to have a persistent HTTP session with Kodi binary interface, at least not in the usual way.

Curl library implementation in kodi manage a pool of connections (called sessions) by domain, and this actually keeps the cookies in memory but not in a persistent way

when a connection remain unused for some seconds will be deleted and so delete cookies,
or when there is a multithreading access could pick a session (easy_handle) currently busy,
therefore it create a new empty session (easy_handle) ofc with no cookies,
because they are stored on another curl session that could be meantime deleted, see:
https://github.com/xbmc/xbmc/blob/21.0a3-Omega/xbmc/filesystem/DllLibCurl.cpp#L143-L146
https://github.com/xbmc/xbmc/blob/21.0a3-Omega/xbmc/filesystem/DllLibCurl.cpp#L173

A solution is the use of curl shared data CURLOPT_SHARE to share data with multiple handles,
but there is still a need to review how to implement this in the binary addons interface as well,
this seem however not easy fix to do in kodi core and could be done in future kodi versions only

~currently WIP until someone can give me a feedback~
**UPDATE:**
looks like atm no one can provide a feedback, i still finished the PR in principle it should work at least now i can see persistent cookies from curl log.

Since cookies are a special use case of some HLS/DASH services so not commonly used,
i add a property to enable the internal cookies management "workaround", so by default its _disabled_.
I choosen to default disable it, because from a web search seem that other players not support cookies, or support them but require to be enabled by player properties etc... and also because its not full tested yet (no users feedbacks)
maybe in future this could be changed as default enabled, depends on whether the problem will be fixed/implemented in kodi core itself or what will be done

**New property:** `inputstream.adaptive.internal_cookies`
**Accepted values:** `true` or `false`

## Motivation and context

followup of PR #1390

fix #184
fix #1195

## How has this been tested?

I have "tested" with a RAI live stream that share a weird cookie that its not really needed for the playback

**NOT tested** with a service that use cookies as authentication to allow download streams

## Types of change

## Checklist:

### DIFF
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -10,7 +10,7 @@
     name="adaptive"
     extension=""
     tags="true"
-    listitemprops="license_type|license_key|license_url|license_url_append|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_upd_params|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max|live_delay"
+    listitemprops="license_type|license_key|license_url|license_url_append|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_upd_params|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max|live_delay|internal_cookies"
     library_@PLATFORM@="@LIBRARY_FILENAME@"/>
   <extension point="xbmc.addon.metadata">
     <platform>@PLATFORM@</platform>
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -17,6 +17,7 @@ set(HEADERS
   Session.h
   SrvBroker.h
   CompKodiProps.h
+  CompResources.h
   CompSettings.h
   Stream.h
   aes_decrypter.h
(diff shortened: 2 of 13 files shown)


glennguy (approved):
Awesome, thanks @CastagnaIT !

PR #1449, opened 2024-01-09
Some tree / session cleanups

CastagnaIT:
## Description

some cleanups while investigating on current issues

commits:
1) We want avoid print a log error when an HLS child manifest has not been prepared yet
and ContainerType result as NOTYPE
2) cleanups
3) cleanups
4) Prevent to show empty resolution to GUI, can happens when there is no HLS RESOLUTION info, like on some live streams
5) Removed not needed method argument stream type
6) Removed needRefetch, the reasons behind this of this forced behaviour are unclear to me, `OpenStream` require return value to be true only when there is need to update stream info, but this is already done by the reader with `GetInformation` and extradata if needed is already taken in account. Tested with and without DRM HLS stream not found weirdness. @glennguy if you remember something let me know, also too much old GH history there are no explanations
7) small fix, if there is no CHANNELS info on rendition, fallback to 2 channels
8) as title, reproducible with royal family stream https://cdn-ue1-prod.tsv2.amagi.tv/avod/simplestream-lds-ldtimln/ODYPPCS001EP001/ODYPPCS001EP001.m3u8

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -902,9 +902,8 @@ void CSession::UpdateStream(CStream& stream)
   stream.m_info.SetCodecInternalName(codecStr);
 }
 
-void CSession::PrepareStream(CStream* stream, bool& needRefetch)
+void CSession::PrepareStream(CStream* stream)
 {
-  needRefetch = false;
   CRepresentation* repr = stream->m_adStream.getRepresentation();
 
   switch (m_adaptiveTree->prepareRepresentation(stream->m_adStream.getPeriod(),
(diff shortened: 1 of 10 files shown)


glennguy:
Peak3d implemented the needRefetch iirc when he did some refactoring after the initial HLS/WV implementation. I think the real test for this is when periods or reps change but PSSH remains the same (common key for all streams). Disney+ is probably in this case if you still have an account to test with?

glennguy (approved):

matthuisman:
i emailed @CastagnaIT a disney+ login 10 days ago. can re-send if needed :)

CastagnaIT:
this week im short in time to give to developing
hoping tonight i find some time to make a test with D+ (if login works)

CastagnaIT:
i have tested D+ (thanks @matthuisman works!) with a couple of tvshows having more periods at the near the end of video
and i havent found playback problems, tried also seek backward / forward in the middle of chapters and seem ok

Issue #1132, opened 2023-02-10
Broken HLS AES-128 decryption and live audio buzzy


CrAcK75:
### Describe the bug

The latest versione of Inputstream 20.3.3 has broken something on Kodi 20 with this stream, now there is no audio:
Rai1 => https://mediapolis.rai.it/relinker/relinkerServlet.htm?cont=2606803&output=7
If I downgrade to the previous 20.3.2 it works again

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

## Actual Behavior

Audio not present

## Possible Fix

No possible fix found, only if downgrade to the previous 20.3.2 it works again

### To Reproduce

1.  start the stream
2.
3.

### Debuglog

The debuglog can be found here:
this a portion of log when the channel starts:
[kodi.log](https://github.com/xbmc/inputstream.adaptive/files/10711704/kodi.log)

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
https://mediapolis.rai.it/relinker/relinkerServlet.htm?cont=2606803&output=7

this is the m3u8 that in this moment become from the url:
[playlist_ma.m3u8.txt](https://github.com/xbmc/inputstream.adaptive/files/10711651/playlist_ma.m3u8.txt)

### Screenshots 
Here are some links or screenshots to help explain the problem:

![image](https://user-images.githubusercontent.com/66442920/218209137-d5cc9960-0251-47ed-a39f-212759742c85.png)

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

### Your Environment
Used Operating system:

 - [X] Android
 - [ ] iOS
 - [ ] tvOS
 - [X] Linux
 - [ ] OSX
 - [X] Windows
 - [ ] Windows UWP

 - Operating system version/name: Windows (10, 11) or AndroidTV (Box, Firestick)
 - Kodi version: 20
 - inputstream.adaptive version: 20.3.3

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

glennguy:
Thanks for the report @CrAcK75 

I was able to have a brief look shortly after you posted this using the links in the log but now I get geoblocked.

@CastagnaIT since this stream is local to you are you able to investigate?

[...]

PR #1459, opened 2024-01-21
[AdaptiveStream] Rework to improve HLS manifest updates

### LINKED ISSUE #1109: Subtitles from German live broadcasters are not displayed in Kodi 20 (Nexus)
## Bug report
Subtitles from German live broadcasters (ARD, ZDF, etc.) are not displayed in Kodi 20 (Nexus), although selected in the OSD
### Describe the bug
It shows a single subtitle available: "German - Untertitel deutsch (Standard)" but nothing shows when it is turned on.

## Expected Behavior
Subtitles should be displayed on the screen

## Actual Behavior
Subtitles are not displayed in Kodi 20 (Nexus), although selected in the OSD

### To Reproduce

1. Call up the stream (using PVR IPTV Simple Client etc.) 
2. Enable subtitles via the OSD menu

### Debuglog

[kodi_debug_nexus.log](https://github.com/xbmc/xbmc/files/10491763/kodi_debug_nexus.log)

### Screenshots

Streaming Adresses of German live broadcasters (ZDF): 
https://zdf-hls-15.akamaized.net/hls/live/2016498/de/veryhigh/master.m3u8

### Your Environment
Used Operating system:

 - [x ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: LineageOS 19 based on Android for Raspberry Pi 400
 - Kodi version: 20 (Nexus)

*note: Once the issue is made we require you to update it with new information or Kodi versions should that be required.

### LINKED ISSUE #1438: HLS live stream stops after several seconds in version 21.4.1
## Bug report
There seems to be a bug introduced with version 21.4.1 concerning playback of certain HLS streams. The stream starts but terminates after 10-60 seconds. With version 21.4.0 and the same configuration the same streams play without problems.
### Describe the bug

The stream is terminated and the last kodi screen appears. Streams of the same provider stop at different times but all within 1 minute.
The streams originate from a service provider in SEE. The streams are geoblocked and require a valid subscription. The addon which uses inputstream is a PVR addon. 

## Expected Behavior

Expected behaviour is that the stream plays the same way as in previous version like 21.4.0

## Actual Behavior

Instead the stream terminates without user interaction

## Possible Fix

Maybe the changes between 21.4.0 and 21.4.1 concerning HLS (I can see two in the changelog) could be reverted.

### To Reproduce

1. Start stream from service provider
2. Wait for the crash

### Debuglog

A debug log of the issue is here:
https://paste.kodi.tv/ugacunumad

### MPD/M3U8s/ISM

Will be provided  in comments

### Screenshots 

### Your Environment
Used Operating system:

 - [ x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Kodi version: Omega Beta
 - inputstream.adaptive version: 21.4.1

CastagnaIT:
## Description

While was developing this i found many weirdness that i think are not new bugs,
most of these bugs can be reproduced by using "Stream selection" setting to "Test" by switching per 1 segment.
That could happens for example on live stream, that when switch quality kodi call OpenStream, but when AdaptiveStream::start_stream is called, dont find the next segment because not yet available and stop the playback wrongly....

One single time, i catched two kodi freeze/crash, when played "akmai" local stream:
1 case when doing video seek, two times, in fast way on the "same" point (with mouse)
1 case when stop playback
more likely i think are already existings bugs

PR Changes for live streams:
- HLS: Fix problem of playback stops after some time
- HLS: Fix playback stop with adaptive stream switching (not fully but much better than before)
- HLS: RAI live buzz sound mentioned in issue #1132 seem not happens anymore, and instead of buzz sound the stream pause for in right way buffering, but the workaround its still needed, to avoid better the buffer loading
- HLS: Fix child manifest subtitles still in download also when are disabled
- HLS: Fix subtitles segments not downloaded from new manifest updates (only show subtitles for first secs/segments) related to #1109

Unsolved bug was already existing:
if you play a HLS live stream with subtitles that are disabled,
when switch to a new period (DEMUX_SPECIALID_STREAMCHANGE) the subtitle stream will be re-enabled, this cause error
GenerateSidxSegments: [AS-20] Cannot generate segments
due to not downloaded child manifest

Unsolved bug i think was already existing:
Chapter change (DEMUX_SPECIALID_STREAMCHANGE) may be executed 2 times! by audio and video stream, i think due to slight differrent pts, could be?!, not easy to catch but its happened to me 1 or 2 times
idk if kodi expect one single chapter change?

~tomorrow i will comment some code parts... to give some details on code changes~ done

## Motivation and context

I have lost 2 weeks of nights on this also because its a big mess from old messy implementations and things that from my part i find difficult to learn how works

i think however that what i have done here should be better of what was there before and after estenuants tests seem almost stable, at least without hard use of adaptive streaming quality switching (on my tests always: "Test" mode per 1 segment)

fix #1438

this will fix also subtitles segment problem of #1109
where subtitles was displayed only for the first segments only and then stopped

## How has this been tested?

For multiperiod/chapters/ads:
Live Pluto TV i cant post the master manifest m3u8 link because expire after hours, but can be found on network flow of:
https://pluto.tv/en/live-tv/

http://mcdn.daserste.de/daserste/de/master.m3u8 (also for subs, not always there are news/tvshow)

"Akmai local stream"

Other live, no chapters:
https://cdndai.pl/tvp1hd/index.m3u8 (sometime the server may temporary ban your IP for ~30minuts)

RAI:
#KODIPROP:inputstream.adaptive.live_delay=30
https://mediapolis.rai.it/relinker/relinkerServlet.htm?cont=2606803&output=7

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -720,15 +720,34 @@ bool AdaptiveStream::start_stream(uint64_t startPts)
             current_rep_->get_segment(static_cast<size_t>(segmentId - current_rep_->GetStartNumber()));
       }
     }
+    else if (stream_changed_) // switching streams, align new stream segment no.
+    {
+      uint64_t segmentId = segment_buffers_[0]->segment_number;
+      if (segmentId >= current_rep_->GetStartNumber() + current_rep_->SegmentTimeline().GetSize())
+      {
+        segmentId = current_rep_->GetStartNumber() + current_rep_->SegmentTimeline().GetSize() - 1;
+      }
+      current_rep_->current_segment_ = current_rep_->get_segment(
+          static_cast<size_t>(segmentId - current_rep_->GetStartNumber()));
+    }
     else
+    {
       current_rep_->current_segment_ = nullptr; // start from beginning
+    }
   }
+
+  stream_changed_ = false;
 
   const CSegment* next_segment =
       current_rep_->get_next_segment(current_rep_->current_segment_);
 
   if (!next_segment)
   {
+    //! @todo: THIS MUST BE CHANGED - !! BUG !!
+    //! this will broken/stop playback on live streams when adaptive stream change stream quality (so representation)
+    //! and there are no new segments because not immediately available (despite child manifest updated),
+    //! because we cant be ensure next segment without wait the appropriate timing...imo its not the right place here.
+    //! Can be reproduced with HLS live and by using "Stream selection" setting to "Test" by switching per 1 segment
     absolute_position_ = ~0;
     state_ = STOPPED;
     return true;
@@ -772,11 +791,11 @@ bool AdaptiveStream::start_stream(uint64_t startPts)
     return false;
   }
 
-  currentPTSOffset_ = (next_segment->startPTS_ * current_rep_->timescale_ext_) /
-    current_rep_->timescale_int_;
+  currentPTSOffset_ =
+      (next_segment->startPTS_ * current_rep_->timescale_ext_) / current_rep_->timescale_int_;
   absolutePTSOffset_ =
       (current_rep_->SegmentTimeline().Get(0)->startPTS_ * current_rep_->timescale_ext_) /
-    current_rep_->timescale_int_;
+      current_rep_->timescale_int_;
 
   if (state_ == RUNNING)
   {
@@ -838,7 +857,6 @@ bool AdaptiveStream::ensureSegment()
     if (m_fixateInitialization)
       return false;
 
-    stream_changed_ = false;
     CSegment* nextSegment{nullptr};
     last_rep_ = current_rep_;
 
@@ -849,16 +867,18 @@ bool AdaptiveStream::ensureSegment()
                   segment_buffers_.begin() + available_segment_buffers_);
       --valid_segment_buffers_;
       --available_segment_buffers_;
-      // Check if next segment use same representation of previous one
-      // if not, update the current representation
+      // Adaptive stream has changed quality (and so changed representation)
       if (segment_buffers_[0]->rep != current_rep_)
       {
         current_rep_->SetIsEnabled(false);
         current_rep_ = segment_buffers_[0]->rep;
         current_rep_->SetIsEnabled(true);
+        // When stream changed flag is signalled, kodi reopen the stream and AdaptiveStream::start_stream method
+        // will be called also to align the "current segment" for the current representation
         stream_changed_ = true;
       }
     }
+
     if (valid_segment_buffers_ > 0)
     {
       if (!segment_buffers_[0]->segment.IsInitialization())
@@ -910,20 +930,25 @@ bool AdaptiveStream::ensureSegment()
           newRep = prevRep;
         else
           newRep = m_tree->GetRepChooser()->GetNextRepresentation(current_adp_, prevRep);
+
+        if (newRep != prevRep) // Stream quality changed
+        {
+          // On manifests type like HLS we need also to get update segments
+          // because need to be downloaded/parsed from different child manifest files
+          bool isDrmChanged;
+          m_tree->PrepareRepresentation(current_period_, current_adp_, newRep, isDrmChanged);
+
+          // If the representation has been changed, segments may have to be generated (DASH)
+          if (newRep->SegmentTimeline().IsEmpty())
+            GenerateSidxSegments(newRep);
+        }
       }
 
-      // If the representation has been changed, segments may have to be generated (DASH)
-      if (newRep->SegmentTimeline().IsEmpty())
-        GenerateSidxSegments(newRep);
+      // Add to the buffer next segment (and the next following segments, if available)
 
-      if (!newRep->IsPrepared() && m_tree->SecondsSinceRepUpdate(newRep) > 1)
-      {
-        m_tree->prepareRepresentation(current_period_, current_adp_, newRep, m_tree->IsLive());
-      }
-
-      // Add to the buffer next future segment
-
+      // Align to new segment numeration (that could be changed if manifest update occurred with quality switching)
       size_t nextsegmentPos = static_cast<size_t>(nextsegno - newRep->GetStartNumber());
+
       if (nextsegmentPos + available_segment_buffers_ >= newRep->SegmentTimeline().GetSize())
       {
         nextsegmentPos = newRep->SegmentTimeline().GetSize() - available_segment_buffers_;
@@ -963,7 +988,8 @@ bool AdaptiveStream::ensureSegment()
       if (!current_rep_->IsWaitForSegment())
       {
         current_rep_->SetIsWaitForSegment(true);
-        LOG::LogF(LOGDEBUG, "[AS-%u] Begin WaitForSegment stream %s", clsId, current_rep_->GetId().data());
+        LOG::LogF(LOGDEBUG, "[AS-%u] Begin WaitForSegment stream id \"%s\"", clsId,
+                  current_rep_->GetId().data());
       }
       std::this_thread::sleep_for(std::chrono::milliseconds(100));
       return false;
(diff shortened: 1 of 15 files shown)


src/common/AdaptiveStream.cpp
```diff
@@ -720,15 +722,32 @@ bool AdaptiveStream::start_stream(uint64_t startPts)
             current_rep_->get_segment(static_cast<size_t>(segmentId - current_rep_->GetStartNumber()));
       }
     }
+    else if (stream_changed_) // switching streams, align new stream segment no.
```
CastagnaIT, line 723: I have lost most of the time due to this method AdaptiveStream::start_stream
because i was not aware that the AdaptiveStream::start_stream method is used to adjust/set the "current segment" and so all my changes that i was doing on ensure_segment always was failing.
After discovered this, i realized that was missing set "current segment" for a stream quality switching case

src/common/AdaptiveStream.cpp
```diff
@@ -720,15 +722,32 @@ bool AdaptiveStream::start_stream(uint64_t startPts)
...
   const CSegment* next_segment =
       current_rep_->get_next_segment(current_rep_->current_segment_);
 
   if (!next_segment)
   {
+    //! @todo: THIS MUST BE CHANGED - !! BUG !!
+    //! this will broken/stop playback on live streams when adaptive stream change stream quality (so representation)
+    //! and there are no new segments because not immediately available (despite child manifest updated),
+    //! because we cant be ensure next segment without wait the appropriate timing...imo its not the right place here.
+    //! Can be reproduced with HLS live and by using "Stream selection" setting to "Test" by switching per 1 segment
```
CastagnaIT, line 750: I also found this problem (read todo) i tried force remove this block of code that stop playback, and a bit below avoid set currentPTSOffset_ when there is no next_segment, playback works but show weird pts on Kodi GUI the value of time becomes sometimes negative and it can lead to weirdness in the long run.
Need to be changed on other way but idk atm, can be done separately, maybe can you help here

src/common/AdaptiveStream.cpp
```diff
@@ -910,20 +930,25 @@ bool AdaptiveStream::ensureSegment()
           newRep = prevRep;
         else
           newRep = m_tree->GetRepChooser()->GetNextRepresentation(current_adp_, prevRep);
-      }
 
-      // If the representation has been changed, segments may have to be generated (DASH)
-      if (newRep->SegmentTimeline().IsEmpty())
-        GenerateSidxSegments(newRep);
-
-      if (!newRep->IsPrepared() && m_tree->SecondsSinceRepUpdate(newRep) > 1)
```
CastagnaIT, line 919: On representation class `IsPrepared` and `m_isDownloaded` plus the `SecondsSinceRepUpdate` imo was really a mess of behaviours, i have removed them all, and add a new `IsNeedsUpdates`, used only internally in the HLSParser to distinguish whether a representation needs an update of some thing (in this case for live updates)

src/common/AdaptiveStream.cpp
```diff
@@ -1212,17 +1238,22 @@ size_t adaptive::AdaptiveStream::getSegmentPos()
...
   if ((m_tree->HasManifestUpdates() || m_tree->HasManifestUpdatesSegs()) && state_ == RUNNING)
   {
     std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(m_tree->GetTreeUpdMutex());
 
     if (current_rep_ && current_rep_->IsWaitForSegment())
     {
-      return !checkTime ||
-             (current_adp_->GetStreamType() != StreamType::VIDEO &&
-              current_adp_->GetStreamType() != StreamType::AUDIO) ||
-             SecondsSinceUpdate() < 1;
```
CastagnaIT, line 1224: `SecondsSinceUpdate` uses (now removed) above and here should be the cause of the RAI buzz/frezze sound, by debugging this waitingForSegment method, was producing messed up return values

src/common/AdaptiveStream.cpp
```diff
@@ -1212,17 +1238,22 @@ size_t adaptive::AdaptiveStream::getSegmentPos()
...
+      return !checkTime || (current_adp_->GetStreamType() != StreamType::VIDEO &&
+                            current_adp_->GetStreamType() != StreamType::AUDIO);
     }
+
+    // We can fall here when we are waiting for new segments, but the download it not started yet
+    // so return true only when checkTime is false, to avoid trigger EOS on sample reader and stop playback
+    return !checkTime;
```
CastagnaIT, line 1256: This is the main cause because the Pluto TV stream was stopping playback (before was returning "false")

src/common/AdaptiveStream.cpp
```diff
@@ -1289,16 +1320,15 @@ bool AdaptiveStream::GenerateSidxSegments(PLAYLIST::CRepresentation* rep)
...
-
   if (thread_data_)
   {
     thread_data_->Stop();
     StopWorker(STOPPED);
   }
+  // Disable representation only after stopped the worker
+  // otherwise if read some segments may invalidate this change
+  if (current_rep_)
+    current_rep_->SetIsEnabled(false);
```
CastagnaIT, line 1331: Here the problem should be due to `AdaptiveStream::ensureSegment`
since force change `SetIsEnabled` on the representation and race condition may happens

src/common/Representation.h
```diff
@@ -259,8 +257,7 @@ class ATTR_DLL_LOCAL CRepresentation : public CCommonSegAttribs, public CCommonA
   uint32_t assured_buffer_duration_{0};
   uint32_t max_buffer_duration_{0};
 
-  //! @todo: used for HLS only, not reflect the right meaning
-  bool m_isDownloaded{false};
+  uint64_t m_segPosition{SEGMENT_NO_NUMBER};
```
CastagnaIT: todo for me: remove me!

src/parser/DASHTree.cpp
```diff
@@ -1552,7 +1552,6 @@ void adaptive::CDashTree::RefreshSegments(PLAYLIST::CPeriod* period,
 {
   if (adp->GetStreamType() == StreamType::VIDEO || adp->GetStreamType() == StreamType::AUDIO)
   {
-    m_updThread.ResetStartTime();
```
CastagnaIT, line 1555: this should be a remnant of the old manifest update code, has no sense reset the worker thread here, isnt used when we update manifest at each new segment request "manually"

src/parser/HLSTree.cpp
```diff
@@ -173,442 +173,469 @@ bool adaptive::CHLSTree::Open(std::string_view url,
   return true;
 }
 
-PLAYLIST::PrepareRepStatus adaptive::CHLSTree::prepareRepresentation(PLAYLIST::CPeriod* period,
-                                                                     PLAYLIST::CAdaptationSet* adp,
-                                                                     PLAYLIST::CRepresentation* rep,
-                                                                     bool update)
+bool adaptive::CHLSTree::PrepareRepresentation(PLAYLIST::CPeriod* period,
```
CastagnaIT, line 176: Here github show a big mess on code diff, there are no functional code changes on PrepareRepresentation, the only difference is that has been decoupled per functionality.
The old `prepareRepresentation` now has been splitted on more methods:
- PrepareRepresentation: Act more or less as before
- DownloadChildManifest: To download the child manifest only
- ParseChildManifest: Parse the child manifest data
- PrepareSegments: Update the segments (to adjust/set "current segment")

now ParseChildManifest could be reused in future to parse in right way the media playlist on CHLSTree::ParseManifest (see todo), not tested now, but a step forward to achieve it

CastagnaIT:
@glennguy when you will have time try test this PR and review
i add some comments in code and also on github to try explain the code changes

src/common/AdaptiveStream.cpp
glennguy: yes the scenario makes sense why it fails. We could try a hacky workaround until something better comes along - assume when the next segment comes it will be the same size (usually is) and add that PTS to current segment's start PTS. Sound good or not?

src/common/AdaptiveStream.cpp
glennguy: Good find!

glennguy (approved):
Thanks so much for this work and fixing some of the issues in adaptivestream.

src/common/AdaptiveStream.cpp
CastagnaIT: i understood but as you said its hacky im not 100% sure that could be so good

i found this problem even when trying to fix bugs to make this live sample playable

This live sample have periods but no new segments provided,
then its required the use of "InsertLiveSegment" in conjuction with the manifest update worker.
This sample somewhat works on kodi <=v20 because of old code hacks and periods seem "ignored",
on kodi 21 is ofc broken, from what i see seem there are multiple problems on  `CDashTree::RefreshLiveSegments` that seem dont take in account of "inserted" segments on timeline, delete them all and so seem to loose the last inserted segment with the updated pts (because you have to take in account that the downloaded updated manifest dont provide new segments), this will make problems with `AdaptiveStream::ensureSegment` that its not able to find next segment, `AdaptiveStream::start_stream` have problem when switch period, and also with `CSession::OnSegmentChangedRead` that its not able to insert new segments anymore
(these problems may be also similar with https://github.com/xbmc/inputstream.adaptive/issues/1461)

PR #1464, opened 2024-01-31
[widevine] Fix bad data pointer on clrb_out variable

### LINKED ISSUE #1394: Amazon VOD: Kodi crashes opening a video with inputstream.adaptive >21.1.2
Starting a video, kodi crashes in 5 of 10 attempts if using inputstream.adaptive with versions > 21.1.2 - mainly if resuming a video.
I compiled the version 21.1.2 for actual Kodi 21a3 and it works - changing to higher versions Kodi sometimes (very often) crashes - also with most actual "Omega"-version.
See also [this bug-report](https://github.com/Sandmann79/xbmc/issues/706) where i was advised go create an issue here.
Attached please find a crashlog using actual Omega-version of inputstream.adaptive.
[kodi_crashlog_20231007200445.log](https://github.com/xbmc/inputstream.adaptive/files/12840071/kodi_crashlog_20231007200445.log)

### LINKED ISSUE #1452: Version 20.3.16 won't play anything in my firestick
### Describe the bug

I was using inputstream.adaptive 20.3.11 in my firestick (old model) with kodi 20. It worked fine. But after updating to version 20.3.16 it seems playback doesn't work with any addon.

There's no error but playback never starts (circle spinning or maybe the controls appear but the movie never starts).

## Expected Behavior

Playback should work.

## Actual Behavior

Spinning circle. Playback never starts.

### To Reproduce

1. Try to play anything with an addon that uses inputstream.adaptive, for example with the prime video addon on a (old?) firestick.

### Debuglog

This is the log with 20.3.16 (not working):
https://paste.kodi.tv/dosemiyedi.kodi

This is the log with 20.3.11 (working):
https://paste.kodi.tv/evucowufow.kodi

### MPD/M3U8s/ISM

I think a manifest is not relevant since the problem seems to happen with all addons.

### Screenshots 

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Amazon AFTT with Android TV 5.1.1 
 - Kodi version: 20.2 
 - inputstream.adaptive version: 20.3.16 

CastagnaIT:
## Description

Premise, i havent understood the data target format of `dataOut`
this prevents me from understanding how the data should be composed/constructed in to dataOut

from my investigations i found that:
dataOut.UseData() was used to get and store current data pointer but after that, many times is used "AppendData",
this method before append data check if the internal buffer of AP4_DataBuffer is enough to store data if not re-create the internal buffer, and so the data pointer before stored (to `clrb_out`) will point to a bad memory address causing the segmentation fault crash

NOTE: the crash affect most of times aarch64/ARM64 cpu's, other systems looks like problem remains hidden maybe due to system optimization the "old" memory address still working

## Motivation and context

fix #1394
fix #1452

## How has this been tested?

Tested am@zon VOD videos episodes on:
CoreElec with aarch64 cpu
Android nvidia shield pro
Windows

## Types of change

## Checklist:

### DIFF
--- a/src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
+++ b/src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
@@ -823,12 +823,10 @@ AP4_Result CWVCencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 poolId,
       //check NAL / subsample
       const AP4_Byte *packetIn(dataIn.GetData()),
           *packetInEnd(dataIn.GetData() + dataIn.GetDataSize());
-      AP4_UI16* clrb_out(
-          iv ? reinterpret_cast<AP4_UI16*>(dataOut.UseData() + sizeof(subsampleCount))
-             : nullptr); // is there a use for this?
-      size_t nalUnitCount = 0; // is there a use for this?
+      unsigned int clrbPos = sizeof(subsampleCount);
+      // size_t nalUnitCount = 0; // is there a use for this?
       size_t nalUnitSum = 0;
-      size_t configSize = 0; // is there a use for this?
+      // size_t configSize = 0; // is there a use for this?
 
       while (packetIn < packetInEnd)
       {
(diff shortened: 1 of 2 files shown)


glennguy (approved):
Thanks! Looks like this would have been a PITA to chase down, great work

CastagnaIT:
quite... in the future more likely would better add code comments to explain data construction that at least for me is not so immediately clear

PR #1486, opened 2024-02-21
[workflows] Try fix Invalid format error

CastagnaIT:
## Description

Looks like the equal sign must not be used
and "EOF" delimiter name should not used within the "value" so changed to "EOFn"

## Types of change

## Checklist:

### DIFF
--- a/.github/workflows/changelog-and-release.yml
+++ b/.github/workflows/changelog-and-release.yml
@@ -93,9 +93,9 @@ jobs:
           then
             changes=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/extension/news)' | awk -v RS= 'NR==1')
           fi
-          echo "changes=<<EOF" >> $GITHUB_OUTPUT
+          echo "changes<<EOFn" >> $GITHUB_OUTPUT
           echo "$changes" >> $GITHUB_OUTPUT
-          echo "EOF" >> $GITHUB_OUTPUT
+          echo "EOFn" >> $GITHUB_OUTPUT
           version=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/@version)')
           echo "version=$version" >> $GITHUB_OUTPUT
           branch=$(echo ${GITHUB_REF#refs/heads/})


MoojMidge:
Struggled with this myself. As an FYI it recommended that the delimiter used is random (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings) because otherwise the delimiter could be inserted into the value being outputted, followed by arbitrary commands, that would then be run in the context of the action. This could be used to exfiltrate project tokens amongst other things. 

Examples of random delimiters:
https://stackoverflow.com/a/74232400
And outputting a changelog from a file, line by line:
https://github.com/anxdpanic/plugin.video.youtube/blob/ee7979eb0b9d433cf05776a5fa4e7bbf6b192157/.github/workflows/make-release.yml#L40

This is just an FYI, and probably won't be an issue for this particular workflow, because it is only manually run by dispatch, but is something to be aware of if the workflow is run automatically on a PR.

phunkyfish:
This will also need to be done in release.yml.

I can handle that one too when I make then change for all the binary addons

.github/workflows/changelog-and-release.yml
```diff
@@ -93,9 +93,9 @@ jobs:
           then
             changes=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/extension/news)' | awk -v RS= 'NR==1')
           fi
-          echo "changes=<<EOF" >> $GITHUB_OUTPUT
+          echo "changes<<EOFn" >> $GITHUB_OUTPUT
           echo "$changes" >> $GITHUB_OUTPUT
```
phunkyfish, line 97: Should this be `echo "changes=$changes" >> $GITHUB_OUTPUT`?

phunkyfish: Or does the first line essentially do that assignment?

phunkyfish: Ok, that is exactly the case. I sent an update PR that introduces best practice for EOF and makes the same changes for release.yml: https://github.com/xbmc/inputstream.adaptive/pull/1487

CastagnaIT: > Should this be `echo "changes=$changes" >> $GITHUB_OUTPUT`?

this is right but only when you have a string, without multilines

PR #1487, opened 2024-02-21
Use a random EOF (best practice) for multiline changes variable and update label to Omega

phunkyfish:
Ok, can you test this is ok?

It introduces best practice for a random EOF and makes the same changes to release.yml

Once we know this ok, I'll set up the same change for the Nexus/Omega branch for all the binary add-ons

### DIFF
--- a/.github/workflows/changelog-and-release.yml
+++ b/.github/workflows/changelog-and-release.yml
@@ -85,6 +85,7 @@ jobs:
       # - steps.required-variables.outputs.version: version element from addon.xml.in
       # - steps.required-variables.outputs.branch: branch of the triggering ref
       # - steps.required-variables.outputs.today: today's date in format '%Y-%m-%d'
+      # Note: we use a random EOF for 'changes' as is best practice for for multiline variables
       - name: Get required variables
         id: required-variables
         run: |
@@ -93,9 +94,10 @@ jobs:
           then
             changes=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/extension/news)' | awk -v RS= 'NR==1')
           fi
-          echo "changes<<EOFn" >> $GITHUB_OUTPUT
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "changes<<$EOF" >> $GITHUB_OUTPUT
           echo "$changes" >> $GITHUB_OUTPUT
-          echo "EOFn" >> $GITHUB_OUTPUT
+          echo "$EOF" >> $GITHUB_OUTPUT
           version=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/@version)')
           echo "version=$version" >> $GITHUB_OUTPUT
           branch=$(echo ${GITHUB_REF#refs/heads/})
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog-and-release.yml
```diff
@@ -93,6 +93,7 @@ jobs:
           then
             changes=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/extension/news)' | awk -v RS= 'NR==1')
           fi
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
           echo "changes<<EOFn" >> $GITHUB_OUTPUT
```
MoojMidge: To use the `EOF` variable this line should be
```
echo "changes<<${EOF}" >> $GITHUB_OUTPUT
```

.github/workflows/changelog-and-release.yml
```diff
@@ -93,6 +93,7 @@ jobs:
...
           echo "$changes" >> $GITHUB_OUTPUT
           echo "EOFn" >> $GITHUB_OUTPUT
```
MoojMidge: To use the `EOF` variable this line should be
```
echo "${EOF}" >> $GITHUB_OUTPUT
```

.github/workflows/release.yml
```diff
@@ -38,20 +38,18 @@ jobs:
...
             changes=$(xmlstarlet fo -R "$(find . -name addon.xml.in)" | xmlstarlet sel -t -v 'string(/addon/extension/news)' | awk -v RS= 'NR==1')
           fi
-          changes="${changes//'%'/'%25'}"
-          changes="${changes//$'\n'/'%0A'}"
-          changes="${changes//$'\r'/'%0D'}"
-          changes="${changes//$'\\n'/'%0A'}"
-          changes="${changes//$'\\r'/'%0D'}"
-          echo "changes=$changes" >> $GITHUB_OUTPUT
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "changes<<EOFn" >> $GITHUB_OUTPUT
```
MoojMidge: To use the `EOF` variable this line should be
```
echo "changes<<${EOF}" >> $GITHUB_OUTPUT
```

.github/workflows/release.yml
```diff
@@ -38,20 +38,18 @@ jobs:
...
+          echo "$changes" >> $GITHUB_OUTPUT
+          echo "EOFn" >> $GITHUB_OUTPUT
```
MoojMidge: To use the `EOF` variable this line should be
```
echo "${EOF}" >> $GITHUB_OUTPUT
```

MoojMidge (commented):
Just need to update the delimiter to use the `EOF` variable

phunkyfish:
@MoojMidge please check it's all in order now.

MoojMidge:
My bash-fu is not great but LGTM

.github/workflows/release.yml
```diff
@@ -38,20 +39,18 @@ jobs:
...
-          changes="${changes//'%'/'%25'}"
-          changes="${changes//$'\n'/'%0A'}"
-          changes="${changes//$'\r'/'%0D'}"
-          changes="${changes//$'\\n'/'%0A'}"
-          changes="${changes//$'\\r'/'%0D'}"
-          echo "changes=$changes" >> $GITHUB_OUTPUT
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "changes<<${EOF}" >> $GITHUB_OUTPUT
+          echo "$changes" >> $GITHUB_OUTPUT
+          echo "${EOF}" >> $GITHUB_OUTPUT
```
CastagnaIT: i have already seen your proposed changes on other repos
but the brackets added to EOF var `${EOF}`
seem to be not required, so should appears as `$EOF`

CastagnaIT:
i propose to remove brackets to EOF,
i will test when we will do a future new release,
but should work without problems

.github/workflows/release.yml
MoojMidge: Yeah there is no grouping or expansion, so `$EOF` will work fine

phunkyfish: Ok, I removed the brackets. So we should be good now.

CastagnaIT (approved):

PR #1489, opened 2024-02-21
[DashTree] Fix setting default AdaptationSet using ISA specific default attribute

MoojMidge:
## Description
Fix what is possibly a typo, to allow setting an AdaptationSet as default using the corresponding ISA specific `default` attribute

## Motivation and context
It is not otherwise possible to set a default AdaptationSet without changing its role to `main`, which is not appropriate for subtitles.

## How has this been tested?
Sorry don't have a build environment, so this hasn't been tested at all.

I have marked this as a breaking change, because I am not sure how this attribute is commonly used, but this should provide what I think is the expected functionality of this attribute.

## Types of change

## Checklist:


CastagnaIT (approved):
thanks, yes looks like a typo

PS. for next times, remember to prefix commits with class names or context like explained on wiki
https://github.com/xbmc/inputstream.adaptive/wiki/Contributing-and-code-guidelines
this help to keep a good git history

Issue #1341, opened 2023-07-26
Many subtitle fragments in ism manifests are skipped


Paco8:
### Describe the bug
Here is a clear and concise description of what the problem is:

The VOD videos from [plugin.video.orange.spain](https://github.com/Paco8/plugin.video.orange.spain) use ism manifests with fragmented ttml subtitles. There's one fragment for every 2 seconds.

From time to time but very often Kodi skips one of those fragments (it's downloaded but from some reason it's not displayed).
If I rewind some seconds just after this happened maybe then the fragment is displayed.

This happens in all versions of Kodi, from 18 to 21, and all operating systems.

I think it has to do with the timestamps. If (using a proxy) I modify the last digit of the timestamps with a random number then the fragments are not skipped.

There's a easy way to test the problem. If I replace the original texts in the fragments with this:
```
<p begin="00:00:00.000" end="00:00:02.000">Hello World</p>
```

it should display "Hello World" during all video, right? But it doesn't. It displays the text for 2 seconds, then it displays nothing for 2 seconds, the it displays the text for 2 seconds, then it displays nothing...

For some reason in this case it displays the even fragments but not the odd ones.

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

Some fragments are not displayed on screen.

## Actual Behavior

All fragments should be displayed.

### To Reproduce

I couldn't find a free stream to reproduce the issue, and I'm afraid I can't provide either a strm to test because the token expires after 2 minutes. However, if you need to play a video to see the problem, I think I can provide you with a cookie file for the addon. Write me to mitvbox.spain at gmail dot com

### Debuglog

The debuglog can be found here:
https://paste.kodi.tv/kifuretato.kodi

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:

[manifest.zip](https://github.com/xbmc/inputstream.adaptive/files/12173654/manifest.zip)

### Screenshots 
Here are some links or screenshots to help explain the problem:

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

I attach all subtitle fragments from an episode.

[fragments.zip](https://github.com/xbmc/inputstream.adaptive/files/12173660/fragments.zip)

During the first 11 minutes (I didn't test any further) the following fragments are not shown:
```
49:
    <div>
      <p begin="00:00:00.000" end="00:00:02.000" region="bottomCenter">Si me matas, cerrarás el portal.</p>
[...]

PR #1499, opened 2024-03-01
[Codechandler][TTML] Add workaround to overlapped subtitles

CastagnaIT:
## Description

This is a workaround for Kodi overlapped subtitles rendering,
when there are multiple lines of text provided by multiple "p" elements usually with same begin/end timings (overlapping)

the kodi subtitle renderer (libass) show on screen these lines of text in "reversed order"
(last added on top, the first on bottom) that is the SRT behaviour,
and causes the display of inverted sentences with TTML.

308976449-81fada44-069d-4e7e-a432-12786b129947

where should be displayed as:
```
Su tercera pareja
fue la periodista checa,
```

So TTML multilines should be shown "as is" as it appears in the xml file.
```xml
  <body>
    <div>
      <p begin="00:00:00.000" end="00:00:00.359" region="speaker">de GermÃ¡n Kafka, su padre. </p>
      <p begin="00:00:00.599" end="00:00:02.000" region="speaker">Su tercera pareja </p>
      <p begin="00:00:00.599" end="00:00:02.000" region="speaker">fue la periodista checa, </p>
    </div>
  </body>
```

The workaround consists in to merge these lines with same begin/end timing,
but can fail when the end time is different or if "p" elements come from different media segments.

not a perfect solution but for now nothing better can be done

DEPENDS FROM PR #1497 

## Motivation and context

overlapping problem #https://github.com/xbmc/inputstream.adaptive/issues/1341#issuecomment-1971379639

## How has this been tested?

i done have sample streams,
user tested from issue #1341

## Types of change

## Checklist:

### DIFF
--- a/src/codechandler/ttml/TTML.cpp
+++ b/src/codechandler/ttml/TTML.cpp
@@ -356,6 +356,28 @@ void TTML2SRT::StackSubtitle(std::string_view id,
   newSub.end = GetTime(endTime);
   newSub.text = text;
 
+  if (!m_subtitlesList.empty() && !newSub.text.empty())
+  {
+    // This is a workaround for Kodi overlapped subtitles rendering,
+    // when there are multiple lines of text provided by multiple "p" elements
+    // usually with same begin/end timings (overlapping)
+    // the subtitle renderer show on screen these lines of text in "reversed order"
+    // (last added on top, the first on bottom) that is the SRT behaviour,
+    // and causes the display of inverted sentences with TTML.
+    // So TTML multilines should be shown "as is" as it appears in the xml file.
+    // The workaround consists in to merge these lines with same begin/end timing,
+    // but can fail when the end time is different or if "p" elements come from different media segments.
+
+    //! @todo: a better solution should come from the libass library, but it is also possible that converting
+    //! TTML directly to ASS format (and use CDVDOverlayCodecSSA decoder) may solve the problem, need a test.
+    SubtitleData& lastSub = m_subtitlesList.back();
+    if (lastSub.start == newSub.start && lastSub.end == newSub.end)
+    {
+      lastSub.text += "<br/>" + newSub.text;
+      return;
+    }
+  }
+
   m_subtitlesList.emplace_back(newSub);
 }
 


glennguy:
> but can fail when the end time is different or if "p" elements come from different media segments.

I think this scenario is not really that valid right? The solution is much better than the current, thanks for sorting.

glennguy (approved):

CastagnaIT:
is not considered, I'm not so happy about this workaround that should not be implemented here,
but an appropriate solution is currently not possible and so better than nothing...
(also Webvtt can be affected of same problem for same reasons)

MoojMidge:
> > but can fail when the end time is different or if "p" elements come from different media segments.
> 
> I think this scenario is not really that valid right? The solution is much better than the current, thanks for sorting.

This solution is better than the old behaviour, but the scenario where both the begin and end times are different, is valid.

Youtube creates auto-generated subtitles with deliberate overlaps in begin and end times because it expects the subtitles to be displayed on two lines and to roll/crawl up, a feature that is implemented by their various players, but apparently not implemented in libass.

This creates problems because you have the same kind of disorientating display, where every second text chunk is shown in a different position or wrong order.

An example of normal subs:
```
<p begin="00:00:00.066" end="00:00:02.869" style="s2">We just got dropped off<br />in the middle of an abandoned city.</p>
<p begin="00:00:02.969" end="00:00:04.938" style="s2">And we're going to spend the next<br />seven days here.</p>
<p begin="00:00:04.971" end="00:00:06.406" style="s2">And there goes our only way out.</p>
```

And the equivalent auto-generated subs:
```
<p begin="00:00:00.080" end="00:00:03.320" style="s2">we just got dropped off in the middle of</p>
<p begin="00:00:01.680" end="00:00:05.319" style="s2">an abandoned city and we're going to</p>
<p begin="00:00:03.320" end="00:00:06.960" style="s2">spend the next 7 days here and there</p>
<p begin="00:00:05.319" end="00:00:09.759" style="s2">goes our only way out we are now</p>
```

A workaround that works for these type of subs, as well as the overlapping subs in https://github.com/xbmc/inputstream.adaptive/issues/1341#issuecomment-1971379639, would be to change
https://github.com/xbmc/inputstream.adaptive/pull/1499/commits/299402c7d269261f66db77b82bb6cad788f881f5#diff-dc78d222194c853feaf38252c2e4cdf22f4ec36db926cdd808473f911c7bd7caR374
```
    if (lastSub.start == newSub.start && lastSub.end == newSub.end)
```

to
```
    if (newSub.start < lastSub.end)
```


PS. sorry for the late comment, should I make a new issue instead?

CastagnaIT:
as i said above this is literally a hack and its wrong to include it here
i agreed to include this PR only because it is limited exclusively on cues with same begin/end timings

the proposed change invalidates the original cue timings and can lead to other side effects, i will make no further changes here

the right thing to do is:
1) try to do an attempt by converting/using ASS format
2) investigate for a fix on kodi side, so a solution has to be found with the renderer libass, since this affect not only ttml but also webvtt and more likely other formats

MoojMidge:
I agree that this is not a perfect workaround however as far as I can see, any fix that is attempted by converting to ASS will also need to similarly modify the original cue timings, because the subtitles won't move unless specifically formatted to do so.

The only situation where the original timings won't need to be modified is when the two begin and two end times are exactly the same, as already catered for in this PR, because the second subtitle line does not need to move, it is just completely replaced along with the first subtitle line, by the next (third) subtitle line.

Where they don't exactly overlap, what essentially needs to be done is to interleave the two overlapping subtitles lines (let's call them line A and line B) together, so you end up with three new lines:

1. Line A_new
    - content: line A original content
    - begin time: line A original begin time
    - end time: line B original begin time
2. Line A_B
    - content: line A and B joined together
    - begin time: line B original begin time
    - end time: line A original end time
    - format: optional ASS override tags (move + clip?)
4. Line B_new
    - content: line B original content
    - begin time: line A original end time
    - end time: line B original end time

Then the next line would need to be checked to see whether it overlaps with line B_new, and the process repeated.

MoojMidge:
In comparison to interleaving the lines, simply joining them, if the end and begin times of consecutive lines overlap, means that line B is potentially shown for a shorter duration than originally intended, but is at least shown in the correct position/order.

In the situation where the line B end time has been specifically extended to allow it to move, there is little/no actual decrease in duration that the line is displayed.

Where there can be an issue is if line B duration has not been extended for visual presentation purposes, but overlaps based on what is happening in the video/audio i.e. descriptive subtitles, background music, multiple people talking over themselves etc. 

Regardless these types of overlaps are currently going to result in extremely difficult to read subtitles with the position/order switching from line to line.

Joining based on the overlapping end/begin time at least gives the viewer half a chance to (quickly) read the subtitles without being disoriented.

MoojMidge:
Just to add, Youtube appears to do the following when converting ttml subs to vtt subs:

1. Line A_new_cue
    - content: line A original content, new cue per word
    - begin time: line A original begin time
    - end time: line B original begin time - 0.01s
2. Line A_new_full
    - content: line A original content
    - begin time: line B original begin time - 0.01s
    - end time: line B original begin time
3. Line A_B_cue
    - content: line A original content, line B appended with new cue per word
    - begin time: line B original begin time
    - end time: line A original end time - 0.01s
4. Line B_new_full
    - content: line B original content
    - begin time: line A original end time - 0.01s
    - end time: line A original end time
5. Line B_C_cue
    - content: line B original content, line C appended with new cue per word
    - begin time: line C original begin time
    - end time: line B original end time - 0.01s
6. Line C_new_full
    - content: line C original content
    - begin time: line B original end time - 0.01s
    - end time: line B original end time

And so on

CastagnaIT:
as said before no workarounds will be add or changed in addition to what is already do this PR in a limited way,
if you want you can investigate to achieve a proper fix that can be used across all affected subtitles types

MoojMidge:
Sure I can have a look, but wanted to clarify what you consider a workaround vs a proper fix. Also is it OK to create draft PRs for the purpose of generating test builds?

IMO converting to ASS is not a proper/required fix. It would add significant complexity, for marginal benefit (better formatting and other fancy ASS features), but still require changing the original cue times.

If changing cue times is not an option then the only solution is to rework how Kodi renders overlays, and realisticly that is not going to happen without having significant impacts on Kodi, as a whole across various platforms.

If changing cue times is an acceptable compromise, then as mentioned there are two ways to do it:
1. combine the subs in a naive way based on `next start time < previous end time`, which is effective for the vast majority of subs impacted by this issue because it just reverses the process used to automatically create the overlapping subs in the first place
2. interleave the subs (which Youtube, as a reference, appears to do). I assume this would be preferred over the simpler/naive approach?

Regarding handling other subtitle formats, would the preference be to handle this in the individual codec handlers, duplicating the same thing for TTMLCodecHandler and WebVTTCodecHandler, or try to implement this at the SampleReader level?

CastagnaIT:
> IMO converting to ASS is not a proper/required fix. It would add significant complexity

you should learn more about ISA interfaces and kodi decoders, because is the exact opposite, creating ASS demuxer packets the data will be directly handled by the renderer without an additional conversion step == lower computation

> If changing cue times is an acceptable compromise, then as mentioned there are two ways to do it:
> Regarding handling other subtitle formats, would the preference be to handle this in the individual codec handlers, duplicating the same thing for TTMLCodecHandler and WebVTTCodecHandler, or try to implement this at the SampleReader level?

nothing of changes you are talking will be accepted to do in InputStream Adaptive
invalidating timing is not acceptable solution
if you really want your hacks you can do it for your own use
here i lock PR

PS. since the problem is in the Kodi subtitles renderer/decoders, and not here, the first right thing to do is open an Issue on xbmc repository, so as to highlight the problem and taken into consideration

PR #1508, opened 2024-03-11
[DashTree] Rework to improve live streaming

CastagnaIT:
## Description

Its difficult to explain every single change here, i would end up writing a book!
So i try to summarize most important changes/problems

Unfortunately while i was fixing one thing it was breaking another, this due to wrong or missing implementations done in the past, since everything is connected there is no way to create separate small PRs you need to include all changes together

1) Nasty bug! on segment_buffers_ uses
https://github.com/xbmc/inputstream.adaptive/blob/df8b32db36cbf4e2e14441ebe533d547ebf3c947/src/common/AdaptiveStream.cpp#L882-L883
`segment_buffers_` is unrelated to MPD manifest updates! and can lead to wrong next segment!
because between MPD updates old segments can disappears from the timeline
and new one added, this will produce misaligned segment_number saved in to segment_buffers_
because the stored segment_number is not aware of MPD updates and keep the reference
to the initial (start number) segment position referred to oldest manifest!
with segment_buffers_ you need to find the next segment by compare segment startPTS.

2) Mess of segment timeline update on `CDashTree::RefreshLiveSegments`
https://github.com/xbmc/inputstream.adaptive/blob/df8b32db36cbf4e2e14441ebe533d547ebf3c947/src/parser/DASHTree.cpp#L1750-L1885
All this code was a lot messy, i have removed all, by looking one another video player source code, was checking to align updated segments by using segment start PTS, and not the start number as was done here. This should also allow more clean implementation for the misalignment workaround that before was force changing the start number, this was causing problems with some live streams and make working some others (like facebook), force change start number is a killer operation to do that may lead problems on many source code parts, often if you works with multiple periods

3) Nasty multiple bugs in the segment alignment code on representation segments (template) generation
https://github.com/xbmc/inputstream.adaptive/blob/df8b32db36cbf4e2e14441ebe533d547ebf3c947/src/parser/DASHTree.cpp#L1084-L1099
from the earliest versions of ISA i never understood this unclear code part, so it was used to adjust the start number for live streams, but...
after tons of try and check, i understood that:
3.1) Bug! this move the start number for the current live edge position, was ok, but if you move the number you must change also the start PTS! this last one was missing, and least to hidden weird behaviours
3.2) Bug! you can align the initial segment, but dont do it for all periods! this will kill the other period timeline! leading to unclear weird behaviours
3.3) Bug! you can align the initial segment, but if you dont know the period duration you will generate segments that can overlap up to and beyond the next period! with live streaming you need to wait the MPD update to know the period duration, so limit to generate the 1st segment and leaves the task to InsertLiveSegment
3.4) Bug! Dont align with MPD updates it its needed only for playback start or will affect the right periods timeline

4) Removed m_allowInsertLiveSegments workaround
the solution was if the SegmentTemplate provide the timeline dont insert live segments

5) InsertLiveSegment bug with multiple periods
"insert live segment's" when was used, was a non ending operation, that was preventing switch to the next period.
I found a way to limit the "live segment insert" that consist in to fetch the duration from MPD update during playback, and so i have add a code to check on-fly that inserted segment dont exceeds the period duration.
Also i tried to move the InsertLiveSegment call to the AdaptiveStream::ensureSegment instead of use Session,
this IMO will result in a more clean code

6) Reworked in full generation of segments for SegmentTemplate, there were wrong parts

7) `CSegment` `startPTS_` now is always referred to be within the period Start pts

8) ~Add support to MPD having SegmentTemplate and PresentationTimeOffset~ changes reverted/removed

9) Removed all Youtube outdated specific code about `$START_NUMBER$` on manifest update URL for MPD update and all code workarounds (`inputstream.adaptive.manifest_update_parameter` or `inputstream.adaptive.manifest_upd_params`)

10) `GetFragmentInfo` on `CSession::OnSegmentChangedRead` has been full removed, this because `InsertLiveSegment` is now done inside `AdaptiveStream::ensureSegment` that allow to remove specific code on Session like `m_hasSegmentChanged`

11) `AdaptiveStream::waitingForSegment` has been simplified and removed the weird `checkTime` var, i could not understand the reason for checking GetStreamType, where imo the only thing to take in account is make sure that available_segment_buffers_ is empty, if someone has some info please let me know, from my part i havent found regressions (i tested also HLS stream with audio "included" to video)

12) HLS add support to `EXT-X-PROGRAM-DATE-TIME` for live streams, this has been mandatory since now the `CSegment` `startPTS_` use PTS within the period, and so you need the datetime anchor (when possible)

13) some code TODO's added, for future cleanups

14) changes to gtest's. Some tests has been adjusted due to segment startPts fixes, two cases that may require some explanation are:
`DASHTreeTest updateParameterLiveSegmentStartNumber` removed was outdated youtube test
`DASHTreeAdaptiveStreamTest, MisalignedSegmentTimeline` i have reflected the changes to have the real segment timing as explained of old PR #724

15) Changes in to `CSubtitleSampleReader::ReadSample`, done because it was calling two times ensureSegment and was not so nice, so i have add a way to get the full data from segment buffer.

## Motivation and context

I was investigating to fix the problem with `m_allowInsertLiveSegments` workaround
that since very old ISA versions was always handled in bad way,
i took time to try understand things that was not working good and i endedup finding a lot of bugs...

## How has this been tested?

This has been tested mainly with these live MPD,each one has different behaviour:
https://livesim.dashif.org/livesim/periods_20/testpic_2s/Manifest.mpd (+periods+insert segs)
https://d3cglye48cl6m2.cloudfront.net/out/v1/563ec3e9487340528616d736e1202232/index.mpd (+periods)
https://htv-rrs.mts.ru/PLTV/88888888/224/3221228154/3221228154.mpd (+insert segs, idr if there are periods maybe rare)
facebook live, mpd not linked find it on website flow (it expire) (NO insert segs, but with MPD updates old segments can disappears from timeline that can lead to  1° bug and others)

## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -129,20 +129,6 @@ bool adaptive::CDashTree::Open(std::string_view url,
   return true;
 }
 
-void adaptive::CDashTree::PostOpen()
-{
-  AdaptiveTree::PostOpen();
-
-  // If live delay is lower than the manifest update period
-  // there are not enough segments to ensure playback, so allow insert live segments.
-  if (m_isLive && m_liveDelay <= m_minimumUpdatePeriod)
-    //! @todo: to investigate, its possible that the appropriate way is not enable
-    //! m_allowInsertLiveSegments right here, but add a more check to AdaptiveStream::ensureSegment
-    //! where when there are no more segments we try force request a manifest update,
-    //! and so if the manifest update dont provide new segments allow insert live segments.
-    m_allowInsertLiveSegments = true;
-}
-
 bool adaptive::CDashTree::ParseManifest(const std::string& data)
 {
   xml_document doc;
@@ -207,8 +193,8 @@ bool adaptive::CDashTree::ParseManifest(const std::string& data)
   // - If "duration" attribute is missing, but "start" attribute, use this last one to calculate the duration
   // - If both attributes are missing, try get the duration from a representation
 
-  uint64_t totalDuration{0}; // Calculated duration, in seconds
-  double mpdTotalDuration = m_mediaPresDuration; // MPD total duration, in seconds
+  uint64_t totalDuration{0}; // Calculated duration, in ms
+  uint64_t mpdTotalDuration = m_mediaPresDuration; // MPD total duration, in ms
   if (mpdTotalDuration == 0)
     mpdTotalDuration = m_timeShiftBufferDepth;
 
@@ -226,22 +212,25 @@ bool adaptive::CDashTree::ParseManifest(const std::string& data)
     auto nextPeriod = itPeriod + 1;
     if (nextPeriod == m_periods.end()) // Next period, not found
     {
-      if (period->GetStart() != NO_VALUE && mpdTotalDuration > 0)
+      if (!IsLive())
       {
-        period->SetDuration(static_cast<uint64_t>((mpdTotalDuration * 1000 - period->GetStart()) *
-                                                  period->GetTimescale() / 1000));
-      }
-      else // Try get duration / timescale from a representation
-      {
-        CAdaptationSet* adp = CAdaptationSet::FindByFirstAVStream(period->GetAdaptationSets());
-        if (adp)
+        if (period->GetStart() != NO_VALUE && mpdTotalDuration > 0)
         {
-          auto& rep = adp->GetRepresentations()[0];
-          if (rep->GetDuration() > 0)
+          uint64_t durMs = mpdTotalDuration - period->GetStart();
+          period->SetDuration(durMs * period->GetTimescale() / 1000);
+        }
+        else // Try get duration / timescale from a representation
+        {
+          CAdaptationSet* adp = CAdaptationSet::FindByFirstAVStream(period->GetAdaptationSets());
+          if (adp)
           {
-            period->SetDuration(rep->GetDuration());
-            period->SetTimescale(rep->GetTimescale());
-            totalDuration += rep->GetDuration() / rep->GetTimescale();
+            auto& rep = adp->GetRepresentations()[0];
+            if (rep->GetDuration() > 0)
+            {
+              uint64_t durMs = rep->GetDuration() * 1000 / rep->GetTimescale();
+              period->SetDuration(durMs * period->GetTimescale() / 1000);
+              totalDuration += durMs;
+            }
           }
         }
       }
@@ -250,8 +239,8 @@ bool adaptive::CDashTree::ParseManifest(const std::string& data)
     {
       if (period->GetStart() != NO_VALUE && (*nextPeriod)->GetStart() != NO_VALUE)
       {
-        period->SetDuration(static_cast<uint64_t>((*nextPeriod)->GetStart() - period->GetStart()) *
-                            period->GetTimescale() / 1000);
+        uint64_t durMs = (*nextPeriod)->GetStart() - period->GetStart();
+        period->SetDuration(durMs * period->GetTimescale() / 1000);
       }
       else // Try get duration / timescale from a representation
       {
@@ -261,9 +250,9 @@ bool adaptive::CDashTree::ParseManifest(const std::string& data)
           auto& rep = adp->GetRepresentations()[0];
           if (rep->GetDuration() > 0)
           {
-            period->SetDuration(rep->GetDuration());
-            period->SetTimescale(rep->GetTimescale());
-            totalDuration += rep->GetDuration() / rep->GetTimescale();
+            uint64_t durMs = rep->GetDuration() * 1000 / rep->GetTimescale();
+            period->SetDuration(durMs * period->GetTimescale() / 1000);
+            totalDuration += durMs;
           }
         }
       }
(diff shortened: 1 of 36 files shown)


CastagnaIT:
hi @MoojMidge im partial rework dash support for live stream
and im stuck on a kind of exclusive ISA feature for Youtube DASH manifest
and so i need an your feedback to prevent broken some kind of your addon functionality

im talking of the use of `inputstream.adaptive.manifest_update_parameter` (or new `inputstream.adaptive.manifest_upd_params`) used with placeholder `$START_NUMBER$`
https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptivemanifest_update_parameter

its used to pass as argument the next segment number in the url to request live manifest update,
is this nowaday still used? or has it been removed from youtube backend?

In case it has been removed, i proceed to the complete removal of this exclusive youtube code
If not, I need you explain me way to test its operation since need to be reworked

MoojMidge:
@CastagnaIT - It can be removed. There are currently no specific url or query parameters required for the manifest updates to work. Normal full manifest updates work with a `$Number$` template now.

`<SegmentTemplate presentationTimeOffset="80326043" startNumber="19863" timescale="1000" media="sq/$Number$">`

CastagnaIT:
@MoojMidge thank you for feedback, so i can make code much more simpler

glennguy:
Wow looking forward to this.

DASH live refreshes have been totally broken for multi period streams since day 1 of multi period implementation. Doing the live HLS was hard enough, thinking about mpd gave me headaches!

CastagnaIT:
i need to investigate better about startPts/time set to generated segments
i realized that with a my recent code change i feel i have broken something

CastagnaIT:
@glennguy i have finally finished...
- i have reverted PTO uses since i have misinterpreted its use (should be used to sync streams at demuxer level)
- i have reverted all changes that have replaced `startPTS_` with `m_time`

i found additional bugs that affect pts on demuxers that will be not addressed here
i will add a comment to highlight it

src/samplereader/TSSampleReader.cpp
```diff
@@ -69,6 +69,18 @@ AP4_Result CTSSampleReader::ReadSample()
...
+    //! so the substraction "m_pts - m_ptsOffs" it is not a clear thing.
+    //! There is also something weird on HLS discontinuities (multiple chapters/periods)
+    //! where after a discontinuity the packet pts is lower than the last segment of previous period/discontinuity
+    //! this cause a VP resync e.g:
+    //!   debug <general>: CVideoPlayer::CheckContinuity - resync backward :1, prev:587175999.000000, curr:577170666.000000, diff:-10005333.000000
+    //!   debug <general>: CVideoPlayer::CheckContinuity - update correction: -10026666.000000
+    //! not sure if this is correct or not (tested with pluto-tv)
+    //! This code is present also on the others sample readers, that need to be verified
     m_dts = (GetDts() == PTS_UNSET) ? STREAM_NOPTS_VALUE : (GetDts() * 100) / 9;
     m_pts = (GetPts() == PTS_UNSET) ? STREAM_NOPTS_VALUE : (GetPts() * 100) / 9;
```
CastagnaIT, line 85: pts/dts operations to me is full wrong, here and more likely also on all other demuxers/samplereaders
that result in a wrong pts/dts fed to kodi

src/parser/DASHTree.cpp
```diff
@@ -1021,127 +975,127 @@ void adaptive::CDashTree::ParseTagRepresentation(pugi::xml_node nodeRepr,
...
-      if (adpSet->HasSegmentTimelineDuration())
-      {
-        segmentsCount = adpSet->SegmentTimelineDuration().GetSize();
+          do
+          {
+            CSegment seg;
+            seg.startPTS_ = time;
+            // If no PTO, the "t" value on <SegmentTimeline><S> element should be relative to period start
+            if (!hasPTO)
+              seg.startPTS_ += periodStartScaled;
```
CastagnaIT, line 1011: particular note to this `if (!hasPTO)` that add period start pts to segment pts,
this is to fix this sample stream:
```
#KODIPROP:inputstream.adaptive.license_key=https://lic.staging.drmtoday.com/license-proxy-widevine/cenc/?specConform=true||R{SSM}|
https://d24rwxnt7vw9qb.cloudfront.net/v1/dash/e6d234965645b411ad572802b6c9d5a10799c9c1/All_Reference_Streams//6e16c26536564c2f9dbc5f725a820cff/index.mpd
```
usually segments have pts within the period, but this sample stream has some periods where segments pts are not within the period, I don't really understand why and I don't have enough info on how to deal with this.
maybe this sample stream is broken but i have no idea, atm i have no more sample stream to make a comparation so idk...

to me this sound more a workaround but i dont know, anyway its easy to remove, we'll see what happens

src/samplereader/TSSampleReader.cpp
glennguy: Hmmm if the test was on pluto tv the segment lengths are 10 seconds, and the resync is 10 seconds (plus looks like difference between video/audio)... maybe old value is sent initially after discontinuity?

I still remember there is a bug (IMO) in VP when demuxread can't fill the internal 8s buffer fast enough on a period change, which causes VP to think it's out of sync by the same amount of time the entire stream has been playing. But that doesn't look like the case here. Could be related though? Something not being reinitialised to 0 in VP on period change and VP takes care of it this way usually?

src/parser/DASHTree.cpp
glennguy: If by 'segment pts' you mean the pts of the sample data then this stream is probably trying to simulate having ads inserted which would usually never line up with PTS, compared to streams that just have 'chapters'. I don't know if this helps...

glennguy:
> @glennguy i have finally finished...
> 
>     * i have reverted PTO uses since i have misinterpreted its use (should be used to sync streams at demuxer level)
> 
>     * i have reverted all changes that have replaced `startPTS_` with `m_time`
> 
> 
> i found additional bugs that affect pts on demuxers that will be not addressed here i will add a comment to highlight it

Thanks, I reviewed last week so this saves a lot of time knowing what changed

glennguy (approved):
Awesome!

src/samplereader/TSSampleReader.cpp
CastagnaIT: > Could be related though? Something not being reinitialised to 0 in VP on period change and VP takes care of it this way usually?

imo its different if you place a log in to "m_pts" value
on pluto you can see for example (random values)
packet pts 10000
packet pts 20000
packet pts 30000
---- discont - ads
packet pts 10000
packet pts 20000
---- discont - return to movie
packet pts ... a more higher value, now i dont remember exaclty if continue from last of 30000

idk if there is a way to intercept in right way these disc changes to prevent forward/backward of the VP that cause the resync
I have little info to understand if this is correct or if it needs proper fix

src/parser/DASHTree.cpp
CastagnaIT: > If by 'segment pts' you mean the pts of the

its from manifest, example:
```xml
<Period id="43047706_1" start="PT25610H40M24.076S">
       <AdaptationSet bitstreamSwitching="false" frameRate="25/1" mimeType="video/mp4" segmentAlignment="true" startWithSAP="1" subsegmentAlignment="true" subsegmentStartsWithSAP="1">
            <SegmentTemplate startNumber="1" timescale="90000"/>
            <Representation bandwidth="5000000" codecs="avc1.4d4028" height="1080" id="4" width="1920">
                <SegmentTemplate initialization="asset_1080_5_2init.mp4" media="asset_1080_5_2_$Number%09d$.mp4" startNumber="1" timescale="90000">
                    <SegmentTimeline>
                        <S d="180000" t="0"/>
```
you can see that "t" timestamp of the segment its "0" instead to be within the priod start
usually the "t" timestamp should be as follow:

```xml
<Period duration="PT86.600S" id="43047663" start="PT92198337.476S">
        <AdaptationSet bitstreamSwitching="true" id="792039359" mimeType="video/mp4" segmentAlignment="true" startWithSAP="1" subsegmentAlignment="true" subsegmentStartsWithSAP="1">
            <SegmentTemplate initialization="index_video_$RepresentationID$_0_init.mp4?m=1704732393" media="index_video_$RepresentationID$_0_$Number$.mp4?m=1704732393" presentationTimeOffset="4609916873831" startNumber="43047692" timescale="50000">
                <SegmentTimeline>
                    <S d="100000" r="13" t="4609919673831"/>
```
"t" == 4609919673831
4609919673831 / timescale (50000) = 92198393,476
period start is: 92198337,476
you can see that fall in the period

since in the example above the "t" is "0" segment fall in a bad timing
and so when you check for last segment with `AdaptiveTree::IsLastSegment` is not possible detect the last segment due to wrong segment pts, if you try to remove this code line, you can see that cause a playback problem

but this sample stream play bad also on web player testers, so idk if this is a sample stream bug or something i don't know/understand yet...

CastagnaIT: i do the PR merge and release, we will see how to improve this when nevagive feedback will come

PR #1518, opened 2024-04-03
[DashTree] Restored TSB and fix period start

### LINKED ISSUE #574: [MPD] Multiple-period live starting at end of 1st period instead of end of 2nd period
with attached mpd, IA will start playback at the end of the 1st period instead of the end (live) of the 2nd period. 

[master.mpd.txt](https://github.com/peak3d/inputstream.adaptive/files/5683917/master.mpd.txt)
[kodi.log](https://github.com/peak3d/inputstream.adaptive/files/5691789/kodi.log)

1st period = segments 1 to 891
2nd period = segments 892 to live

You will see from the log, IA starts playing at segment number 888 (end of 1st period)

### LINKED ISSUE #1066: Inputstream Adaptive plays the first chapter instead of live in a live stream
### Describe the bug
Trying to play the live of the french channel TF1,  stream lasts about 4 hours and has several chapters. Usually, the live is not the beginning of the file, but the end.
So, stream is 4 hours behind live.

## Expected Behavior
Inputstream plays the current live, not 4 hours behind.

## Actual Behavior

Stream is always about 4 hours behind live.

## Possible Fix
Sorry, I don't know how to fix that. As a workaround, one can move the cursor at the end of the stream to have the live actually.

### To Reproduce
1. Live is available here: https://www.tf1.fr/tf1/direct. Be careful, stream is France geoblocked. And you need a free subsciption to watch the live from the site.
2. Here is a python code to adapt to check by yourself what happens (still geoblocked, but no needed subscription)
```python
from __future__ import unicode_literals
from builtins import str
import json

import inputstreamhelper
from codequick import Route, Resolver, Listitem, Script

import inputstreamhelper
import inputstream
from kodi_six import xbmcgui, xbmcplugin
import urlquick

URL_VIDEO_STREAM = 'https://mediainfo.tf1.fr/mediainfocombo/L_TF1?context=MYTF1&pver=5000002&platform=web&device=desktop&os=linux&osVersion=unknown&topDomain=www.tf1.fr'

URL_LICENCE_KEY = 'https://drm-wide.tf1.fr/proxy?id=L_TF1|Content-Type=&User-Agent=Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3041.0 Safari/537.36&Host=drm-wide.tf1.fr|R{SSM}|'

json_parser = urlquick.get( URL_VIDEO_STREAM, GENERIC_HEADERS, max_age=-1).json()

item = Listitem()
item.path = json_parser['delivery']['url']
item.label = get_selected_item_label()
item.art.update(get_selected_item_art())
item.info.update(get_selected_item_info())
item.property[inputstream] = 'inputstream.adaptive'
item.property['inputstream.adaptive.manifest_type'] = 'mpd'
item.property['inputstream.adaptive.license_type'] = 'com.widevine.alpha'
item.property['inputstream.adaptive.license_key'] = URL_LICENCE_KEY

return item
```

### Debuglog

[LOG.TXT](https://github.com/xbmc/inputstream.adaptive/files/9915410/LOG.TXT)

### MPD/M3U8s/ISM
[index.mpd.zip](https://github.com/xbmc/inputstream.adaptive/files/9915386/index.mpd.zip)

### Screenshots 

### Your Environment
Used Operating system:
 - [] Android
 - [ ] iOS
 - [ ] tvOS
 - [x] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Kodi version: 20.alpha2
 - inputstream.adaptive version:20.2.0
 I tested with the 20.3.1 version with the same result.

### LINKED ISSUE #1517: [K21, DASH, live stream with time shifting buffer] playback start from first period not last (like VOD content)
### Describe the bug

From ver. 21.4.4 ISA play live stream with time shifting buffer like VOD content.

## Expected Behavior

Playback live stream with time shifting buffer from last (current) period with the ability to change the period during playback.

## Actual Behavior

playback live stream from wrong period.

### To Reproduce

1.  Playback live stream (with time shifting buffer)
2.
3.

### Debuglog

[log.txt](https://github.com/xbmc/inputstream.adaptive/files/14818927/log.txt)

### MPD/M3U8s/ISM

[manifest_1711939990.txt](https://github.com/xbmc/inputstream.adaptive/files/14818929/manifest_1711939990.txt)

### Screenshots 

Current behavior:
TVN nieprawidłowo
Expecting behavior:
tvn_prawidłowo

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Windows 10 22H2, kernel: Windows NT x86 64-bit version
 - Kodi version: 22.0-ALPHA1 (21.90.700) Git:20240326-e70f0d83f3
 - inputstream.adaptive version: 21.4.4

CastagnaIT:
## Description

Continue the dash fixes from last rework
changes are focused to fix the TSB (timeshift buffer) on MPDs that provide the SegmentTemplate without segments
with the previous rework PR i had not correctly interpreted how a part of the code works, moreover there were too many variables to test

anyway i found a better solution than how it was handled TSB before and hopefully clearer
within my human limits certainly in the future there will be other changes to be made all this is not perfection but it works

Changes:
- Handled TSB between multiple periods
- Added limit for TSB size, since TSB can be very large it may happen that we generate segments for tens or hundreds of hours... so limited to 4 hours to avoid excessive memory consumption (shaka player do something similar). New property to override default value see below
- Fix wrong start period (was playing the first period/chapter instead of the relative NOW period)
- Add 3 new test cases for TSB

### RapidJSON dependency
i have add tarball from last commit (at today) instead use the release tarball version 1.1.0,
because the master branch has more required improvements,
most important support for c++17 and arm64 that allow us to avoid to add custom patches

### Introduced new property `inputstream.adaptive.manifest_config`
this property is designed to be extensible to avoid add new properties each time,
the accepted value is a JSON dictionary with following structure:
```json
{ "config_name": "value", "config_name2": true, "config_name3": 1500, ... }
```
Currently supported config:
- Config name `timeshift_bufferlimit`, value in seconds as number, allow to override the default limit of 4 hours of TSB
example: `inputstream.adaptive.manifest_config={"timeshift_bufferlimit":120}`
- On future PR `live_delay` can be moved here

**About question https://github.com/xbmc/inputstream.adaptive/pull/1176#issuecomment-1451801747, about possible broken relations about "period id" <> "stream id"**
By testing https://livesim.dashif.org/livesim/periods_20/testpic_2s/Manifest.mpd
you can see that when you play the video at the time when you have two periods, and NOW time falls in the second period, that is, playback starts from the second period, you can see that the stream ID is set exactly starting from 2000 instead of 1000, follow log:
```
2024-04-04 08:00:49.060 T:28420    info <general>: AddOnLog: inputstream.adaptive: Manifest successfully parsed (Periods: 2, Streams in first period: 2, Type: live)
2024-04-04 08:00:49.061 T:28420   debug <general>: AddOnLog: inputstream.adaptive: Created AdaptiveStream [AS-0] with adaptation set ID: "", stream type: video
2024-04-04 08:00:49.061 T:28420   debug <general>: AddOnLog: inputstream.adaptive: Created AdaptiveStream [AS-1] with adaptation set ID: "", stream type: audio
2024-04-04 08:00:49.134 T:28420   debug <general>: AddOnLog: inputstream.adaptive: GetStreamIds()
2024-04-04 08:00:49.134 T:28420   debug <general>: AddOnLog: inputstream.adaptive: GetStream(2001)
2024-04-04 08:00:49.135 T:28420   debug <general>: AddOnLog: inputstream.adaptive: GetStream(2002)
2024-04-04 08:00:49.182 T:28420   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 2001 with codec_id 27
2024-04-04 08:00:49.182 T:28420   debug <general>: CDVDDemuxClient::RequestStream(): added/updated stream 2002 with codec_id 86018
2024-04-04 08:00:49.197 T:28420    info <general>: Opening stream: 2001 source: 256
2024-04-04 08:00:49.197 T:28420   debug <general>: AddOnLog: inputstream.adaptive: OpenStream(2001)
```
imo the stream ID is correctly related to the period, if that was not the purpose of the question please reformulate

## 

Motivation and context

fix #1517
fix #1066
fix #574

## How has this been tested?

mainly with mpd above and
http://mcdn.daserste.de/daserste/dash/manifest.mpd (geolock germany)

## Types of change

## Checklist:


CastagnaIT:
@glennguy i requested review, but i forgot to test the new property, i will do later or tomorrow

EDIT: tested ok

mtr81:
About timeshift buffer... From now ISA ignores "timeShiftBufferDepth" tag in DASH manifest and shortens this period to 2 hours? I think it's not good solution, because in some services TV stations have different time shift buffer periods and sametimes there's no information about the length of these periods in api - only in manifests. This leads to opening manifests before playback begins to setting manifest_config property. 
In addition, there is illogical information on the counter (default settings):
c_plus_k21

CastagnaIT:
before this PR and older reworks, ISA never supported a change of timeShiftBufferDepth between manifest MPD updates,
so the timeShiftBufferDepth has always been a fixed value, taken from the first manifest only
with this PR i do not want to make any further new improvements, can be done separately

what shown on GUI screenshot was already happening from old ISA versions its not new thing,
this can happens for example on manifests that include template segments

perhaps it may seem strange but its how works currently ISA,
the GUI time to the left is current played time, the GUI time on the right of "/" is the total stream duration, in this case the timeshift depth. So if the value on the right would be equal to the value on the left will means from your screenshot ~8 hours of TSB...

> I think it's not good solution, because in some services TV stations have different time shift buffer periods and sametimes there's no information about the length of these periods in api - only in manifests. This leads to opening manifests before playback begins to setting manifest_config property

If a service does not have a common timeshift depth for all his live channels, then you can find a compromise for a good value or leave to the user the choice, or full disable that limit by set a very higher value to `timeshift_bufferlimit`,
but at least you are aware of the possible problems that can happens...
I would never want to generate segments for a timeshift that may be several days, more weeks or longer,
would end up generating _thousands or millions of segments_, that on a low end devices will cause system stability or also crashes due to OOM 

this TSB limit also exists on other web players... with even lower value...
for example theorizing:
5h with segments of 1secs == 18000 objects * ~1KB = 18MB
mostly depends on segment size,
if 2 hours seems too low value limit, we can increase it
i would avoid values higher than 4 or 5h

src/parser/DASHTree.cpp
```diff
@@ -124,7 +124,27 @@ bool adaptive::CDashTree::Open(std::string_view url,
...
+  //! can potentially fall on the previous period and so its needed to set in advance the right period
+  //! so live delay and period selection could be merged
+  //! to do this its needed to test also the behaviour of HLS streams with discontinuities
+  uint64_t now = stream_start_ - available_time_;
+  if (m_isLive && !kodiProps.IsPlayTimeshift())
+  {
+    for (auto& period : m_periods)
+    {
+      if (period->GetStart() != NO_VALUE && now >= period->GetStart())
+        m_currentPeriod = period.get();
```
glennguy, line 140: worth putting  a break here?

glennguy (approved):

glennguy:
> imo the stream ID is correctly related to the period, if that was not the purpose of the question please reformulate

My comments in the other PR really just related to the special case of HLS. The PR was wanting to use the manifest supplied period ID rather than IA generated one, and my note was that HLS period ID is always 1000 to start with because of master manifest and at that time we don't know what periods there are. The code quoted is a workaround for that.

CastagnaIT:
> My comments in the other PR really just related to the special case of HLS. The PR was wanting to use the manifest supplied period ID rather than IA generated one, and my note was that HLS period ID is always 1000 to start with because of master manifest and at that time we don't know what periods there are. The code quoted is a workaround for that.

hmm now i think to understand, so for HLS we will see in the future how to do

PS: updated PR just to increase default TSB limit value to 4h

src/parser/DASHTree.cpp
CastagnaIT: no because you need to find the last period that fall in to "now" not the first one

Issue #1419, opened 2023-11-15
Disney+ - no subtitles on resume


matthuisman:
not working on both kodi 20 and 21
as per https://github.com/matthuisman/slyguy.addons/issues/500

i have confirmed the same. 
if i start normally - subtitles are ok
if I resume - subtitles dont work. They are selectable just dont show anything

Log
First play is no resume.
turn on subs and they display ok
2nd play resume
subs are turned on but nothing shown

[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/13361398/kodi.log.txt)
[master.m3u8.txt](https://github.com/xbmc/inputstream.adaptive/files/13361410/master.m3u8.txt)
[subs.m3u8.txt](https://github.com/xbmc/inputstream.adaptive/files/13361409/subs.m3u8.txt)
[video.m3u8.txt](https://github.com/xbmc/inputstream.adaptive/files/13715440/video.m3u8.txt)
[audio.m3u8.txt](https://github.com/xbmc/inputstream.adaptive/files/13715443/audio.m3u8.txt)

maybe related to X-DISCONTINUITY ?

matthuisman:
we did have https://github.com/xbmc/inputstream.adaptive/issues/836 but i thought that got fixed?
or just a new issue

matthuisman:
Kodi 19 works. 

Kodi 19 + IA 19.0.7
Remember subs enabled = no
Resume + subs = **YES**

Kodi 20 + IA 20.3.14
Remember sub enabled = no
Resume + subs = no

Kodi 21 + IA 21.3.1
Remember sub enabled = yes
Resume + subs = no

CastagnaIT:
with issue https://github.com/xbmc/inputstream.adaptive/issues/836
there was more problems where one has been closed by linked fix PR
but at that time no one has tested with D+, my subscription was expired
and no one had a subscription to make tests
i was thinking that the sample stream was similar because of similar problems

one of the problems was with DEMUX_SPECIALID_STREAMCHANGE, has not been solved
(this in some situations kodi core could reset audio/subtitle settings without trying keep the used settings like language, on/off... while change chapter, or quality change ISA "Adaptive" stream selector, a very nonsense kodi behaviour...)

but here you affirm

> if I resume - subtitles dont work. They are selectable just dont show anything

the log show dropping packets 
[...]

PR #1522, opened 2024-04-07
[SubtitleSampleReader] Restored old pts workaound

CastagnaIT:
## Description

This problem has been hidden since early ISA versions,
but it was highlighted when i reworked Dash parser and introduced also EXT-X-PROGRAM-DATE-TIME on HLS
the main problem is on DTSorPTS() comparison

https://github.com/xbmc/inputstream.adaptive/blob/a82e3eb8333187d8bb346bf5eca384524a553e57/src/Session.cpp#L1067-L1069

we are comparing audio/video/subtitles,
for audio/video the pts/dts come from demuxer, but subtitles use pts from manifest or from subtitle data "cue" timing
these values not always are comparable with audio/video, because pts/dts that come from demuxer packet data
can be different and makes this package selection totally ineffective

other details on comments:
https://github.com/xbmc/inputstream.adaptive/issues/1507#issuecomment-2029867689
https://github.com/xbmc/inputstream.adaptive/issues/1507#issuecomment-2028774719

this PR for now i restore the old behaviour by removing the period start pts from segment pts
this will result that the subtitle "packet" pts will have most of time always a lower pts compared to audio/video
and then will be processed immediately without any control

an appropriate solution will be needed in future
since i think that weird thing can happens anyway
for example with all uses of lastPts variable:

https://github.com/xbmc/inputstream.adaptive/blob/a82e3eb8333187d8bb346bf5eca384524a553e57/src/main.cpp#L396

that need to be at same time re-checked

## Motivation and context

"fix" https://github.com/xbmc/inputstream.adaptive/issues/1419#issuecomment-2041299336
"fix" https://github.com/xbmc/inputstream.adaptive/issues/1507#issuecomment-2028488867

## How has this been tested?

D+

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1065,6 +1065,12 @@ bool CSession::GetNextSample(ISampleReader*& sampleReader)
       {
         if (AP4_SUCCEEDED(streamReader->Start(isStarted)))
         {
+          //!@ todo: DTSorPTS comparison is wrong
+          //! currently we are compare audio/video/subtitles
+          //! for audio/video the pts/dts come from demuxer, but subtitles use pts from manifest
+          //! these values not always are comparable because pts/dts that come from demuxer packet data
+          //! can be different and makes this package selection ineffective
+          //! see also workaround on CSubtitleSampleReader::ReadSample
           if (!res || streamReader->DTSorPTS() < res->GetReader()->DTSorPTS())
           {
             if (stream->m_adStream.waitingForSegment())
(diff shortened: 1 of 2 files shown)


MauriceW67:
Downloaded the version with the fix from here:

https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1522/1/artifacts

Can confirm Disney+ is showing subtitles again with this version. Thanks @CastagnaIT  👍

glennguy (approved):
Changes are fine.  We have to think carefully regarding how to solve the problem outlined in todo comments

PR #1532, opened 2024-04-15
[HLSTree] New hls_ignore_endlist, hls_fix_mediasequence, hls_fix_discsequence params to fix faulty live services

### LINKED ISSUE #1507: Aborting playback in the event of commercial breaks or X-discontinuity with the inputstream.adaptive addon
### Describe the bug
Hi @CastagnaIT I noticed that the stream from DAZN Darts x PlutoTV simply stops after a certain time and you end up back in the Kodi main menu. It doesn't matter which Kodi version is used. According to the responsible developer @matthuisman , this is due to the lack of full support for ad breaks or X-discontinuity in the inputstream.adaptive addon. (https://github.com/matthuisman/slyguy.addons/issues/673#issuecomment-1987384548)
For the same stream: https://pluto.tv/de/live-tv/64b67f0424ade50008a3be17, playback via browser works without any problems.

## Expected Behavior
The stream should run through without stopping after a certain time.

## Actual Behavior

The stream from DAZN Darts x PlutoTV simply stops and ends up back in the Kodi main menu.

### To Reproduce

1. Start Kodi
2. Play Stream: https://r.mjh.nz/PlutoTV/64b67f0424ade50008a3be17-alt.m3u8 with IPTV Simple Client (with activated option to use HLS streams via the inputstream.adaptive Addon under "Advanced")

### Debuglog

https://github.com/matthuisman/slyguy.addons/files/14551853/kodi.log

### MPD/M3U8s/ISM

https://r.mjh.nz/PlutoTV/64b67f0424ade50008a3be17-alt.m3u8

### Screenshots 

### Your Environment
Used Operating system:

 - [X] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Operating system version/name: Android 10 / Windows 10
 - Kodi version: 18/19/20/21.0-BETA3
 - inputstream.adaptive version: inputstream.adaptive v21.4.3 or similar

CastagnaIT:
## Description

Plutotv its a good example of faulty live service that violates the HLS specifications in different ways (not all specified here)
by providing malformed data on variant child manifests:

EXT-X-DISCONTINUITY-SEQUENCE: has inconsistent number between manifest updates, for example on first downloaded manifest could be of value 2, on the next update become 0 or 1. A decreasing number is not allowed on sane HLS.

EXT-X-MEDIA-SEQUENCE: has inconsistent number between manifest updates, that cause impossibility to keep track of the sequence of segments between updates, i couldn't figure out if there is a logical pattern, because the numbers seem pretty random...

EXT-X-ENDLIST: make a wrong use of EXT-X-ENDLIST tag, sometimes the service send a manifest update without segments, and specifying this tag although the live stream is not ended, thereby forcing the stop of playback

EXT-X-PROGRAM-DATE-TIME: sometime there is not a precise value, so on first downloaded manifest there is X PTS value, on next manifest update X varies by some seconds, or also they replace the original value with others, this leads to any kind of side effect since the fixes are based on this

To "fix" this bad situation i have add these new parameters to `inputstream.adaptive.manifest_config` property:

`hls_ignore_endlist`: allow to ignore EXT-X-ENDLIST, there are two situations
1) At playback start, can be sent the first manifest without segments and with EXT-X-ENDLIST
2) While in playback, can be sent the manifest update without segments and with EXT-X-ENDLIST
for the case 2, to mitigate buffering problems the time to perform the next manifest update is temporarily halved

`hls_fix_mediasequence`: allow to try correct a malformed EXT-X-MEDIA-SEQUENCE value, the corrected value is determined by finding the corresponding segment in the updated playlist referring to the PTS value of the segment provided by the EXT-X-PROGRAM-DATE-TIME tag. so to use this paramenter it is strictly required that the manifest provides EXT-X-PROGRAM-DATE-TIME tag

`hls_fix_discsequence`: allow to try correct a malformed EXT-X-DISCONTINUITY-SEQUENCE value, the corrected value is determined by checking whether a segment falls within an existing period if found use that sequence number to fix EXT-X-DISCONTINUITY-SEQUENCE in order to work EXT-X-PROGRAM-DATE-TIME tag is needed.

**However malformed manifest update can lead to any kind of playback oddities, such as segments played multiple times, or periods switched before the end of their playback, due to the impossibility of determining the right sequence number and segment to be played. In the worst case it will cause unexpected problems with Kodi VideoPlayer by freezing the playback either temporarily (a few seconds) or permanently.**

## Motivation and context

fix #1507 
## How has this been tested?

```
#KODIPROP:inputstream.adaptive.manifest_config={"hls_ignore_endlist":true,"hls_fix_mediasequence":true,"hls_fix_discsequence":true}
```
https://r.mjh.nz/PlutoTV/64b67f0424ade50008a3be17-alt.m3u8

played for more times until to catch ADS discontinuities

## Types of change

## Checklist:

### DIFF
--- a/src/CompKodiProps.cpp
+++ b/src/CompKodiProps.cpp
@@ -262,6 +262,18 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseManifestConfig(const std::string& dat
       if (jDictVal.GetUint() > 0)
         m_manifestConfig.timeShiftBufferLimit = jDictVal.GetUint();
     }
+    else if (configName == "hls_ignore_endlist" && jDictVal.IsBool())
+    {
+       m_manifestConfig.hlsIgnoreEndList = jDictVal.GetBool();
+    }
+    else if (configName == "hls_fix_mediasequence" && jDictVal.IsBool())
+    {
+       m_manifestConfig.hlsFixMediaSequence = jDictVal.GetBool();
+    }
+    else if (configName == "hls_fix_discsequence" && jDictVal.IsBool())
+    {
+       m_manifestConfig.hlsFixDiscontSequence = jDictVal.GetBool();
+    }
     else
     {
       LOG::LogF(LOGERROR, "Unsupported \"%s\" config or wrong data type on \"%s\" property",
--- a/src/CompKodiProps.h
+++ b/src/CompKodiProps.h
@@ -45,7 +45,17 @@ struct ChooserProps
 
 struct ManifestConfig
 {
-  std::optional<uint32_t> timeShiftBufferLimit; // Limit the timeshift buffer depth, in seconds
+  // Limit the timeshift buffer depth, in seconds
+  std::optional<uint32_t> timeShiftBufferLimit;
+  // Faulty HLS live services can send manifest updates with inconsistent EXT-X-ENDLIST
+  // when the stream is not finished, enabling this will ignore EXT-X-ENDLIST tags
+  bool hlsIgnoreEndList{false};
+  // Faulty HLS live services can send manifest updates with inconsistent EXT-X-MEDIA-SEQUENCE
+  // enabling this will correct the value by using EXT-X-PROGRAM-DATE-TIME tags
+  bool hlsFixMediaSequence{false};
+  // Faulty HLS live services can send manifest updates with inconsistent EXT-X-DISCONTINUITY-SEQUENCE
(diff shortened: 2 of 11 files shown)


glennguy (approved):
Looks good @CastagnaIT

PR #1558, opened 2024-05-17
DTS codec name should be "dca"?

matthuisman:
## Description

Trying to get DTS-X working on Disney+ (https://github.com/matthuisman/slyguy.addons/issues/742)

I notice on Nexus the codecname was dca:
https://github.com/xbmc/inputstream.adaptive/blob/Nexus/src/Session.cpp#L880
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StreamUtils.cpp#L27

HLS
```
#EXT-X-MEDIA:TYPE="AUDIO",GROUP-ID="DTS-X",NAME="English",LANGUAGE="en",AUTOSELECT="YES",CHANNELS="10",URI="r/composite_448k_dtsx_en_PRIMARY_bc81d2d4-da6d-42d5-92ee-2810badb8d94_f50c42ab-1461-4efa-b36d-124d4da9487e.m3u8",DEFAULT="YES"
#EXT-X-STREAM-INF:BANDWIDTH=5653092,AVERAGE-BANDWIDTH=3039650,CODECS="hvc1.2.4.L93.90,dtsx",RESOLUTION=1280x720,FRAME-RATE=24,VIDEO-RANGE=SDR,HDCP-LEVEL=TYPE-0,CHARACTERISTICS="com.dss.ctr.hd",AUDIO="DTS-X",SUBTITLES="sub-main"
r/composite_3500k_CENC_CTR_FHD_SDR_4c288d85-d15a-401d-a644-f38b64d3328b_818720d3-7fad-4f62-ab81-73e9d85fcf83.m3u8
```

IA is detecting the stream as if i use "blah" instead of "dtsx", it defaults to AAC.
So, it is detecting it - but just KODI doesnt show any stream for it.

## Motivation and context
Get Disney+ new DTS-X working

## How has this been tested?
Windows x64
i dont get any audio but its now showing as a audio track.
If i select it, kodi shows "dca" as audio codec with 10 channels.
But the audio stream just shows as unknown with no audio

Before this change:
image

after this change:
image

## Types of change

## Checklist:

### DIFF
--- a/src/utils/Utils.h
+++ b/src/utils/Utils.h
@@ -79,7 +79,7 @@ constexpr std::array VIDEO_NAME_LIST = {NAME_MPEG1, NAME_MPEG2, NAME_MPEG4, NAME
 // Audio definitions
 
 constexpr const char* NAME_AAC = "aac";
-constexpr const char* NAME_DTS = "dtshd_ma";
+constexpr const char* NAME_DTS = "dca";
 constexpr const char* NAME_AC3 = "ac3";
 constexpr const char* NAME_EAC3 = "eac3"; // Enhanced AC-3
 constexpr const char* NAME_OPUS = "opus";
--- a/src/utils/Utils.h
+++ b/src/utils/Utils.h
@@ -79,7 +79,7 @@ constexpr std::array VIDEO_NAME_LIST = {NAME_MPEG1, NAME_MPEG2, NAME_MPEG4, NAME
 // Audio definitions
 
 constexpr const char* NAME_AAC = "aac";
-constexpr const char* NAME_DTS = "dts";
+constexpr const char* NAME_DTS = "dca";
 constexpr const char* NAME_AC3 = "ac3";
 constexpr const char* NAME_EAC3 = "eac3"; // Enhanced AC-3
 constexpr const char* NAME_OPUS = "opus";
--- a/src/utils/Utils.h
+++ b/src/utils/Utils.h
@@ -79,7 +79,7 @@ constexpr std::array VIDEO_NAME_LIST = {NAME_MPEG1, NAME_MPEG2, NAME_MPEG4, NAME
 // Audio definitions
 
 constexpr const char* NAME_AAC = "aac";
-constexpr const char* NAME_DTS = "dca";
+constexpr const char* NAME_DTS = "dtshd_ma";
 constexpr const char* NAME_AC3 = "ac3";
 constexpr const char* NAME_EAC3 = "eac3"; // Enhanced AC-3
 constexpr const char* NAME_OPUS = "opus";


matthuisman:
@kyl416 are you able to test this build on your Shield with DTS-X receiver?
There are builds here: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1558/3/artifacts

CastagnaIT:
imo this is wrong
SetCodecName require to have ffmpeg defines, "dca" dont exist on ffmpeg there is only "dts"
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/codec_desc.c#L2712

then if codec var is the same (i havent check atm) the https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StreamUtils.cpp#L27 looks wrong

matthuisman:
kodi core is broken then and ignoring "dts" ?

matthuisman:
if you still have disney+ login, the latest version of the addon adds support to get the manifest with DTS-X (enable in settings)

CastagnaIT:
yes i still have your access should still works
i will check 
do you some movie/tvshow title that have dtsx?
the problem could be derived from tsreader

> kodi core is broken then and ignoring "dts" ?

atm i cant give you an answer

OT: i noticed that D+ have so much better audio quality than N€tflix where audio is more muffled also on the dialogues

matthuisman:
https://press.disneyplus.com/news/disney-plus-queen-rock-montreal-imax-enhanced-sound-dts

Ive been testing with "QUEEN ROCK MONTREAL" but there are some marvel with it too

matthuisman:
maybe we need to be using dtshd_ma ?

CastagnaIT:
~dtsx its different than dts hd ma~ edit: i remember wrong i think
i think you will need to wait ffmpeg bump:
https://github.com/xbmc/xbmc/pull/24972
but i will check better next days

matthuisman:
ok. you think with new ffpmeg - IA will work without this change? I guess i can see if i can find a build of kodi with that ffmpeg 7

https://forum.kodi.tv/showthread.php?tid=373441 / https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/98a4699216fbc3a4e382ba6a10026eab39b5214e

CastagnaIT:
> ok. you think with new ffpmeg - IA will work without this change? 

now remember a bit more, when time ago i cleanedup codecs on ISA Omega
i havent implemented codecs profiles not yet supported, such as DTSCodecProfileHDMAX

these should be implemented on session.cpp, but also on demuxers that require it (e.g. tsreader)
but since isnt yet merged the new ffmpeg on kodi, imo is a bit useless
for sure we can prepare it for the future new K22 "P" branch
but should not works, at least not completely on Omega branch

for sure need to investigate the root of problem that cause AAC to be shown instead of DTS

matthuisman:
thank you :)
Will wait for ffmpeg 7.0 to go in and go from there

CastagnaIT:
i found some time to made a search, and i found that ffmpeg internally use "dca" despite "dts" its used as codec name on the codec list, i did not expect a discrepancy of names since the codec list is declared that way

so i think its a valid change restore it if you can reopen the PR

anyway i dont think however dtsx can be decoded,
but maybe you have some chance that works by using audio passthrough, i cant test it my amplifier dont support it

kyl416:
It might be something specific to the Disney+ audio stream that it can't handle.

DTSX is backwards compatible so even if you don't have something that supports DTS:X, it should still decode as DTS-HD MA without the DTSX enhancements. i.e. I used one of the DTSX samples here and converted it to self hosted HLS and DASH manifests and was able to get DTS-HD MA audio with the dca patch:
https://www.demolandia.net/cinema/dts-demo-trailers-hd/page-6.html

matthuisman:
i hade deleted the fork so had to create a new PR:
https://github.com/xbmc/inputstream.adaptive/pull/1559

PR #1594, opened 2024-07-01
[Cleanups] Get rid of CSpinCache / new segment container

CastagnaIT:
## Description

I think it's time to get rid of `CSpinCache`
then create a kind of container for segments where we can move all segment methods which are also on Representation.h on the container itself,
a starting point where to have a common place where manage segments not a final solution but an intermediate step (ideally it would be ideal to have the interface for manifests _tree_ detached from the data of streams, but needs too much changes)

i have leave untouched the `CRepresentation::current_segment_`
i think can be possible to move it to the new CSegContainer but you need to check some parts of code well (more likely for live) here there are already many changes and i would like to avoid possible regressions

## Motivation and context

a long-delayed cleanup,
this also helps to have a way to know the period duration "on the fly" without the need to do calculations separately for each use case along the parsers

## How has this been tested?

couple of HLS/Dash streams

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -938,7 +938,7 @@ void CSession::PrepareStream(CStream* stream)
 
   // Prepare the representation when the period change usually its not needed,
   // because the timeline is always already updated
-  if ((!m_adaptiveTree->IsChangingPeriod() || !repr->Timeline().IsEmpty()) &&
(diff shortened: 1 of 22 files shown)


glennguy (approved):
Looks good!

PR #1607, opened 2024-07-22
[Session] Use total segments duration to check the period

### LINKED ISSUE #1500: [Sky Sport Now] - possible chapter seek problem (missing last chapter duration?)
it doesnt switch the baseurl when going to the next period.
Always uses the baseurl from the first period (<BaseURL>exchange219026anubr_219026_1600/**hZrPcUpQ4**/</BaseURL>)

Log is me playing the 24 hour buffer live stream from the start.
It plays ok.
I then seek to future periods and start getting 404 due to it still trying to use the first periods baseurl

inputstream.adaptive v21.4.3
(doesnt seem to work on kodi 20 either)

[manifest_1709440931.txt](https://github.com/xbmc/inputstream.adaptive/files/14471802/manifest_1709440931.txt)
[kodi.log.txt](https://github.com/xbmc/inputstream.adaptive/files/14471804/kodi.log.txt)

CastagnaIT:
## Description

On the issue ISA its not able to switch to the right period when you try to do a video seek
manifest case: [segtpl_multiperiod_ads.mpd.txt](https://github.com/user-attachments/files/16334134/segtpl_multiperiod_ads.mpd.txt)

from the manifest the ADS periods (marked with `urn:scte:scte35:2013:xml`) dont have the "duration" attribute but only "start" attrib.
with the parser we calculate the period duration by looking for the "start" attrib of the next period (by difference),
but if you calculate the duration, you can see that dont match with the total segments duration (of SegmentTemplate),
this seem to lead to in a segments hole in part of the period

It seems that Dash when periods are reported as ADS the period duration is not exact but estimated, but i have not yet found enough info to confirm this.

Then since there are parts of code that use "Period->GetDuration" lead to a wrong behaviour, as you can see on `CSession::SeekTime` try find the chapter on a wrong time

my idea is instead looking to each period duration, use total segments duration, that must be a safe value

todo:
1) need a feedback, i cant test it by myself atm since my only MPD multiperiod TV streaming is completely dead...
2) check for some regressions?
3) it would be useful to have confirmation from other streams that Dash period ADS have estimated and not precise durations

## Motivation and context

attempt to fix #1500
## How has this been tested?

NOT TESTED!

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -12,6 +12,7 @@
 #include "CompSettings.h"
 #include "SrvBroker.h"
 #include "aes_decrypter.h"
+#include "common/AdaptationSet.h"
 #include "common/AdaptiveDecrypter.h"
 #include "common/AdaptiveTreeFactory.h"
 #include "common/Chooser.h"
@@ -1120,7 +1121,7 @@ bool CSession::SeekTime(double seekTime, unsigned int streamId, bool preceeding)
 
   for (; pi != m_adaptiveTree->m_periods.cend(); pi++)
   {
-    chapterTime += double((*pi)->GetDuration()) / (*pi)->GetTimescale();
+    chapterTime += double((*pi)->GetSegDuration()) / (*pi)->GetTimescale();
     if (chapterTime > seekTime)
       break;
   }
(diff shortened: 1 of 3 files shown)


CastagnaIT:
superseeded by #1867

PR #1625, opened 2024-08-03
Fix build error with gcc 14

bkuhls:
## Description
This PR fixed a build error with gcc 14
```
src/decrypters/Helpers.h:55:13: error: 'uint8_t' was not declared in this scope
   55 | std::vector<uint8_t> ConvertKidStrToBytes(std::string_view kidStr);
```
## Motivation and context
Fix a build error with gcc 14

## How has this been tested?
Build-tested with gcc 14

## Types of change

## Checklist:

### DIFF
--- a/src/decrypters/Helpers.h
+++ b/src/decrypters/Helpers.h
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 #include <string_view>
 #include <vector>


jpalus:
Note that there lots of other places that use such int types but do not include `<cstdint>` directly. This change is enough to fix build though.

CastagnaIT (approved):
thank you
i will add a CI for GCC14 to avoid future problems

PR #1639, opened 2024-08-17
[Session] Reworked widevine pssh generation for smoothstreaming

CastagnaIT:
## Description

The `CreateISMlicense` code was used to construct only a single part of the PSSH data (the WidevinePsshData structure that has been never specified) that also was not including the PSSH box.

I reworked whole code, in a very simplified way

by verifing the PSSH generated by using axinom PSSH Box decoder, reported the lack of data from `content_id` field of WidevinePsshData data structure, it could be a bug introduced from Kodi 20 (i have not check kodi<=19), now fixed

axinom PSSH Box decoder also report a lack of `protection_scheme` field, but i found constraining docs on the net about this field, where often is signed as not mandatory field, i made some attempts to add it but i failed to constructing the correct protobuf for that field, due to different data type

i removed the limitation where the `{KID}` is must appears before the `{UUID}`.

i have add a little check to prevent that if a PSSH box is set on license_data property cannot be injected on the `content_id` field, but will be used directly.

_Signed as backport needed, to do at later time after merged, i want to see if there are complaining, should not because the data is correct, i think there are few addons that use this SmoothStreaming mixture_

--

Add also two more commits that i used to debug it

## Motivation and context

With the DRM config rework i was looking to this ugly code
that was needed some cleanups in advance to understand how better proceeds

The use case for this code is:
Use a SmoothStreaming manifest, PlayReady protected, with a service that provide also the Widevine license server url
then the configuration to allow ISA to generate a Widevine PSSH for SS will be as:
```
#KODIPROP:mimetype=application/vnd.ms-sstr+xml
#KODIPROP:inputstream.adaptive.license_key=https://widevinelicenseserverurl
https://playready.server.com/media/example.ism.smoothstreaming/manifest
```
optionally is possible provide set a custom `content_id` field data by using license_data prop:
```
#KODIPROP:inputstream.adaptive.license_data="your base64 data" (placeholders if used must be encoded base64 with the raw data)
```

Plus on the inputstream.adaptive.license_data, for this use case, can be used two placeholders to inject data to the custom data:
`{KID}` to inject the KID as bytes
`{UUID}` to inject the KID as UUID string
a limitation: the `{KID}` is must appears before the `{UUID}` or the generation fails
a problem: if a user try to set a real PSSH box on the license_data, will cause unexpected behaviours

all this has been never documented on Wiki due to missing test streams

## How has this been tested?

Not tested directly, i forced a playready smoothstreaming, without playback

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -382,7 +382,8 @@ bool CSession::InitializeDRM(bool addDefaultKID /* = false */)
       if (sessionPsshset.adaptation_set_->GetStreamType() == StreamType::NOTYPE)
         continue;
 
-      std::string_view licenseData = CSrvBroker::GetKodiProps().GetLicenseData();
+      const std::vector<uint8_t> defaultKid = DRM::ConvertKidStrToBytes(sessionPsshset.defaultKID_);
+      std::string_view licenseDataStr = CSrvBroker::GetKodiProps().GetLicenseData();
 
       if (m_adaptiveTree->GetTreeType() == adaptive::TreeType::SMOOTH_STREAMING)
       {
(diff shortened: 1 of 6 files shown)


CastagnaIT:
@matthuisman do you by chance have any addons that make use of this, or do you know some?

matthuisman:
hi @CastagnaIT i dont have any addons using smoothstreaming  sorry :(

PR #1649, opened 2024-08-21
[DashTree] Improved clearkey defaultkid workaround

### LINKED ISSUE #1646: [Feature request] Reproduce mpd manifests that don't have default:kid.
### Describe the feature

I was testing the new inputstream clearkey support and I can't reproduce manifests that don't have cenc:default_KID inside. Am I doing something wrong? Or this feature is not yet suppported? Is there any way to force the player to load the KID?

### Motivation

Increase the compatibility for every kind of mpd manifest.

Yes

No

CastagnaIT:
## Description

Better handle missing kid, this will cover use case of manifest provided on #1646
it works because the manifest provide the playready scheme where the kid is extracted from the data header, then if a manifest dont provide playready will fails

I will make no further changes for this because its needed implementing in right way the kid extraction from mp4 box, that will cover the other use cases, but this at least from my part will be done in future but only after the DRM property rework

--

commit 2: fix corrupted playback when you play HLS with unsupported encryption, so kodi player still starts, for example because there is video encrypted but subtitles are unencrypted and are hoverer initialized, the kodi demuxer starts read in any case the subitles packets, nothing prevent to stop streams

## Motivation and context

closes #1646
## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -65,7 +65,7 @@ CSession::CSession(const std::string& manifestUrl) : m_manifestUrl(manifestUrl)
 CSession::~CSession()
 {
   LOG::Log(LOGDEBUG, "CSession::~CSession()");
-  m_streams.clear();
+  DeleteStreams();
   DisposeDecrypter();
 
   if (m_adaptiveTree)
@@ -77,6 +77,12 @@ CSession::~CSession()
 
   delete m_reprChooser;
   m_reprChooser = nullptr;
+}
+
+void SESSION::CSession::DeleteStreams()
+{
+  LOG::Log(LOGDEBUG, "CSession::DeleteStreams()");
+  m_streams.clear();
 }
 
 void CSession::SetSupportedDecrypterURN(std::vector<std::string_view>& keySystems)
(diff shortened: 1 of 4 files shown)


CastagnaIT:
@glennguy can you take a look on last commit? [Force delete streams, when decrypter is not initialized](https://github.com/xbmc/inputstream.adaptive/pull/1649/commits/283b24d927f096830e45b871bd9d35bf259789f3)

glennguy:
Does GetStream returning false cause Kodi to stop playback altogether? If so then this should be good

CastagnaIT:
> Does GetStream returning false cause Kodi to stop playback altogether?

no, the current behaviour of kodi core is never stop playback...
if you return false on GetStream only the specified stream ID will be interrupted

if a single GetStream fails (return false) kodi dont stop playbackbut will continue to process the available streams, if any

that's why i had to force delete all streams on `m_streams`, without `m_streams` nothing works and kodi is forced to terminate the playback since DemuxRead callbacks will not works anymore (always return nullptr)

PR #1653, opened 2024-08-23
[Session][CDM] Dont initialize already initialized sessions

CastagnaIT:
## Description

Fix a very old bug, at least i assume it as a bug, where each session on `m_cdmSessions` are initialized more times

`InitializeDRM` can be called more times depends on use cases:
1) CSession::Initialize->InitializePeriod->InitializeDRM - Used by DASH/SS  --> single call
2) CInputStreamAdaptive::DemuxRead->m_session->InitializePeriod()->InitializeDRM - On chapter change --> single call
3) CInputStreamAdaptive::OpenStream->m_session->PrepareStream->InitializeDRM - Used by HLS --> a call for each stream

the `m_cdmSessions` is every time resized but existing sessions are permanent
and when they are once time initialized, when `InitializeDRM` is called again for any reason, the loop 
`for (size_t ses{1}; ses < m_cdmSessions.size(); ++ses)`
will "reinitialize" every session

i find a bit weird this thing, and i assume that its a bug

this can be clearily visible with this widevine HLS sample stream
```
#KODIPROP:mimetype=application/vnd.apple.mpegurl
#KODIPROP:inputstream.adaptive.license_key=https://cwip-shaka-proxy.appspot.com/no_auth||R{SSM}|
https://storage.googleapis.com/shaka-demo-assets/angel-one-widevine-hls/hls.m3u8
```
Here the log before the changes [log_before.txt](https://github.com/user-attachments/files/16729470/log_before.txt)
Here the log after the changes [log_after.txt](https://github.com/user-attachments/files/16729478/log_after.txt)

you can see that the old behaviour sessions are initialized 4 times !!
look at `Initializing stream with KID:` log prints

---

while i was writing this, a question arose to me,
what happens if at chapter/period change the pssh change and KIDs change?
seem that at least for DASH usually the protection between periods should be the same (in theory)
but if change, by following the code on InitializeDRM, the decrypter is kept and seem nothing do a real reinitialization,
i think this is an unhandled situation that can lead to problems

---

backport later times, just to see that there are no new bad feedbacks

## Motivation and context

I have never given too much weight to this until now that i am making big changes
while playing with ClearKey i noticed a weird thing
multiple `Initializing stream with KID:` was printed on log where should not, so investigated better

## How has this been tested?

see above
## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -381,6 +381,17 @@ bool CSession::InitializeDRM(bool addDefaultKID /* = false */)
     // cdmSession 0 is reserved for unencrypted streams
     for (size_t ses{1}; ses < m_cdmSessions.size(); ++ses)
     {
+      CCdmSession& session{m_cdmSessions[ses]};
+
+      // Check if the decrypter has been previously initialized, if so skip it,
+      // sessions are collected and never removed and InitializeDRM can be called more times
+      // depending on how it is used:
+      // 1) CSession::Initialize->InitializePeriod->InitializeDRM - Used by DASH/SS (single call)
+      // 2) CInputStreamAdaptive::DemuxRead->m_session->InitializePeriod()->InitializeDRM - On chapter change (single call)
+      // 3) CInputStreamAdaptive::OpenStream->m_session->PrepareStream->InitializeDRM - Used by HLS (a call for each stream)
+      if (session.m_cencSingleSampleDecrypter)
+        continue;
+
       std::vector<uint8_t> initData;
       std::string drmOptionalKeyParam;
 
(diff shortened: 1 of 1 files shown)


glennguy:
> i think this is an unhandled situation that can lead to problems

This is possibly why it is this way...

Protection is usually the same across periods but you can in theory do what you like according to the spec.

Checking if the SSD object is there doesn't tell us whether the correct keys are sitting in the decrypter right?
At the period change level we currently avoid InitializeDRM altogether if psshsets compare the same, so maybe we should be looking to (optionally?) do something similar for the openstream/preparestream call? Do we have enough info to recreate psshsets from cdmsessions?

CastagnaIT:
> Do we have enough info to recreate psshsets from cdmsessions?

I think you mean to reassign the psshset indexes in the representations of period,
if the period changes, the psshset are all new, theorising that the new period has different defaultKeys, then on DRM init we should probably search if there is a session with same defaultKey used for each representation, then update the psshindex on each representation, it's quite a mess since the code will be also rendundant due to multiple calls, 

in the rework that i am making locally, i have totally removed assignment index pssh in the representations
and i was thinking of applying a code that searches a cdm session based on defaultKid and stream type (audio/video)

I think i will not make these changes on the current code was to understand how to proceed mainly on rework

PR #1688, opened 2024-10-07
[FragmentedSampleReader] Create senc when saio/saiz/senc are missing

CastagnaIT:
## Description

specs says:
immagine

senc may be present or not
saio saiz mandatory
now bento require to have at least the senc box in the moof/traf box

i tried create a empty senc box, and the decryption works correctly despite the sample count is not provided
may not cover all use cases, but still we are out of specs

## Motivation and context

fix #1502

## How has this been tested?

```
#KODIPROP:mimetype=application/dash+xml
#KODIPROP:inputstream=inputstream.adaptive
#KODIPROP:inputstream.adaptive.manifest_type=mpd
#KODIPROP:inputstream.adaptive.license_type=com.widevine.alpha
[...]

PR #1710, opened 2024-10-29
[FragmentedSampleReader] Check default_isProtected before add senc

### LINKED ISSUE #1708: [Bug] [ism] problems with playback from v21.5.5/22.1.6
### Describe the problem

Since ISA v21.5.5 / v22.1.6 there are problems with playback DRM protected Microsoft Smooth Streaming Media manifests.  Result you can see in the screenshot below: 
screen

### Possible fix

_No response_

### Steps to reproduce

1. start to playback ism manifest (DRM protected)

### Debug log

[log_K22_ISAv22.1.7.txt](https://github.com/user-attachments/files/17534379/log_K22_ISAv22.1.7.txt)

### Stream manifest file(s)

[manifest_1730039657.txt](https://github.com/user-attachments/files/17534380/manifest_1730039657.txt)

### Additional info

_No response_

### Operating system(s)

Windows

Windows 10 22H2, kernel: Windows NT x86 64-bit version 10.0.19045.5011

v21.5.5 / v22.1.6 / v22.1.7

### Kodi version(s)

21.1 / 22.0-ALPHA1 (21.90.700) Git:20241026-60a52e2caf

CastagnaIT:
## Description

PR #1688 introduced a kind of workaround to add SENC box when missing
this caused a kind of problem with smoothstreaming

smoothstreaming as usual is a bit shitting format
since dont have initialization segment a custom MOOV box is created manually from
https://github.com/xbmc/inputstream.adaptive/blob/c77b4061214de2c41779a1516bab25f93d7c91bf/src/common/AdaptiveUtils.cpp#L76-L78

the problem is a bit wide, i think that FragmentedSampleReader code can be potentially improved
but its impossible to me make a good code cleanup and or a better fix
since atm i dont have at my hand all streams that i can use to test all use cases

this part of code is as cat and mice
https://github.com/xbmc/inputstream.adaptive/blob/c77b4061214de2c41779a1516bab25f93d7c91bf/src/samplereader/FragmentedSampleReader.cpp#L391-L409
we need to check for missing SENC before call `AP4_CencSampleInfoTable::Create` to prevent that Create method to fails
but on Issue thread the smoothstreaming (playready) manifest converted to widevine, must not use `CAdaptiveCencSampleDecrypter`,
since the video stream of the Issue dont have SAIZ/SAIO/SENC,  SENC is so created,
and if i understand correctly this seem to prevent the use of widevine decrypter from `CFragmentedSampleReader::ReadSample()`

as kind of solution i have set TENC `default_isProtected` parameter to 0 that means "not encrypted", this is not so clean solution, but i have no better ideas atm since i have no way to make tests for each use case

anyway TENC `default_isProtected` parameter is not so important to bento4 source code, and also we dont have code that depends from it, then should be a kind of solution that at least does not produce side effects

Sidenote: TENC `default_isProtected` on bento4 allow to have values different from 0 and 1, its weird,
looks like bento4 `default_isProtected` param support is a sort of custom implementation that dont follow the MP4 standard
at least im not able to find similar specs

## Motivation and context

fix #1708 
## How has this been tested?

user tested his SS video
and some segments are here [IssueSegments.zip](https://github.com/user-attachments/files/17588907/IssueSegments.zip)

another use case is from sample provided by PR #1688

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveUtils.cpp
+++ b/src/common/AdaptiveUtils.cpp
@@ -133,7 +133,10 @@ AP4_Movie* PLAYLIST::CreateMovieAtom(adaptive::AdaptiveStream& adStream,
       defaultKid = DRM::ConvertKidStrToBytes(psshSet.defaultKID_);
 
     AP4_ContainerAtom schi{AP4_ATOM_TYPE_SCHI};
-    schi.AddChild(new AP4_TencAtom(AP4_CENC_CIPHER_AES_128_CTR, 8, defaultKid.data()));
+    // Note TENC default_isProtected parameter is intentionally set to 0 (not encrypted)
+    // this is to prevent CFragmentedSampleReader::ProcessMoof from trying to create a SENC atom
+    // and so avoid decrypting with a CAdaptiveCencSampleDecrypter
+    schi.AddChild(new AP4_TencAtom(0, 8, defaultKid.data()));
     sampleDesc = new AP4_ProtectedSampleDescription(0, sampleDesc, 0,
                                                     AP4_PROTECTION_SCHEME_TYPE_PIFF, 0, "", &schi);
   }
--- a/src/samplereader/FragmentedSampleReader.cpp
+++ b/src/samplereader/FragmentedSampleReader.cpp
@@ -388,10 +388,24 @@ AP4_Result CFragmentedSampleReader::ProcessMoof(AP4_ContainerAtom* moof,
       if (!m_protectedDesc || !traf)
         return AP4_ERROR_INVALID_FORMAT;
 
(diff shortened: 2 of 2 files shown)


CastagnaIT:
@glennguy no idea if you are still around here, if you want to take a look here

glennguy:
Sorry @CastagnaIT - yes still here :)
Looks like you've done a thorough investigation and without a big test suite all we can do is try and see how it goes

CastagnaIT:
as i thought... you should try to make a list of test streams for each use case, a boredom
but ISA samples test addon could help to make a video list in a new menu

PR #1672, opened 2024-09-04
New DRM Config property implementation rework

CastagnaIT:
## Description

**_I remind everyone that the DRM configuration transition is designed to be backward compatible,_**
with some particular exceptions, specified on "breaking changes" section.

### New `inputstream.adaptive.drm` property
The property has been for now limited to support only a unique DRM configuration.
Support to multiple DRM configurations, will be implemented at later time, with future PR's

Since there is no automatic drm selection yet, it is necessary for now to always add the keysystem with protected streams, even with empty parameters if there is no need to change the default values.

JSON structure example (just a unique key system for now):
```json
{
    "[key system string]": {
        "persistent_storage": bool,
        "secure_decoder_disabled": bool,
        "init_data": str,
        "optional_key_req_params": dict,
        ...
        "license": {
            "server_certificate": str,
            "server_url": str,
            "req_headers": str,
            ...
        }
    }
}
```

In order not to do double work for the documentation, I will write directly on the wiki integration page the instructions on how to use it

Here is an example of a working STRM to compare the old with the new:

OLD PROPERTIES:
```
#KODIPROP:mimetype=application/dash+xml
#KODIPROP:inputstream=inputstream.adaptive
#KODIPROP:inputstream.adaptive.license_type=com.widevine.alpha
#KODIPROP:inputstream.adaptive.license_key=https://cwip-shaka-proxy.appspot.com/no_auth||R{SSM}|R
https://bitmovin-a.akamaihd.net/content/art-of-motion_drm/mpds/11331.mpd
```

NEW DRM PROPERTY:
```
#KODIPROP:mimetype=application/dash+xml
#KODIPROP:inputstream=inputstream.adaptive
#KODIPROP:inputstream.adaptive.drm={"com.widevine.alpha": {"license": {"server_url": "https://cwip-shaka-proxy.appspot.com/no_auth"}}}
https://bitmovin-a.akamaihd.net/content/art-of-motion_drm/mpds/11331.mpd
```
ofc python addons can take the advantage to use python dict and use `json.dumps` to set the "drm" property value

### Add XML license wrapper support
License wrapper term i mean the things that are set in to [inputstream.adaptive.license_key](https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptivelicense_key) [Response-Data] field, where we currently support base64 and JSON, wrappers

Since here the code that manage the license wrapper has been reworked (for widevine), add the XML wrapper support its a very small change, so is included here
fix #267

### Breaking changes
[...]

PR #1165, opened 2023-02-24
[DRM] Revert to single session decryption (when possible)

glennguy:
## Description

Many problems are now being observed from changes relating back to the introduction of CBCS decryption. For this change separate sessions were maintained, partly in an effort to avoid having more Bento4 patches. Now we observe the multiple license requests involved appear to inform service providers that there are multiple concurrent streams playing, causing license rejection after some duration of playback.

Another prior change was to force separate audio/video sessions on Android to resolve 'pixelation' issues. This workaround has worked but again as above has side effects. This issue should be investigated separately and a proper solution found.

## Motivation and context

In an effort to resolve issues experienced in #1147

## How has this been tested?

Runtime tests on CBCS and CENC test streams show no side effects. Artifacts of this PR will be distributed for user testing in particular on Android.

## Types of change

- [x] **Bug fix** (non-breaking change which fixes an issue)
- [ ] **Clean up** (non-breaking change which removes non-working, unmaintained functionality)
- [ ] **Improvement** (non-breaking change which improves existing functionality)
- [ ] **New feature** (non-breaking change which adds functionality)
- [ ] **Breaking change** (fix or feature that will cause existing functionality to change)
- [ ] **Cosmetic change** (non-breaking change that doesn't touch code)
- [ ] **None of the above** (please explain below)

## Checklist:

- [x] I have read the **[Contributing](https://github.com/xbmc/xbmc/blob/master/docs/CONTRIBUTING.md)** document
- [x] My code follows the **[Code Guidelines](https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md)** of this project 
- [ ] My change requires a change to the Wiki documentation
- [ ] I have updated the documentation accordingly

Issue #1147, opened 2023-02-19
DRM-DASH Live Stream fails Android 32 Bit


scottydulton:
### Describe the bug
Here is a clear and concise description of what the problem is:

DRM-DASH live streams are freezing after 5 or so minutes of playback, i cannot reproduce this issue on windows, or my phone Armv8a 64 bit

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

Stream to play without issue.

## Actual Behavior

Stream Freezes after a few minutes

### To Reproduce

1. Use Android Armv7a(32bit) device (im using Fire Tv 4k)
2. Play a live stream on Kayo Sports (i can provide account credentials for testing if need be)
3.

### Debuglog

The debuglog can be found here:
https://pastebin.com/0Kh7y0dr
### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
 

### Screenshots 
Here are some links or screenshots to help explain the problem:

 

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

 

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Windows
 - [ ] Windows UWP

 - Operating system version/name: Android TV 7.1.2 API Level 25 (kernal: Linux 4.4.120+)
 - Kodi version: 20.0
 - inputstream.adaptive version: 20.3.3

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

glennguy:
Hi @scottydulton 

This looks like the issue in the log:
```
CCurlFile::Open - <http://127.0.0.1:52103/https://drm.streamotion.com.au/licenseServer/widevine/v1/streamotion/license> Failed with code 403:
{"code":130401,"message":"Too many concurrent streams."}
```

@matthuisman this must be what you mentioned the other day?

scottydulton:
Hi @glennguy 
yes, however windows and my phone doesnt give that error and playback is fine.

glennguy:
Are there other devices using this same account that are streaming at the same time you're trying to play on your firetv 4k?
It seems to be a mechanism to prevent having the same account play back more than 2/3 streams at the same time i.e to enforce that you're only getting what you pay for.

scottydulton:
[...]

PR #1724, opened 2024-11-09
Implement DRM auto-selection

### LINKED ISSUE #1585: [Bug] ORF ON DRM Streams not working on Android Devices
### Describe the problem

The livestreams for the ORF On Addon cannot be played on Android devices (Tested on Nvidia Shield and Pixel 8 with GrapheneOS). The stream works flawless on Linux/Win

### Possible fix

_No response_

### Steps to reproduce

1. Enable the "Timeshift" option for the ORF ON Addon
2. Open "Livestream" and "ORF 2"

Alternatively use this in a strm file

```
#KODIPROP:inputstream.adaptive.license_url=https://drm.ors.at/acquire-license/widevine
#KODIPROP:inputstream.adaptive.license_key=|Content-Type=application/octet-stream|R{SSM}|R
#KODIPROP:inputstream.adaptive.license_url_append=?BrandGuid=319f2ca9-0d0c-4e5b-bb70-72efae61dad7&userToken=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJraWQiOlsiKiJdLCJleHAiOjE3MjAwMTc4ODZ9.gcYTr_2cKlQYR-OH9Ca1C6DKKhrFsLbY_VVzetwZ2Bo
https://orf2-247.mdn.ors.at/orf/orf2/drmqxa-247/manifest.mpd?offset=5&begin=20240619T141239|user-agent=Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537.36
```

### Debug log

[drm_kodi.log](https://github.com/user-attachments/files/15903551/drm_kodi.log)

### Stream manifest file(s)

[manifest.txt](https://github.com/user-attachments/files/15903563/manifest.txt)

### Additional info

The stream has a geolock so it might be hard to debug without an austrian ip. Let me know if I can help with any other logs/files/etc

### Operating system(s)

Android

Android 11|Android 14

21.4.9

### Kodi version(s)

21.0.1

CastagnaIT:
## Description

This is the last big step of my drm rework

This PR completely rewrites the code of how DRM info provided by manifests are handled, and how DRM/CDM are initialized and managed.

The main goal is to introduce a common place where manage DRM info and CDM so in the DRMEngine,
this avoids the need for each manifest parser to have the task of choosing and managing DRM info (which parsers should not do, as is self-explained by the name) and along this introduce the automatic selection of DRM.

To reach this point, i started working on this 2 years ago, initially with the RFC test PR #1415, then with PR #1672 and in between other minor PRs not mentioned here, thousands of changes were needed to achieve something of acceptable quality with this PR, although there are some poor code parts that require more attention and much more time.

The amount of work to achieve this is very high, and despite my efforts to do dozens of tests, i will never have the opportunity to verify every single use case, so it is possible that this imlementation introduces some bugs that do not allow correct playback. It would be oppurtune to have some feedback from some users who use ISA with some video add-ons, if you read this please give your feedback.

Main changes:

### DRM auto-selection (new)
This requires a manifest that supports multiple DRMs, 
in this case you will have to provide each DRM configuration by using the new property:
`inputstream.adaptive.drm`
as explained on Wiki https://github.com/xbmc/inputstream.adaptive/wiki/Integration-DRM (that i will update later time)
therefore ISA will automatically determine which DRM to use, but you will also be able to give some sort of priority to the type of DRM to be used by making use of the `priority` parameter.

### HDCP (breaking changes)
In my research it seems to me (but i could be wrong since i haven't found comprehensive docs) that a media player should not have a way to handle HDCP, take task is inherent possibly in the DRM (CDM), which should block non-usable Keys and nothing more.
The possibility of managing the rejection of keys seem not so good handled on ISA, and cannot be done on this PR, also because there is a problem in stream ID selection in Kodi core, that prevent a fallback an appropriate stream ID when one fails, which should be fixed before that.

Breaking changes 1:
The ISA setting "Ignore HDCP status" is now removed,
the current behavior is that the HDCP "status" is never verified, but can be re-enabled by using the new `check_hdcp`, read below.

Breaking changes 2:
In the `unwrapper_params`, child parameter of the `license` parameter,
the `path_hdcp` and `path_hdcp_traverse` has been renamed into `path_hdcp_res` and `path_hdcp_res_traverse` (that handle resolution limit),
and there is a new `path_hdcp_ver` and `path_hdcp_ver_traverse`, that handle an HDCP Version comparison with the manifest value.
Details on wiki page https://github.com/xbmc/inputstream.adaptive/wiki/HDCP-resolution-limit

### New property parameters
The `inputstream.adaptive.config` have two new parameters:
- `resolution_limit` - Allow to limit the playable resolutions provided by manifests. This is a kind of workaround since currently we cant handle appropriately HDCP use cases. Just set the max resolution that can be used e.g. "1024x768"
- `check_hdcp` - Allows to enable the check of the HDCP values provided in the license response. Details on wiki page https://github.com/xbmc/inputstream.adaptive/wiki/HDCP-resolution-limit

The `inputstream.adaptive.drm` have a new parameter:
- `force_single_session` - Allow to enforce a single DRM session, when possible. (while with new "drm" prop this is disabled by default, in the old properties and "drm_legacy" still use the enforced single session. more details below)

### DRM configuration changes
1) When a manifest include DRM protections that dont require additional configurations, it will be played directly, so there is no need to set `inputstream.adaptive.drm` or `inputstream.adaptive.drm_legacy` property.
This is the case for example for PlayReady, or ClearKey case that embed its license url in the manifest.

2) In the ClearKey configuration with  `inputstream.adaptive.drm` or `inputstream.adaptive.drm_legacy`, its not only supported to set the license URL, but now its also supported to set a URI with "data" scheme. The data follow the W3C standard, so: `data:application/json;base64,REPLACE_WITH_CK_LICENSE_RESPONSE_AS_BASE64`

### DRM Sessions (breaking change)
With older ISA versons (older kodi versions) sessions created most of time was a single one (when license provide all KIDs) and on android was always enforced into a single session, that was leading to possible decryption problems by displaying corrupted/pixellated videos.

Now by using the new properties to configure the DRM so `inputstream.adaptive.drm` and `inputstream.adaptive.drm_legacy`
the DRM sessions sessions are distinguished by media type and will be reused when:
- They have same init data (default KID)
or when:
- The obtained license support multiple KIDs, but this only applies to DRMs that allow it to be known, for example on Widevine for android this is not possible and so this option will be excluded (means no session reused if there are different initdata/KIDs). 

So if there are multiple DRM sessions opened means also multiple license requests,
if this is a problem for your provider and you are sure that the license response provide all KIDs in a single request,
**you can enforce the use of a single session, by using `force_single_session` parameter of `inputstream.adaptive.drm`,**
this way you can also save the time it takes to request multiple licenses and start playback faster.

_Brief summary session breaking-changes:_
- Nothing is changed if you are using old DRM properties (e.g. inputstream.adaptive.license_key)
- `inputstream.adaptive.drm_legacy` unlike Kodi 21, it no longer forces single session
- `inputstream.adaptive.drm` never force a single session, if not specified by using `force_single_session` parameter

### Stream selector type: Ask quality (improvement)
In multi-codec manifests, all codec variants will now be shown. I had to include this change here out of test necessity.

### Deprecated old DRM properties
All following old DRM properties:
```
inputstream.adaptive.license_type
inputstream.adaptive.license_key
inputstream.adaptive.license_data
inputstream.adaptive.server_certificate
inputstream.adaptive.license_flags
inputstream.adaptive.pre_init_data
```
Are now --officially-- deprecated, in favor of the more advanced methods:
`inputstream.adaptive.drm_legacy` and `inputstream.adaptive.drm`
Its recommended to use the new properties from Kodi v22, see Wiki for more details.
Old properties still work, for now, but will be removed on future Kodi versions.

### Other changes under the hood
- ~Now the initialization of DRM/CDM Sessions are done on streams to be played only. where previously sessions were created for all streams even if unused, this improve also the playback start speed (with the exception of HDCP license based is enabled).~ Changes restored to old behavior due to lack of Widevine CDM Secure decoder for audio media preventing code optimization
- Add support to SmoothStreaming to allow override PlayReady DRM with ClearKey DRM
- Removed HLS parser hack that used DRM objects to store AES data, now the AES data is stored on each segment, with an appropriate cache for downloads
- Fix JSON traverse feature used on JSON license response to find data by path, now correctly traverse also json array type cases

### To do / tests
[x] Rework drm info on Dash parser
[x] Rework drm info on HLS parser
[x] Rework drm info on Smoothstreaming parser
[x] Reimplement CheckHDCP
[x] Adapt test project, if needed
[x] Enable multi-drm support to "drm" property
[x] Deprecate old properties
[x] Test stream with single drm widevine
[x] Test stream with single drm playready
[x] Test stream with clearkey
[x] Test stream with widevine "to" clearkey
[x] Test multi drm's stream widevine+playready, between windows and android
[x] Test HLS AES-128
[x] Test HLS AES-128 - key on each segment
[x] Test HLS multi-period, encryption transitions: unencrypted->AES-128->unencrypted
[x] Some tests on android (tested SS+PR, some widevine)
[x] Some tests on linux (under Ubuntu widevine)

fix #1585 about to separate audio/video sessions

## Motivation and context

This PR will finish the multi-drm configuration support for `inputstream.adaptive.drm` property

### What is auto selection of DRM?
in this case we are talking about manifests (usually HLS/DASH) that support multiple DRMs at same time.
Until now, every python video add-on was forced to manually set the appropriate DRM for the type of operating system in use
by setting appropriate `inputstream.adaptive.license_type` and `inputstream.adaptive.license_key` (and all other properties).
for example a manifest that support playready and widevine,
on Windows OS is needed that the python addon set the drm configuration for widevine, and on android could be for playready.

This implementation allows a video add-ons to provide all the DRM configurations supported by the provider in a distinct way, so that ISA can automatically select a DRM by finding a match between DRM supported in the operating system in use and the manifest, so a video add-on will no longer have to check if a DRM is compatible with the operating system in use, to configure ISA.

This will be even more useful in the future when a support to new DRMs can be implemented.

## Types of change

## Checklist:

### DIFF
--- a/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
+++ b/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
@@ -164,14 +164,7 @@ msgctxt "#30171"
 msgid "Defines the initial bandwidth when it cannot be automatically determined. This value can be overridden by the minimum bandwidth setting."
 msgstr ""
 
-msgctxt "#30172"
-msgid "Ignore HDCP status"
-msgstr ""
-
-#. Description of setting with label #30172
-msgctxt "#30173"
-msgid "Some DRM-protected HD / UHD videos may only be played if the HDCP status is ignored."
-msgstr ""
+#empty strings from id 30171 to 30173
 
 #. To set the stream selection type (refer to RepresentationChooser's)
 msgctxt "#30174"
--- a/inputstream.adaptive/resources/settings.xml.in
+++ b/inputstream.adaptive/resources/settings.xml.in
@@ -161,13 +161,6 @@
           </control>
         </setting>
       </group>
-      <group id="misc">
-        <setting id="HDCPOVERRIDE" type="boolean" label="30172" help="30173">
-          <level>1</level>
-          <default>false</default>
-          <control type="toggle" />
-        </setting>
-      </group>
     </category>
     <category id="expert" label="30120">
       <group id="misc">
--- a/plugin.video.isasamples/menu_data.py
+++ b/plugin.video.isasamples/menu_data.py
@@ -353,7 +353,7 @@ menu_data = {
                     'drm': '{"org.w3.clearkey": {"license": {"keyids": {"feedf00deedeadbeeff0baadf00dd00d": "00112233445566778899aabbccddeeff", "1234f00deedeadbeeff0baadf00dd00d": "8899aabbccddeeff8899aabbccddeeff"}}}}',
                 }
             },
-            'Bitmovin art of motion [widevine to clear key, keys on property]': {
+            'Bitmovin art of motion [WV to CK, keys on property]': {
                 SI_ENCRYPT: 'DRMCK',
                 SI_INFO: 'Override widevine content protection to use clear key',
                 SI_CONFIG: {
@@ -585,6 +585,15 @@ menu_data = {
                     'license_key': 'https://test.playready.microsoft.com/service/rightsmanager.asmx?cfg=(persist:false,sl:150)',
                     'drm_legacy': 'com.microsoft.playready|https://test.playready.microsoft.com/service/rightsmanager.asmx?cfg=(persist:false,sl:150)',
                     'drm': '{"com.microsoft.playready": {"license": {"server_url": "https://test.playready.microsoft.com/service/rightsmanager.asmx?cfg=(persist:false,sl:150)"}}}'
+                }
+            },
+            'Microsoft tears of steel 4k [PR to CK, keys on property]': {
+                SI_ENCRYPT: 'DRMCK',
+                SI_INFO: 'Override PlayReady content protection to use ClearKey',
+                SI_CONFIG: {
+                    'manifest_url': 'https://test.playready.microsoft.com/media/profficialsite/tearsofsteel_4k.ism.smoothstreaming/manifest',
+                    'drm_legacy': 'org.w3.clearkey|6f651ae1dbe44434bcb4690d1564c41c:88da852ae4fa2e1e36aeb2d5c94997b1',
+                    'drm': '{"org.w3.clearkey": {"license": {"keyids": {"6f651ae1dbe44434bcb4690d1564c41c": "88da852ae4fa2e1e36aeb2d5c94997b1"}}}}',
                 }
             },
         },
--- a/src/CompKodiProps.cpp
+++ b/src/CompKodiProps.cpp
@@ -21,12 +21,13 @@
 #include <rapidjson/document.h>
 
 using namespace UTILS;
+using namespace ADP::KODI_PROPS;
 
 namespace
 {
 // clang-format off
-constexpr std::string_view PROP_LICENSE_TYPE = "inputstream.adaptive.license_type"; //! @todo: to be deprecated
-constexpr std::string_view PROP_LICENSE_KEY = "inputstream.adaptive.license_key"; //! @todo: to be deprecated
+constexpr std::string_view PROP_LICENSE_TYPE = "inputstream.adaptive.license_type"; //! @todo: deprecated to be removed on Kodi 23
+constexpr std::string_view PROP_LICENSE_KEY = "inputstream.adaptive.license_key"; //! @todo: deprecated to be removed on Kodi 23
 // PROP_LICENSE_URL and PROP_LICENSE_URL_APPEND has been added as workaround for Kodi PVR API bug
 // where limit property values to max 1024 chars, if exceeds the string is truncated.
 // Since some services provide license urls that exceeds 1024 chars,
@@ -36,9 +37,9 @@ constexpr std::string_view PROP_LICENSE_KEY = "inputstream.adaptive.license_key"
 // -> this problem has been fixed on Kodi 22
 constexpr std::string_view PROP_LICENSE_URL = "inputstream.adaptive.license_url"; //! @todo: deprecated to be removed on Kodi 23
 constexpr std::string_view PROP_LICENSE_URL_APPEND = "inputstream.adaptive.license_url_append"; //! @todo: deprecated to be removed on Kodi 23
-constexpr std::string_view PROP_LICENSE_DATA = "inputstream.adaptive.license_data"; //! @todo: to be deprecated
-constexpr std::string_view PROP_LICENSE_FLAGS = "inputstream.adaptive.license_flags"; //! @todo: to be deprecated
-constexpr std::string_view PROP_SERVER_CERT = "inputstream.adaptive.server_certificate"; //! @todo: to be deprecated
+constexpr std::string_view PROP_LICENSE_DATA = "inputstream.adaptive.license_data"; //! @todo: deprecated to be removed on Kodi 23
+constexpr std::string_view PROP_LICENSE_FLAGS = "inputstream.adaptive.license_flags"; //! @todo: deprecated to be removed on Kodi 23
+constexpr std::string_view PROP_SERVER_CERT = "inputstream.adaptive.server_certificate"; //! @todo: deprecated to be removed on Kodi 23
 
 constexpr std::string_view PROP_COMMON_HEADERS = "inputstream.adaptive.common_headers";
 
@@ -53,7 +54,7 @@ constexpr std::string_view PROP_STREAM_HEADERS = "inputstream.adaptive.stream_he
 constexpr std::string_view PROP_AUDIO_LANG_ORIG = "inputstream.adaptive.original_audio_language";
 constexpr std::string_view PROP_PLAY_TIMESHIFT_BUFFER = "inputstream.adaptive.play_timeshift_buffer";
 constexpr std::string_view PROP_LIVE_DELAY = "inputstream.adaptive.live_delay"; //! @todo: deprecated to be removed on Kodi 23
-constexpr std::string_view PROP_PRE_INIT_DATA = "inputstream.adaptive.pre_init_data"; //! @todo: to be deprecated
+constexpr std::string_view PROP_PRE_INIT_DATA = "inputstream.adaptive.pre_init_data"; //! @todo: deprecated to be removed on Kodi 23
 
 constexpr std::string_view PROP_CONFIG = "inputstream.adaptive.config";
 constexpr std::string_view PROP_DRM = "inputstream.adaptive.drm";
@@ -280,9 +281,23 @@ void ADP::KODI_PROPS::CCompKodiProps::InitStage1(const std::map<std::string, std
     else
     {
       auto first = m_drmConfigs.begin();
-      first->second.license.serverUrl = licenseUrl;
+      first->second.license.serverUri = licenseUrl;
     }
   }
+}
+
+bool ADP::KODI_PROPS::CCompKodiProps::HasDrmConfig(std::string_view keySystem) const
+{
+  return STRING::KeyExists(m_drmConfigs, keySystem);
+}
+
+const ADP::KODI_PROPS::DrmCfg ADP::KODI_PROPS::CCompKodiProps::GetDrmConfig(
+    std::string_view keySystem) const
+{
+  if (STRING::KeyExists(m_drmConfigs, keySystem))
+    return m_drmConfigs.at(keySystem.data());
+
+  return {}; // default values
 }
 
 void ADP::KODI_PROPS::CCompKodiProps::ParseConfig(const std::string& data)
@@ -313,6 +328,38 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseConfig(const std::string& data)
     else if (configName == "internal_cookies" && jDictVal.IsBool())
     {
       m_config.internalCookies = jDictVal.GetBool();
+    }
+    else if (configName == "check_hdcp" && jDictVal.IsString())
+    {
+      std::string_view value = jDictVal.GetString();
+
+      if (value.empty() || value == "default")
+        m_config.hdcpCheck = HdcpCheckType::DEFAULT;
+      else if (value == "license")
+        m_config.hdcpCheck = HdcpCheckType::LICENSE;
+      else
+        LOG::LogF(LOGERROR, "Value \"%s\" isnt supported on \"%s\" config of \"%s\" property",
+                  value.data(), configName.c_str(), PROP_MANIFEST_CONFIG.data());
+    }
+    else if (configName == "resolution_limit" && jDictVal.IsString())
+    {
+      std::string_view value = jDictVal.GetString();
+      if (!value.empty())
+      {
+        auto pos = value.find('x');
+        if (pos != std::string_view::npos)
+        {
+          const int width = STRING::ToInt32(value.substr(0, pos));
+          const int height = STRING::ToInt32(value.substr(pos + 1));
+          m_config.resolutionLimit = width * height;
+        }
+        else
+        {
+          LOG::LogF(LOGERROR,
+                    "Invalid resolution format \"%s\" on \"%s\" config of \"%s\" property",
+                    value.data(), configName.c_str(), PROP_MANIFEST_CONFIG.data());
+        }
+      }
     }
     else
     {
@@ -379,9 +426,7 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseDrmOldProps(
 
   if (!STRING::KeyExists(props, PROP_LICENSE_TYPE))
     return;
-  /*
-   *! @todo: TO UNCOMMENT WHEN DRM AUTO-SELECTION WILL BE FULL IMPLEMENTED
-   *
+
   LOG::Log(LOGWARNING, "<<<<<<<<< DEPRECATION NOTICE >>>>>>>>>\n"
                        "DEPRECATED PROPERTIES HAS BEEN USED TO SET THE DRM CONFIGURATION.\n"
                        "THE FOLLOWING PROPERTIES WILL BE REMOVED FROM FUTURE KODI VERSIONS:\n"
@@ -392,11 +437,11 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseDrmOldProps(
                        "- inputstream.adaptive.server_certificate\n"
                        "- inputstream.adaptive.pre_init_data\n"
                        "YOU SHOULD CONSIDER MIGRATING TO THE NEW PROPERTIES:\n"
-                       "- inputstream.adaptive.drm\n"
                        "- inputstream.adaptive.drm_legacy\n"
+                       "- inputstream.adaptive.drm\n"
                        "FOR MORE INFO, PLEASE READ THE WIKI PAGE: "
                        "https://github.com/xbmc/inputstream.adaptive/wiki/Integration-DRM");
-  */
+
   std::string drmKeySystem{DRM::KS_NONE};
   if (STRING::KeyExists(props, PROP_LICENSE_TYPE))
     drmKeySystem = props.at(PROP_LICENSE_TYPE.data());
@@ -426,6 +471,8 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseDrmOldProps(
 
   // As legacy behaviour its expected to force the unique drm configuration available
   drmCfg.priority = 1;
+  // As legacy behaviour force single session (as in the old ISA versions <= v21)
+  drmCfg.isForceSingleSession = true;
 
   // Parse DRM properties
   const bool isRedacted = !CSrvBroker::GetSettings().IsDebugVerbose();
@@ -485,7 +532,7 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseDrmOldProps(
     {
       // Field 1: License server url
       if (fieldCount >= 1)
-        drmCfg.license.serverUrl = fields[0];
+        drmCfg.license.serverUri = fields[0];
 
       // Field 2: HTTP request headers
       if (fieldCount >= 2)
@@ -567,8 +614,8 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseDrmOldProps(
             // Position 2: The dict Key name to get HDCP value (optional)
             if (jPaths.size() >= 2)
             {
-              drmCfg.license.unwrapperParams["path_hdcp_traverse"] = "true";
-              drmCfg.license.unwrapperParams["path_hdcp"] = jPaths[1];
+              drmCfg.license.unwrapperParams["path_hdcp_res_traverse"] = "true";
+              drmCfg.license.unwrapperParams["path_hdcp_res"] = jPaths[1];
             }
           }
         }
@@ -608,7 +655,7 @@ bool ADP::KODI_PROPS::CCompKodiProps::ParseDrmConfig(const std::string& data)
       continue;
     }
 
-    DrmCfg& drmCfg = m_drmConfigs[keySystem];
+    DrmCfg& drmCfg = m_drmConfigs[keySystem]; // create new configuration
     auto& jDictVal = jChildObj.value;
 
     if (!jDictVal.IsObject())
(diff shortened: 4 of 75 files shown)


CastagnaIT:
@glennguy are you still here? please let me know
i need help about CBCS, is broken from Kodi 21 and i have no clear idea

for now it would be enough to have it working by using this PR (Kodi 22)
have you time to check the problem?

by using "InputStream adaptive samples" addon, im pretty sure that following stream was working:
`List CBCS streams` menu --> `Google Tears [CBCS]`

EDIT: fixed by #1790

Uukrull:
I did a quick test and the only regression I noticed is that UrnToKeySystem returns "" if       
      `<ContentProtection schemeIdUri="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed">`
doesn't exist, but `<ContentProtection schemeIdUri="urn:mpeg:dash:mp4protection:2011" value="cenc" cenc:default_KID="xxxxxxxxx" />`
exists so it should support those cases, otherwise it won't work because keySystem is empty.

CastagnaIT:
in this case the target DRM is unknown because "urn:mpeg:dash:mp4protection:2011" means only CENC protection that should be compatible with all drm's
i think the better thing could be add an additional check that when there is a single DRMInfo with no KS match to the first DRM (in priority order), otherwise another idea is check the initial segment, but this require much more changes, maybe later time

CastagnaIT:
@Uukrull fixup added let me know

Uukrull:
Working fine here. Many thanks!

CastagnaIT:
if no other problems arise
i think that next week i will merge and make a new release

CC @matthuisman if you want take a look and/or test about the changes

CC @Sandmann79 am@zon addon use JSON license,
since i have made many changes, i recalled that i recently saw am@zon logs highlighting problems on JSON HDCP value,
i think you need to fix or remove the HDCP configuration
although it is set up here:
https://github.com/Sandmann79/xbmc/blob/0f2c30a8c02bdea32e151916e51990455fa547fd/plugin.video.amazon-test/resources/lib/playback.py#L480
i suspect that is not used properly by ISA, since in order to work the MPD manifest must also provide the `hdcp` attribute to each `Representation` tag, i cannot verify since i havent an account,
anyway i add a new wiki page about this: https://github.com/xbmc/inputstream.adaptive/wiki/HDCP-resolution-limit

matthuisman:
@CastagnaIT  this sounds really good

I will try to test out various addons of mine with this new version tonight and report back

UPDATE:
Tried a few addons and everything worked as expected :)

CastagnaIT:
good thank you

Sandmann79:
@CastagnaIT you're right, the HDCP key is probably no longer needed. Especially as there are no HDCP versions in representation tags in the manifest.

However, in my short tests with the new DRM implementation I could not play any video, even with the deprecated method and ISA >= 22.1.5. I will create an issue for this in the next few days.

CastagnaIT:
@Sandmann79 yesterday afternoon i have updated commit of this PR
have you tested with this updated build?
just to know because i have add a fix on dash

CastagnaIT:
@matthuisman I recalled this old problem on android 32bit
from issue #1147 should be about "Kayo"
Do you have a chance to check if the problem still occurs with this PR?

matthuisman:
was that problem fixed with https://github.com/xbmc/inputstream.adaptive/pull/1165 ?
Kayo no longer has a Kodi addon to test with unfortunately.

CastagnaIT:
ah ok then nothing,
yes has been fixed, but now the license is requested by stream type (audio/video) in a separate way
should not cause any problems because it is how sessions are generally recommended to be handled
but there may be some special video service case that complain about this
will be addressed when the time comes...

CastagnaIT:
@Sandmann79 i would merge this tomorrow or sunday
Do you have time to take a test?

Sandmann79:
I tested it with the latest commits and unfortunately still get the following error on Windows OS:
` error <general>: AddOnLog: inputstream.adaptive: CWVCencSingleSampleDecrypter::CWVCencSingleSampleDecrypter: Cannot perform License update, no session available`
On Android, playback works fine.

Edit - complete Log: [kodi.log](https://github.com/user-attachments/files/19516196/kodi.log)

CastagnaIT:
i found the problem
```
      <ContentProtection schemeIdUri="urn:mpeg:dash:mp4protection:2011" value="cenc" cenc:default_KID="94D9AACF-783B-4AAB-8340-8AF4679E6606"/>
      <ContentProtection schemeIdUri="urn:uuid:9A04F079-9840-4286-AB92-E65BE0885F95" value="MSPR 2.0">
        <cenc:pssh>bAIAAAEAAQBiAjwAVwBSAE0ASABFAEEARABFAFIAIAB4AG0AbABuAHMAPQAiAGgAdAB0AHAAOgAvAC8AcwBjAGgAZQBtAGEAcwAuAG0AaQBjAHIAbwBzAG8AZgB0AC4AYwBvAG0ALwBEAFIATQAvADIAMAAwADcALwAwADMALwBQAGwAYQB5AFIAZQBhAGQAeQBIAGUAYQBkAGUAcgAiACAAdgBlAHIAcwBpAG8AbgA9ACIANAAuADAALgAwAC4AMAAiAD4APABEAEEAVABBAD4APABQAFIATwBUAEUAQwBUAEkATgBGAE8APgA8AEsARQBZAEwARQBOAD4AMQA2ADwALwBLAEUAWQBMAEUATgA+ADwAQQBMAEcASQBEAD4AQQBFAFMAQwBUAFIAPAAvAEEATABHAEkARAA+ADwALwBQAFIATwBUAEUAQwBUAEkATgBGAE8APgA8AEsASQBEAD4AegA2AHIAWgBsAEQAdAA0AHEAMABxAEQAUQBJAHIAMABaADUANQBtAEIAZwA9AD0APAAvAEsASQBEAD4APABDAEgARQBDAEsAUwBVAE0APgBtAFcARwBBAFMAVgB1AHAAMAA2AHcAPQA8AC8AQwBIAEUAQwBLAFMAVQBNAD4APABMAEEAXwBVAFIATAA+AGgAdAB0AHAAcwA6AC8ALwBwAHIAbABzAC4AYQB0AHYALQBwAHMALgBhAG0AYQB6AG8AbgAuAGMAbwBtAC8AYwBkAHAAPAAvAEwAQQBfAFUAUgBMAD4APAAvAEQAQQBUAEEAPgA8AC8AVwBSAE0ASABFAEEARABFAFIAPgA=</cenc:pssh>
      </ContentProtection>
      <ContentProtection schemeIdUri="urn:uuid:EDEF8BA9-79D6-4ACE-A3C8-27DCD51D21ED">
        <cenc:pssh>CAESEJTZqs94O0qrg0CK9GeeZgYaBmFtYXpvbiI1Y2lkOjU4alZySENPUWZXOGZwbDFxM0pHRnc9PSxsTm1xejNnN1NxdURRSXIwWjU1bUJnPT0qAlNEMgA=</cenc:pssh>
      </ContentProtection>
```
on am@zon manifest there are the pssh that are not in standard CENC PSSH format
they directly set the pssh data (that should be contained in a PSSH box field)
fwik doing this, is out-of-spec
that is the reason why on Android works and not on Windows

to fix this i need to add an additional check in order to manually construct the missing PSSH box

CastagnaIT:
@Sandmann79 fix added if you can confirm thanks

Sandmann79:
Thanks, the initial error has now disappeared. 
But now it takes more than one try to start the video and rewinding during playback causes Kodi to hang.
For comparison, I just tested with ISA 22.1.4 and there are no such errors.

[kodi.log](https://github.com/user-attachments/files/19521189/kodi.log)

CastagnaIT:
i see closed audio session, suggest that DecrypterCapabilites::SSD_INVALID is set to audio stream that make fails
not full sure since i cant test, i tried remove it when you have chance let me know

Sandmann79:
Now the streams start at the first attempt, but some need more time to start (video start at lines: 1171 and 5107).
Seeking is still not possible, see last video in the log.

[kodi.log](https://github.com/user-attachments/files/19521555/kodi.log)

CastagnaIT:
code changed hoping works

Sandmann79:
The streams now starts without delay. However, seeking still causes Kodi to crash.

[kodi.log](https://github.com/user-attachments/files/19524765/kodi.log)

CastagnaIT:
can you provide one log with last working isa version?
because otherwise im not able to understand the log differences
with same behaviour start playback and after video seek

edit:
i suspect the problem is due to new drm multisessions (per a/v mediatype)
I may have forgotten a piece as explained here
https://players.castlabs.com/android/latest/docs/build/html/drm.html#multi-key-playbacks
i need to check also if there are same KID on a/v, this should solve the problem

CastagnaIT:
@Sandmann79  ok pushed changes, this should work something like Omega

Sandmann79:
Unfortunately, seeking is still not working.
After seeking, there is an infinite buffering, the playback time continues, but the video and sound freeze. As soon as playback is stopped, Kodi hangs up.
The logs are attached for comparison.

[kodi-ISA_22_1_4.log](https://github.com/user-attachments/files/19526506/kodi-ISA_22_1_4.log)
[kodi-ISA_22_1_13.log](https://github.com/user-attachments/files/19526507/kodi-ISA_22_1_13.log)

CastagnaIT:
isa has opened again two drm sessions due to different kids,
i think i need to add a new DRM config to force the single session, such as the link i posted above

CastagnaIT:
@Sandmann79  i have done some changes in hope i have right things
this time for test i have hardcoded the single drm session
if works, please provide a new log for Windows, and try test if works also on android

Sandmann79:
On Windows it is the same behavior as before and on Android there are no problems with seeking.

[kodi-Win.log](https://github.com/user-attachments/files/19527380/kodi-Win.log)
[kodi-A14.log](https://github.com/user-attachments/files/19527381/kodi-A14.log)

CastagnaIT:
its weird i will need to get an am@zon account to try understand better

Uukrull:
You can get Amazon Prime for less than 5€ per year on the egyptian Amazon (amazon.eg) You only need an address and phone number from Egypt and you can use your existing Amazon account if you have it.
The only thing I noticed when using it from Europe if that with the official app the message "It seems you are travelling" appears from time to time. Other than that, no problems at all.

CastagnaIT:
i tried looking at my account today and strangely i still have 30 days free trial, i remember that my trial was expired

CastagnaIT:
i found two problems,
one problem has been introduced from 22.1.5 with the new properties changes that add pointer reset to CVideoCodecAdaptive::Reset https://github.com/xbmc/inputstream.adaptive/blob/a1a6ff339f6e20b5df9a747798f7e9d65a5c6baa/src/main.cpp#L628
I didn't expect this problem, at that time with n@tflix it worked, with video seek, reset occurs and broken playback
the second problem has been introduced here due to oversight
i'll have to figure out how to solve

CastagnaIT:
@Sandmann79 i have updated code, but i tested only on windows, can you give a feedback also on android?

i added new `DRM Sessions (breaking change)` section on PR description
that explain more or less about the drm sessions changes, anyway am@zon should be compatible with both new/old session methods
anyway i suggest you to try add `force_single_session` parameter to `True` in your addon when you use `inputstream.adaptive.drm`

PS -- OT:
there is somewhere am@zon bug, but dont happens always,
it happens (on windows in my tests) to reproduce usually i do -> play a video -> close kodi with "X" on window frame
happens that kodi try to close itself but it hang when it try to terminate amazon service, it dont free resources and kodi executable become a "zombie" (you need to kill process)
when this problem happens you can see "[Amazon VOD] Service stopped" on log, but its missing "[Amazon VOD] Service: Proxy server stopped"
the python code is freezed on one of these calls
https://github.com/Sandmann79/xbmc/blob/12c1771994c2294d661e8a227fe25eee0bbc46c2/plugin.video.amazon-test/resources/lib/service.py#L56-L59

EDIT: perfect on android is full broken

Sandmann79:
Sorry for the late reply, can you please check the Windows OS again, I have the problem again that several attempts are necessary to start a stream. Even though force_single_
session is set. In the attached log it only works on the 3rd attempt.

Thanks for the tip about the service not terminating. I'll have to take a look at that.

[kodi.log](https://github.com/user-attachments/files/19592611/kodi.log)

CastagnaIT:
@Sandmann79 hi i was trying investigate in these hours
and i found 2 bugs, one on inputstream.adaptive.drm that i have just pushed fixup,
and one on your addon, you are sending wrong drm configuration
at least was wrong in the addon version that i have locally, please check your sources
this is the fixed playback.py:
```py
        req_data = "widevine2Challenge={CHA-B64U}&includeHdcpTestKeyInLicense=true"
        drm_cfg = {'com.widevine.alpha':
                       {'license':
                            {'server_url': licURL,
                             'req_headers': urlencode(lic_headers),
                             'req_data': base64.b64encode(req_data.encode('utf-8')).decode(),
                             'wrapper': "none",
                             'unwrapper': 'json,base64',
                             'unwrapper_params': {'path_data': 'license', 'path_data_traverse': True,
                                                  'path_hdcp': 'hdcpEnforcementResolutionPixels', 'path_hdcp_traverse': True}
                             }
                        }
                   }
        listitem.setProperty('inputstream.adaptive.drm', json.dumps(drm_cfg))
```
with these 2 fixes, playback on windows works with old and new drm properties
the playback problem on Android i will have to investigate better another day...

Sandmann79:
I am currently using the following DRM config, with which playback was possible:
https://github.com/Sandmann79/xbmc/blob/master/plugin.video.amazon-test/resources/lib/playback.py#L418-L428

However, the above-mentioned effect also occurs when using the old properties.

CastagnaIT:
ok discovered now that some videos works and others not
this thing is becoming a nightmare i will investigate on next days

CastagnaIT:
Okay, i think i found the problem affect windows,
this was difficult to understand since the mechanism was not so directly understandable

in short am@zon have some of his audio streams that require CDM "secure path"

since on ISA its missing "secure path" implementation for audio media
will be trigged SSD_INVALID, **silently**
https://github.com/xbmc/inputstream.adaptive/blob/2364435138d600edb3c192a4fa8e31a27f714620/src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp#L202-L207
**and silently also remove all audio streams!!**
https://github.com/xbmc/inputstream.adaptive/blob/2364435138d600edb3c192a4fa8e31a27f714620/src/Session.cpp#L556-L558
https://github.com/xbmc/inputstream.adaptive/blob/2364435138d600edb3c192a4fa8e31a27f714620/src/common/Period.cpp#L69-L83

over the years i have forgotten about this oddity in the removal of repr, also because i didn't know where to get a suitable test stream
and this print nothing on log, so its not possible check from log, this is another bad hack shit from the old code

this lead to BIG PROBLEMS:
1) i have no idea how much efforts are required to implement CDM "secure path" for audio side, because if you try use it ffmpeg complains and throws you out...
2) this lead to difficult in to skip these audio streams that require  CDM "secure path"
3) nullifies all good improvements made here about speedup playback startup and avoid opening unnecessary CDM sessions for a not real use

audio "secure path" is old feature lack in ISA forgotten over all the years,
a kind of solution can be check if there are DRM encrypted audio streams, if so, initialize all streams to CDM in similar way of the old ISA behavior

CastagnaIT:
@Sandmann79 ok i pushed changes, need some cleaning but should be ok

unfortunately this lack in a single decryptor (widevine CDM for non-android) affects the whole DRM management, i had to reintroduce the old hack in a different way (of course that messes up a bit the code again) but at least now its more clear and print on log the disabled (unsupported) audio streams

I tested a couple of videos on windows and android and now it seems to be working fine with old and new props on windows/android

please do some tests so you can give me a feedback

i confirm also that add `force_single_session` parameter to `True` on Android will allow you to keep a single drm session (as on windows) anyway by using old props its enabled by default

CastagnaIT:
@Sandmann79 i pushed last cleanups, tested again seem ok
when you will have time let me know

Sandmann79:
I've tested several videos on different systems and couldn't find any issues. 

Thanks for the really good work and your patience.

CastagnaIT:
Thank you i finally got to the end of this colossal PR

CastagnaIT:
hmf too soon i found now that at least on android there are still problems with some videos
atm its unclear whats wrong

CastagnaIT:
ok, i found that enable 4K on am@zon addon cause decryption problems, whats weird thing
i disable it and now works perfectly as before

mkreisl:
Hi, thanks for the tireless work here.

Could it be that with this PR the previous Python addons no longer work? In my specific case it is the simpliTV addon
On my Trixie test installation, a Kodi 22 alpha built 2 days ago with all relevant binary addons no longer works
Kodi no longer even shows an error message when starting a stream but simply does nothing except write this Depreciated warning in the log

With a slightly older Kodi 22 alpha version with correspondingly older binary addons, however, streaming works perfectly, so it must be due to some new commits

By the way, the version of libwidevine.so is 4.10.2662.3

mirh:
Can you try #1824?

mkreisl:
@mirh 
I will try it tomorrow and report back. I'm too tired to do it today.
I built the 22.1.13 version again to be sure and tested it with the Kodi 22 Alpha from today, that works anyway

mkreisl:
@mirh 
Now I've done it after all.
Unfortunately it doesn't help, the stream doesn't start again, but I had the feeling that it was doing a bit more
I looked in the log and found this:
```
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.401 T:30470   debug <general>: AddOnLog: inputstream.adaptive: Download finished: https://simplitv-live.mdn.ors.at/live/eds/servustv_fhd-1/dash/servustv_fhd-1.mpd (downloaded 16368 byte, speed 765611.00 byte/s)
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470    info <general>: AddOnLog: inputstream.adaptive: Manifest successfully parsed (Periods: 1, Streams in first period: 4, Type: live)
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470   debug <general>: AddOnLog: inputstream.adaptive: [Repr. chooser] Stream selection conditions
                                                   Screen resolution: 1920x1080 (may be limited by settings)
                                                   Initial bandwidth: 196188005 bit/s
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470   debug <general>: AddOnLog: inputstream.adaptive: Initialize crypto session
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470   error <general>: AddOnLog: inputstream.adaptive: Cannot find the cdm_aarch64_loader.so file
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470   error <general>: AddOnLog: inputstream.adaptive: The DRM decrypter cannot be initialized
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470   debug <general>: AddOnLog: inputstream.adaptive: GetCapabilities()
Apr 12 22:47:37 kmpifive 2025-04-12 22:47:37.407 T:30470    info <general>: Creating Demuxer
```
but the libcdm_aarch64_loader.so file is present

```
/usr/local/lib/kodi/addons/inputstream.adaptive $ ll
insgesamt 2208
lrwxrwxrwx 1 root root      28 12. Apr 22:44 inputstream.adaptive.so -> inputstream.adaptive.so.22.0
lrwxrwxrwx 1 root root      30 12. Apr 22:44 inputstream.adaptive.so.22.0 -> inputstream.adaptive.so.22.2.0
-rw-r--r-- 1 root root 2180176 12. Apr 22:44 inputstream.adaptive.so.22.2.0
-rw-r--r-- 1 root root   67192 12. Apr 22:44 libcdm_aarch64_loader.so
```
hth

CastagnaIT:
there was wrong filename "cdm_aarch64_loader.so"
fix on #1825

mkreisl:
Yeah, great  :smile: Works again
Thanks

PR #1746, opened 2024-12-30
Add a simple thread pool

neo1973:
## Description

Implement a simple thread pool so threads can be reused instead of recreated.

## Motivation and context

inputstream.adaptive creates a lot of threads though `std::async` or by instantiating `std::thread` directly. E.g. watching this stream I measured about 90 threads being created and destroyed per second (very inaccurate measure using a stopwatch and GDB debug output):
```
#KODIPROP:mimetype=application/dash+xml
#A new period every 3 min (20 times/hour)
#no segment updates needs inserting
```
Creating threads is a non trivial task and reusing threads instead of recreating them improves efficiency.

## How has this been tested?

Tested with the above stream and https://github.com/anxdpanic/plugin.video.youtube/. I confirmed in GDB on Linux that the thread pool grows to a size of only two and then no new threads are created.

## Types of change

- [X] **Improvement** (non-breaking change which improves existing functionality)

## Checklist:

### DIFF
--- a/src/utils/ThreadPool.cpp
+++ b/src/utils/ThreadPool.cpp
@@ -1,0 +1,72 @@
+/*
+ *  Copyright (C) 2025 Team Kodi
+ *  This file is part of Kodi - https://kodi.tv
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *  See LICENSES/README.md for more information.
+ */
+
+#include "ThreadPool.h"
+
+UTILS::THREAD::ThreadPool::~ThreadPool()
+{
+  Stop();
+}
+
+void UTILS::THREAD::ThreadPool::Stop()
+{
+  {
+    std::lock_guard lock(m_mutex);
+    m_isStopped = true;
+  }
+
+  m_condVar.notify_all();
+
+  for (const auto& executor : m_executors)
+    executor->Join();
+}
+
+std::optional<std::function<void()>> UTILS::THREAD::ThreadPool::TakeTask()
+{
+  std::unique_lock lock(m_mutex);
+
+  m_condVar.wait(lock, [this]() { return !m_taskQueue.empty() || m_isStopped; });
+
+  if (m_isStopped)
+    return {};
+
+  ++m_activeExecutors;
+  auto func = std::move(m_taskQueue.front());
+  m_taskQueue.pop();
+  return {std::move(func)};
+}
+
+void UTILS::THREAD::ThreadPool::TaskFinished()
+{
+  std::lock_guard lock(m_mutex);
+  --m_activeExecutors;
+}
+
+UTILS::THREAD::ThreadPool::Executor::Executor(ThreadPool& threadPool) : m_threadPool(&threadPool)
+{
+  m_thread = std::thread(&Executor::Run, this);
+}
+
+void UTILS::THREAD::ThreadPool::Executor::Join()
+{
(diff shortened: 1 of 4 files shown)


CastagnaIT:
interesting alternative solution, although the ReadSampleAsync "functionality" should be completely rewritten in a future can in the meantime be an improvement
i will take a look on next days

src/common/AdaptiveTree.cpp
```diff
@@ -260,16 +260,16 @@ namespace adaptive
...
   }
 
   void AdaptiveTree::TreeUpdateThread::Initialize(AdaptiveTree* tree)
   {
-    if (!m_thread.joinable())
+    if (!m_future.valid())
     {
       m_tree = tree;
-      m_thread = std::thread(&TreeUpdateThread::Worker, this);
+      m_future = GlobalThreadPool.Execute(&TreeUpdateThread::Worker, this);
```
CastagnaIT: please avoid do these changes to the update thread, this is a single thread not multiple, so it start on playback starts and nothing more

another reason is that the update thread worker must be in the future redone from scratch because currently implementation is not suitable for handling all kind of manifest updates, so i prefer keep its complexity lower to make easier future reworks

src/utils/ThreadPool.cpp
```diff
@@ -0,0 +1,64 @@
+#include "ThreadPool.h"
```
CastagnaIT, line 9: to add copyright/license on top
```
/*
 *  Copyright (C) 2025 Team Kodi
 *  This file is part of Kodi - https://kodi.tv
 *
 *  SPDX-License-Identifier: GPL-2.0-or-later
 *  See LICENSES/README.md for more information.
 */
```

src/utils/ThreadPool.h
```diff
@@ -0,0 +1,104 @@
+#include <condition_variable>
```
CastagnaIT, line 9: to add copyright/license on top
```
/*
 *  Copyright (C) 2025 Team Kodi
 *  This file is part of Kodi - https://kodi.tv
 *
 *  SPDX-License-Identifier: GPL-2.0-or-later
 *  See LICENSES/README.md for more information.
 */
```

src/utils/ThreadPool.h
```diff
@@ -0,0 +1,104 @@
...
+#include <functional>
+#include <future>
+#include <mutex>
+#include <optional>
+#include <queue>
+#include <thread>
+#include <vector>
+
+#pragma once
+
```
CastagnaIT, line 20: please add appropriate namespaces such as the others utils files
at least the "utils"
```
namespace UTILS
{
```

src/test/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ add_executable(${BINARY}
     ../utils/FileUtils.cpp
     ../utils/JsonUtils.cpp
     ../utils/StringUtils.cpp
+    ../utils/ThreadPool.cpp
```
CastagnaIT: when removed from AdaptiveTree (my previous comment) this addition should not be needed

src/common/AdaptiveTree.cpp
neo1973: Indeed. No clue why I identified this a source of (many) spawning threads.

CastagnaIT (approved):
tested on windows and android
not found problems

Issue #1748, opened 2024-12-31
[Bug] Kodi 21.1 crashes when playing videos using inputstream-adaptive (21.5.7)


anyc:
### Describe the problem

Hello,

I updated from a working Kodi 20 to 21.1 on my custom built system on a RPi4. With Omega, Kodi always crashes at the same place inside inputstream-adaptive (21.5.7) when playing videos from Netflix (1.23.3), Prime (1.1.4) or Youtube. Playing an IPTV stream works.

I had to ignore the update request from inputstreamhelper as my glibc is older and does not provide the required symbols for the new widevine lib.

As this is a custom built system, I might be doing something wrong - maybe I missed a minimum version for a dependency or something. Any hint is appreciated.

Thank you!

### Steps to reproduce

1, start Kodi
2. play a video from Netflix, Prime or Youtube

### Debug log

Crash log (with not much info about the crash imho): https://paste.kodi.tv/ijinoraqup.kodi

GDB backtrace:
```
(gdb) back
#0  __libc_do_syscall () at ../sysdeps/unix/sysv/linux/arm/libc-do-syscall.S:47
#1  0xb320e2a0 in __libc_signal_restore_set (set=0x92bfa854) at ../sysdeps/unix/sysv/linux/internal-signals.h:84
#2  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:48
#3  0xb32003f2 in __GI_abort () at abort.c:79
#4  0xb32369ac in __libc_message (action=action@entry=do_abort, fmt=<optimized out>) at ../sysdeps/posix/libc_fatal.c:181
#5  0xb323ba16 in malloc_printerr (str=<optimized out>) at malloc.c:5449
#6  0xb323bd6e in munmap_chunk (p=<optimized out>) at malloc.c:2863
#7  0xb3446c42 in operator delete(void*) () from /lib32/armv7a-hardfloat-linux-gnueabi/8/libstdc++.so.6
#8  0xac5281c0 in std::__detail::_Compiler<std::regex_traits<char> >::_Compiler(char const*, char const*, std::locale const&, std::regex_constants::syntax_option_type) () from /lib32/kodi/addons/inputstream.adaptive/inputstream.adaptive.so.21.5.7
#9  0xac51e620 in UTILS::XML::ParseDuration(std::basic_string_view<char, std::char_traits<char> >) ()
[...]

PR #1750, opened 2025-01-02
Fix incorrect `std::string_view` usage

neo1973:
## Description

In many places it was assumed that `std::string_view::data()` would return a null terminated string that only contains the slice the view is pointing at. This is in general not true.

## Motivation and context

I became aware of this general issue when investigating #1748. This PR probably solved that issue but I'm not certain.

## How has this been tested?

Test suite runs, YouTube works.

## Types of change

## Checklist:

### DIFF
--- a/src/CompKodiProps.cpp
+++ b/src/CompKodiProps.cpp
@@ -69,8 +69,10 @@ constexpr std::string_view PROP_CHOOSER_RES_SECURE_MAX = "inputstream.adaptive.c
 
 void LogProp(std::string_view name, std::string_view value, bool isValueRedacted = false)
 {
-  LOG::Log(LOGDEBUG, "Property found \"%s\" value: %s", name.data(),
-           isValueRedacted ? "[redacted]" : value.data());
+  if (isValueRedacted)
+    value = "[redacted]";
+  LOG::Log(LOGDEBUG, "Property found \"%.*s\" value: %.*s", static_cast<int>(name.length()),
+           name.data(), static_cast<int>(value.length()), value.data());
 }
 
 void LogDrmJsonDictKeys(std::string_view keyName,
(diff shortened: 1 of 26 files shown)


CastagnaIT:
in many places the use of string_view is intentional
so i dont agree in full this pr

all logs changes are confusing due to additional length input
```
LOG::LogF(LOGWARNING, "Skipped AdaptationSet with id: \"%.*s\", has no representations.",
              static_cast<int>(adpSet->GetId().length()), adpSet->GetId().data());
```
fwik all strings manged on ISA are null terminated, if it is missing there is to understand why

neo1973:
> all logs changes are confusing due to additional length input

Probably time to migrate to [std::format](https://en.cppreference.com/w/cpp/utility/format)?

> fwik all strings manged on ISA are null terminated, if it is missing there is to understand why

This might very well be true, but someone who doesn't know the ins and outs of inputstream.adaptive has to check every time a function accepts a `std::string_view` if it in reality expects a c-style string.

See e.g.:
https://github.com/xbmc/inputstream.adaptive/blob/d789f974a8c14c58877517971f6b59319fac7a22/src/parser/HLSTree.cpp#L93
This has been special cased in StringUtils:
https://github.com/xbmc/inputstream.adaptive/blob/d789f974a8c14c58877517971f6b59319fac7a22/src/utils/StringUtils.cpp#L182-L187
Do the same for `UTILS::STRING::ToUint32` and it won't work. And that's the main problem. You can't look at a function and be certain that is does the correct thing even if you use it according to its contract.

CastagnaIT:
PR blocked but i keep open to take a future look
to see if it is worth replacing some of the proposed string_views

CastagnaIT:
closed as superseesed by #1816 
in the end i chose some of your commits others i reworked in a different way

PR #1751, opened 2025-01-02
Migrate logging to `std::format`

neo1973:
## Description

Simple conversion of logging to use [`std::format`](https://en.cppreference.com/w/cpp/utility/format).

## Motivation and context

Primarily to have proper support to print `std::string_view` (see #1750). And also in general the syntax is nicer and compile time format string checks can catch certain errors.

## How has this been tested?

Compiled and tested on x86_64 Linux. YouTube live streams and regular videos work.

## Types of change

- [X] **Improvement** (non-breaking change which improves existing functionality)

## Checklist:

### DIFF
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -240,8 +240,7 @@ void CdmAdapter::Initialize()
 
   if (!library_)
   {
-    LOG::LogF(LOGERROR, "%s: Failed to load library: %s", __FUNCTION__,
-             error.ToString().c_str());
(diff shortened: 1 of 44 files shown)


neo1973:
Old toolchains are a PITA... I'll try to pull in fmtlib until all platforms catch up and support `std::format`.

CastagnaIT:
since this could be an interesting solution, the use of `std::format_to_n` causes the formatting to be applied permanently
and so there is a double pass `format_to_n` and after `kodi::Log` call

while the second can be disabled from kodi debug settings,
the format_to_n will be always called in ISA even when it is not needed by resulting an extra computation

otherwise an additional setting should be added to exclude it, but i don't think it's worth complicating things further for users
just to have a more elegant code formatting

maybe a better solution should be found through the InputStream interface
I'm closing for now

PR #1771, opened 2025-01-30
[DashTree] Changes to parsing values from "Role" tag

### LINKED ISSUE #1768: [Feature request] DASH support froced-subtitle role
### Describe the feature

It will be nice if we add support for "forced-subtitle" role as described in de DASH standard and not only for "forced" role because it is ISA custom value as we can see in the source code.

### Motivation

Having this implemented this we can have DASH parser that respect de standard an we can avoid rewriting DASH manifests in proxys.

No

Yes

CastagnaIT:
## Description

Changes on "Role" tag value parsing:

### Disabled "subtitle", "caption" uses

"subtitle" looks like a kind of old workaround to set content type,
but "should" be used only to signal burned-in subtitles to video stream (not CC 608 etc..)

"caption" instead has been used to set impaired flag, imo its not correct thing
its not a rule that "caption" stand for impaired, as you can read also
on wikipedia this should not be a rule, but only some countries may use it as rule

IOP/ISO EIC specs say that "subtitle" and "caption" have same meaning of burned-in subtitles
so lets try to remove both and we will see if someone complains

(compared to the past the current source code have more robust code to identify the “content type” so it should be low risk of breaking some service by removing "subtitle" case value)

### Deprecated custom "forced" support

i have implemented "forced" value years ago, i was not aware of "forced-subtitle" official specs value,
so lets deprecate it, for future removal

## Motivation and context

fix #1768 
## How has this been tested?

cant be test by my part

## Types of change

## Checklist:

### DIFF
--- a/src/parser/DASHTree.cpp
+++ b/src/parser/DASHTree.cpp
@@ -466,13 +466,14 @@ void adaptive::CDashTree::ParseTagAdaptationSet(pugi::xml_node nodeAdp, PLAYLIST
 
     if (schemeIdUri == "urn:mpeg:dash:role:2011")
     {
-      if (value == "subtitle")
-        contentType = "text";
-      else if (value == "forced") // ISA custom attribute
+      //! @todo: If no complains, remove commented lines on Kodi 23
+      // if ((value == "subtitle" || value == "caption") && contentType.empty())
+      //   contentType = "text";
+      if (value == "forced") // ISA custom attribute
         adpSet->SetIsForced(true);
       else if (value == "main")
         adpSet->SetIsDefault(true);
-      else if (value == "caption" || value == "alternate" || value == "commentary")
+      else if (value == "alternate" || value == "commentary")
         adpSet->SetIsImpaired(true);
     }
   }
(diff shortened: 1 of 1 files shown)


CastagnaIT:
@matthuisman have you some custom uses of "Role" tag values?
in case let me know if this code change may result in any headaches

matthuisman:
i think i can just add both forced and forced-subtitle to keep backwards compatbility?
And swap to alternate or commentary instead of caption

matthuisman:
actually, i think im OK

I do have some code to work-around IA not supporting this already:
image

Thats all i do in regards to "Role"
In the above - i just set forced on the adaption set if i see forced-subtitle role

CastagnaIT:
yes you are doing right things
then you dont need change nothing
for Kodi 22 if you want you can remove the setAttribute “forced” from Role since that case will be handled

PS. anyway "forced" attrib to AdaptationSet will be kept, is not the scope of this PR

matthuisman:
my code is meant to be backwards compatible way back to Kodi 17 hehehe.
So ill just leave it - it wont hurt anything
itll just have 2x indicators for forced

PR #1790, opened 2025-03-23
[FragmentedSampleReader] Fix transitions from unencrypted fragments to encrypted

glennguy:
## Description

UpdateSampleDescription() will update the m_protectedDesc member variable, but never clear it when the sample description box says there is no protection. Since ProcessMoof() makes a decision to process the fragment as decrypted or not based on this variable being assigned or not, we can end up decrypting unencrypted content. On windows, trying to 'play' unencrypted audio content can cause a crash.

This change will clear the m_protectedDesc member when the fragment is not protected, allowing other logic to function as intended

## Motivation and context

Fixes playback of the CBCS Google 'Tears of Steel' sample stream. This stream starts with the first fragment unencrypted. We however populate m_protectedDesc because the Initialize() method uses a magic number (0) get the first sample description, regardless of what sample description is specified for in the first fragment. This should be looked at in the future.

## How has this been tested?

Tested on Windows

## Types of change

## Checklist:

### DIFF
--- a/src/samplereader/FragmentedSampleReader.cpp
+++ b/src/samplereader/FragmentedSampleReader.cpp
@@ -499,6 +499,9 @@ void CFragmentedSampleReader::UpdateSampleDescription()
       return;
     }
   }
+  else {
+    m_protectedDesc = nullptr;
+  }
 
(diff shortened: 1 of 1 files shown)


CastagnaIT (approved):
thank you very much

Issue #1801, opened 2025-04-01
[Bug] widevine version 4.10.2891.0 isn´t working


goofy84:
### Describe the problem

widevine isn´t working for private channels waipu.tv addon

nightly-20250330 but also stable versions like 12.02

x64 generic leagcy nvidia rtx a2000

[kodi.log](https://github.com/user-attachments/files/19546508/kodi.log)
[kodi_crashlog_20250331144728.log](https://github.com/user-attachments/files/19546507/kodi_crashlog_20250331144728.log)

### Debug log

[kodi.log](https://github.com/user-attachments/files/19546518/kodi.log)
[kodi_crashlog_20250331144728.log](https://github.com/user-attachments/files/19546517/kodi_crashlog_20250331144728.log)

### Operating system(s)

Linux

### Operating system version(s)

libreelec 

### InputStream Adaptive version(s)

22.1.13.1

### Kodi version(s)

libreelec 13 nightly 20250330 and stable version 12.02

CastagnaIT:
> AddOnLog: inputstream.adaptive: Initialize: CDM version: 4.10.2891.0

used widevine version is problematic,
seems to have a bug in that it crashes in the call of `OnStorageId`
https://github.com/xbmc/inputstream.adaptive/blob/a1a6ff339f6e20b5df9a747798f7e9d65a5c6baa/lib/cdm/cdm/media/cdm/cdm_adapter.cc#L698

a workaround would be to avoid this call but it doesn't seem like a very appropriate solution

the best for now that you rollback to the previous widevine version: 4.10.2830.0
by using ISHelper addon

i suggest to wait what happens on next weeks, maybe a new release of the fixed library will be released

goofy84:
Thank you very much this solved the problem

goofy84:
Hello, 
can you help again? I updated the nightly version to 02.04.25 reinstalled the recommanded widevine version but once again i can not watch private channels. So i rolled back all versions inputstream adaptive, waiputv, imputstream helper ffmpeg tools ffmpeg imput stream direct
here is the log:
[https://paste.kodi.tv/agofidekoc](url)

CastagnaIT:
> can you help again? I updated the nightly version...

from log looks like you have networks problems

for others users that will read this:
on the new ISA releases i blocked loading of this problematic version just to avoid kodi crashes
ofc its always needed rollback to old widevine version
this until a solution will be found

williamcorney:
[...]

PR #1803, opened 2025-04-02
[cdm] Deny the use of CDM version 4.10.2891.0

CastagnaIT:
## Description

seems that CDM version 4.10.2891.0 have a bug it crashes in the call of `OnStorageId`
https://github.com/xbmc/inputstream.adaptive/blob/a1a6ff339f6e20b5df9a747798f7e9d65a5c6baa/lib/cdm/cdm/media/cdm/cdm_adapter.cc#L698

a workaround could be avoid call `OnStorageId` method,
but this works only with some DRM streams

if you try play streams that require the secure path decoding (e.g. Am@zon) crashes anyway on other CDM decoding methods,
so avoid call `OnStorageId` method is not a safe workaround

better deny in the use of this version, hoping that next CDM version works better
i have no idea if this is a real CDM bug or there is somethingelse that its missing on our interface

## Motivation and context

workaround to #1801 
## Types of change

## Checklist:

### DIFF
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -258,6 +258,17 @@ void CdmAdapter::Initialize()
   }
 
   std::string version{get_cdm_verion_func()};
+
+  if (version == "4.10.2891.0")
+  {
+    // This version have unclear problems
+    // such as crashes on OnStorageId() method calls and others video decoding crashes
+    LOG::Log(LOGERROR,
+             "THE CDM VERSION \"4.10.2891.0\" IS NOT SUPPORTED DUE TO UNCLEAR LIBRARY ISSUES.\n"
+             "------------------------------> PLEASE INSTALL AN OLDER VERSION OF WIDEVINE CDM!");
+    return;
+  }
+
   LOG::LogF(LOGDEBUG, "CDM version: %s", version.c_str());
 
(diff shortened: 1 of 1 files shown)


flubshi:
Would it make sense to make a notification via UI? Or isn't it that easy at this point?

CastagnaIT:
yes it would be great, but currently in kodi completely lacks of an interface to send any kind of message (as feedback)
but i can try to see if use a normal messagebox can be a option

EDIT: I think that in order to use the GUI with a messagebox, you need to study a little bit about how to handle it, because at this point multiple places in the code would require you to display informative error messages, and you need to avoid coupling the GUI code with the CDM library, this would require too many changes and not worth it since it needs to be backported, in the meantime i put this on the todo for future changes

flubshi:
I thought it is as simple as in pvr addons: `kodi::QueueNotification(QUEUE_ERROR, "Title", "Message");`

But I understand - decoupling logic and UI makes sense.

CastagnaIT:
yes all inputstream addons use more or less same interface
the problem is that the code change is into the dependency library that should not get access to kodi APIs
so to route a error feedback out from the dependency require much more code changes
if you have to implement GUI feedback better do it more appropriately from the beginning

CastagnaIT:
@glennguy i don't know if you've tried new approaches to try fix this
but for now i merge this at least prevent kodi to crash and users can read on the log some error details
if you will find a kind of fix, you can always revert it

PR #1824, opened 2025-04-12
Fix regressions and improvements to stream selection

### LINKED ISSUE #1821: [Bug] PR#1724 introduces regression with default AdaptationSet selection
### Describe the problem

There appears to be a change in AdaptationSet selection, introduced in #1724, that attempts to choose a default adaptation set for manifests that have adaptation sets with multiple different codecs:

https://github.com/xbmc/inputstream.adaptive/blob/7e41a82f8733c5993df850e9efb57e131c203a2b/src/Session.cpp#L286-L287
https://github.com/xbmc/inputstream.adaptive/blob/7e41a82f8733c5993df850e9efb57e131c203a2b/src/Session.cpp#L323
https://github.com/xbmc/inputstream.adaptive/blob/7e41a82f8733c5993df850e9efb57e131c203a2b/src/Session.cpp#L338

This change overrides the actual AdaptationSet default attribute supplied by the manifest and also results in default stream selections that are almost certainly undesired (e.g. choosing a 360p AV1 video stream as default, instead of 1080p VP9 or H264 streams) simply based on the fixed codec priority in [PLAYLIST::CAdaptationSet* SESSION::CSession::DetermineDefaultAdpSet()](https://github.com/xbmc/inputstream.adaptive/blob/7e41a82f8733c5993df850e9efb57e131c203a2b/src/Session.cpp#L1169).

While the comments indicate that this is a rough first implementation, the logic used to select the default adaptation set is not really robust enough to be used in its current state as it ignores resolution, framerate, and bitrate, and presumably assumes that multi codec manifests all have adaptation sets with equivalent representations where only the codec varies.

### Possible fix

- Do not override AdaptationSet `@default` provided by the manifest
- Remove the new default adaptation set selection until a less naive approach is implemented. One example would be to prioritise adaption sets based on bits per pixel multiplied by a codec quality factor
- Add support for Quality Equivalence Descriptors and `@qualityRanking` for MPEG-DASH manifests
- Alternatively/additionally add support for `@selectionPriority` for MPEG-DASH manifests
- Add new InputStream.Adaptive property to disable or modify this default adaptation set selection

### Steps to reproduce

_No response_

### Debug log

https://paste.kodi.tv/judaqemedu.kodi

### Stream manifest file(s)

https://paste.kodi.tv/ulelipuyip - 2.7 mb/s 1080p vp9 stream is selected as default rather than the actual default 4.85 mb/s 1080p avc1 stream
https://paste.kodi.tv/emuloyaliq - 0.3 mb/s 360p av1 stream is selected as default rather than the actual default 4.28 mb/s 1080p avc1 stream or the 2.6 mb/s 1080p vp9 stream

### Additional info

_No response_

### Operating system(s)

Windows

_No response_

v22.2.0

### Kodi version(s)

Kodi v22 Alpha 1

CastagnaIT:
## Description

- Fix regression double video quality change with «Ask quality», an attempt to fix has been done on PR #1810 but was not working good, that in some cases cause crashes
- Fix regression on Adaptive quality that dont change while in playback, due to wrong condition on Higher method
- Fix regression on DASH AdaptationSet «default» override parameter for video stream, that was not taken in account
- Add new setting «Limit resolution» to Stream selection «Ask Quality». I thought it might be useful for everyone, when you play multi-codec video a very long list of resolutions can be listed, which you struggle to scroll through, this option limits the list to only the set resolution, at least it tries to limit them based also on the capabilities (streams provided) of the manifest

## Motivation and context

new drm rework has changed some code introducing a codec selection with priority order that was incomplete
and other things was need to be fixed

fix #1821 

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -304,8 +304,12 @@ void SESSION::CSession::InitializePeriod()
     else
       isManualStreamSelection = streamSelectionMode == CHOOSER::StreamSelection::MANUAL;
 
+    const bool isDefaultAdpSet{adp == defVideoAdpSet};
+
     // Get the default initial stream repr. based on "adaptive repr. chooser"
     CRepresentation* defaultRepr{m_reprChooser->GetRepresentation(adp)};
+    if (isDefaultAdpSet)
+      m_reprChooser->LogDetails(defaultRepr);
 
     if (isManualStreamSelection)
     {
@@ -320,9 +324,9 @@ void SESSION::CSession::InitializePeriod()
         if (!currentRepr->isPlayable)
           continue;
 
-        const bool isDefaultRepr{adp == defVideoAdpSet && currentRepr == defaultRepr}; // meant for video only
+        const bool isDefaultVideoRepr{isDefaultAdpSet && currentRepr == defaultRepr};
 
-        AddStream(adp, currentRepr, isDefaultRepr, uniqueId, audioLanguageOrig);
+        AddStream(adp, currentRepr, isDefaultVideoRepr, uniqueId, audioLanguageOrig);
       }
     }
     else
@@ -335,16 +339,14 @@ void SESSION::CSession::InitializePeriod()
       uint32_t uniqueId{adpIndex};
       uniqueId |= reprIndex << 16;
 
-      const bool isDefaultRepr{adp == defVideoAdpSet}; // meant for video only
-
-      AddStream(adp, defaultRepr, isDefaultRepr, uniqueId, audioLanguageOrig);
+      AddStream(adp, defaultRepr, isDefaultAdpSet, uniqueId, audioLanguageOrig);
     }
   }
 }
 
 void SESSION::CSession::AddStream(PLAYLIST::CAdaptationSet* adp,
                                   PLAYLIST::CRepresentation* initialRepr,
-                                  bool isDefaultRepr,
+                                  bool isDefaultVideoRepr,
                                   uint32_t uniqueId,
                                   std::string_view audioLanguageOrig)
 {
@@ -360,7 +362,7 @@ void SESSION::CSession::AddStream(PLAYLIST::CAdaptationSet* adp,
     case StreamType::VIDEO:
     {
       stream.m_info.SetStreamType(INPUTSTREAM_TYPE_VIDEO);
-      if (isDefaultRepr)
+      if (isDefaultVideoRepr)
         flags |= INPUTSTREAM_FLAG_DEFAULT;
       break;
     }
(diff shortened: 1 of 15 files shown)


Ronny-nerd:
Thanks CastagnaIT...it works fine...I noticed on Windows 10 Pro ( 64 BIT ) Kodi 22 Piers.

Android 11...Nvidia Shield TV Pro 2019 = OK

CastagnaIT:
@Ronny-nerd have you tested also the use case of youtube issue https://github.com/xbmc/inputstream.adaptive/issues/1821?
because im not able to test youtube addon, every time ask me to login (using Kodi Cast extension on browser) but signin menu on addon dont works

src/Session.cpp
```diff
@@ -1180,6 +1180,9 @@ PLAYLIST::CAdaptationSet* SESSION::CSession::DetermineDefaultAdpSet()
       CODEC::FOURCC_HEVC, CODEC::FOURCC_AV01, CODEC::NAME_AV1,    CODEC::FOURCC_VP09,
       CODEC::NAME_VP9,    CODEC::FOURCC_AVC_, CODEC::FOURCC_H264};
 
+  CAdaptationSet* overrideAdp{nullptr}; // Override done by manifest custom parameter
```
MoojMidge: This can be simplified to return `currAdp` immediately if it is default, without needing to continue the loops:

```
  CAdaptationSet* defaultAdp{nullptr}; // Default determined by codec order

  for (auto& codecCC : videoCodecOrder)
  {
    CAdaptationSet* currAdp{nullptr};
    uint32_t adpIndex{0};
    while ((currAdp = m_adaptiveTree->GetAdaptationSet(adpIndex++)))
    {
      if (currAdp->GetRepresentations().empty() || currAdp->GetStreamType() != StreamType::VIDEO)
        continue;

      if (currAdp->IsDefault())
      {
        return currAdp;
      }
      else if (CODEC::Contains(currAdp->GetCodecs(), codecCC) && !defaultAdp)
      {
        defaultAdp = currAdp;
      }
    }
  }

  return defaultAdp;
```

CastagnaIT: yeah also one less variable, change done
good to go?

MoojMidge: Sorry, while the other changes seem to make sense don't have a machine/connectivity to properly runtime test. Only really checked the adaptation set default selection on initial playback start.

There have been some reports of both DRM and non-DRM streams failing to play with v22.2.0 but no logs so I don't know what may be causing it.

CastagnaIT: you don't need to test each change, just just to know for your youtube case was ok

PR #1828, opened 2025-04-16
Implemented GUI feedback for playback errors

### LINKED ISSUE #497: IA does not expose playback errors using onPlayBackError()

CastagnaIT:
## Description

Finally users will be happy to know what goes wrong with the playback
without necessarily having to read the debug log

this implementation will not make use of the localizations (language translations) of error details
because it would be too extensive a task and subject to frequent change depending on code change,
the only localization is done in the "generic" error message (Unable to play the stream)

GUI messages cover the most common problems that can occur during playback initialization, such as this example

immagine
immagine

Sidenote:
i have removed `LogF` from cdm library, i noticed only now that was conflicting with `LogF` defines in the utils/log.h

## Motivation and context

When a playback fails for X reasons Kodi most of the time returns to the video list without letting the user know what happened... (sometimes there is a generic message "one or more items failed to play" but is not shown always)
so users are forced to read the debug log to know reasons for playback problems, or randomly try to blame something

a media player usually lets the user know what happened in a minimal way
but this currently is not supported by using InputStream API interface

hint from comment https://github.com/xbmc/inputstream.adaptive/pull/1803#issuecomment-2772783579

and partially fix https://github.com/xbmc/inputstream.adaptive/issues/497
about this last one, they also talk about revoked CDM widevine versions therefore no longer working, in this case we do not have the possibility to verify the situation in advance, the only possible way is to intercept a possible HTTP 400 error from the license request, which however not always this error code can have this specific meaning. theoretically, i think there should be a way to make a call to the ISHelper add-on to initiate a check/download of a different version of the CDM, but this is out of scope of this PR

## Types of change

## Checklist:

### DIFF
--- a/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
+++ b/inputstream.adaptive/resources/language/resource.language.en_gb/strings.po
@@ -352,3 +352,17 @@ msgstr ""
 msgctxt "#30243"
 msgid "Verbose debugging can be useful for debugging components, but in some cases it may expose sensitive information in the log."
 msgstr ""
+
+#empty strings reserved for settings from id 30244 to 30299
+
+#. A GUI window title, "InputStream Adaptive" its the add-on name, not to be translated
+msgctxt "#30300"
+msgid "InputStream Adaptive error"
+msgstr ""
+
+#. A GUI message that is displayed when there is a problem playing a stream
+#. Do not translate placeholders: {error-details}
+#. Use of double graphes e.g. {whatyouwant{error-details}whatyouwant} allows you to add additional text printed only when a value is present
+msgctxt "#30301"
+msgid "Unable to play the stream.{[CR]Error: {error-details}}"
+msgstr ""
--- a/lib/cdm/cdm/debug.h
+++ b/lib/cdm/cdm/debug.h
@@ -20,7 +20,6 @@ enum class LogLevel
 };
 
 void Log(const LogLevel level, const char* format, ...);
-#define LogF(level, format, ...) Log((level), ("%s: " format), __FUNCTION__, ##__VA_ARGS__)
 
 void SetDBGMsgCallback(void (*msgcb)(const LogLevel level, const char* msg));
 
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -193,88 +193,60 @@ CdmAdapter::CdmAdapter(
   const std::string& base_path,
   const CdmConfig& cdm_config,
   CdmAdapterClient *client)
-: library_(0)
-, cdm_path_(cdm_path)
+: cdm_path_(cdm_path)
 , cdm_base_path_(base_path)
 , client_(client)
 , key_system_(key_system)
 , cdm_config_(cdm_config)
-, active_buffer_(0)
-, cdm9_(0), cdm10_(0), cdm11_(0)
 {
   //DCHECK(!key_system_.empty());
-  Initialize();
 }
 
 CdmAdapter::~CdmAdapter()
 {
-  if (cdm9_)
-    cdm9_->Destroy(), cdm9_ = nullptr;
-  else if (cdm10_)
-    cdm10_->Destroy(), cdm10_ = nullptr;
-  else if (cdm11_)
-    cdm11_->Destroy(), cdm11_ = nullptr;
-  else
-    return;
-
-  deinit_cdm_func();
-
-  base::UnloadNativeLibrary(library_);
+  UnloadCDM();
 }
 
-void CdmAdapter::Initialize()
+bool CdmAdapter::LoadCDM()
 {
-  m_isClosingSession = false;
-  if (cdm9_ || cdm10_ || cdm11_)
-  {
-    if (cdm9_)
-      cdm9_->Destroy(), cdm9_ = nullptr;
-    else if (cdm10_)
-      cdm10_->Destroy(), cdm10_ = nullptr;
-    else if (cdm11_)
-      cdm11_->Destroy(), cdm11_ = nullptr;
-    base::UnloadNativeLibrary(library_);
-    library_ = 0;
-  }
+  UnloadCDM();
 
   base::NativeLibraryLoadError error;
   library_ = base::LoadNativeLibrary(cdm_path_, &error);
 
   if (!library_)
   {
-    LogF(LogLevel::ERROR, "Failed to load library: %s", error.ToString().c_str());
-    return;
+    Log(LogLevel::ERROR, "Failed to load library: %s", error.ToString().c_str());
+    return false;
   }
 
-  init_cdm_func = reinterpret_cast<InitializeCdmModuleFunc>(base::GetFunctionPointerFromNativeLibrary(library_, MAKE_STRING(INITIALIZE_CDM_MODULE)));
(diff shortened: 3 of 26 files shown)


ksooo:
@CastagnaIT from architectural point of view I'm asking myself whether an input stream add-on should really have GUI dependencies. I consider those add-ons purely backends, not frontends.

CastagnaIT:
@ksooo mainly the InputStream API interface dont allow a GUI feedback when the `CInputStreamAdaptive::Open` fails for some reason (and other callbacks such as OpenStream)

then users are unable to understand what is not working, and so whether it is something they can fix themselves, or whether they should report it as a possible bug

currently when playback fails for X reason, mainly two things happen:
- Kodi do nothing, when you play a video that fails, kodi return to the video list and nothing more happens, user dont know the reason why kodi is returned to the video list without playing the video
- Kodi (apparently at random) sometimes show a generic error message that dont explain nothing

At the moment i see this as the only possible solution
see screenshots for some examples

PS. anyway was already used GUI also for the manual stream selection (Ask quality)

ksooo:
Firstly, I don't want to block this PR, I'm only sharing some thoughts. If you feel to merge, I will not be in the way.

> the InputStream API interface dont allow a GUI feedback

So that is the actual problem that should be fixed first. Not an easy task, but imo the right approach.

> anyway was already used GUI also for the manual stream selection

That's not an excuse to make it even worse by putting more GUI stuff into the add-on.

If we ever want to have headless Kodi, where ia add-on can be used to serve streams, then things like "stream selection dialog" will just not work. And to make it work at least by means of not failing the add-on then would need to know about that headless mode, just because it mixed backend end frontend functionality together. You see my point?

CastagnaIT:
> You see my point?

yes i had already understood all this since before i did this PR
modifying the IS API is quite chaotic work and i have no clear idea in how this change should be done, since also can affect several callbacks

i don't think it's the right time at today to make big changes on API for the Piers branch, since i guess the first releases aren't long in coming

Have a method to know from the addon whether the headless mode is enabled, yeah can help to handle particular use cases,
but, however i would also expect that in case i make a GUI call (example `kodi::gui::dialogs::OK::ShowAndGetInput`)
kodi should ignore the call when in headless mode, otherwise suggest that there is a underlying problem in kodi core if WindowManager loads GUI in headless mode simply by calling a method...


> where ia add-on can be used to serve streams, then things like "stream selection dialog" will just not work

in case you don't know, the "Ask quality" feature (that make use stream selection dialog) has to be set manually through GUI addon settings or specific properties, whoever sets it "should" know what they are doing, generally if you don't have a GUI you don't go and set up a feature that requires it

ksooo:
> however i would also expect that in case i make a GUI call (example kodi::gui::dialogs::OK::ShowAndGetInput)
kodi should ignore the call when in headless mode

Problem with that is that kodi can’t know how to properly ignore the call. Does ignore mean ‘cancel’, is it automatically return an empty string, the first entry of a list, … ?

> generally if you don't have a GUI you don't go and set up a feature that requires it

works by luck for this settings, may not for others.

CastagnaIT:
> Problem with that is that kodi can’t know how to properly ignore the call. Does ignore mean ‘cancel’, is it automatically return an empty string, the first entry of a list, … ?

for example in the use case of `kodi::gui::dialogs::Select::Show` used for the stream selection dialog
have a argument "selected" that is used to allow a default selection
so could be an idea use that argument as "ignore" behaviour, but more likely not all methods can have this possibility

i agree that knowing in advance if kodi is headless allows to have better code behaviours

ksooo:
> i agree that knowing in advance if kodi is headless allows to have better code behaviours

That is not what I wanted to say. I'm not suggesting an API allowing add-ons to detect headless state. My suggestion is to clearly separate components into GUI and backend (non-GUI) components, not to mix. An an inputstream add-on in my ideal world would clearly be a non-GUI component.

CastagnaIT:
in your ideal world, you can't force binary addons to not have GUI access
a compromise has to be found, so i prefer my vision which dont deny GUI access for all the benefit of users

moreover in the future (although this is still in the theoretical stage atm) i want to try implement calls to external python addons
in order to perform automated widevine cdm installations (and maybe updates)
and so called addons will also require GUI accesses
PS. you cant play DRM videos if you dont install widevine, to install widevine you need GUI...

What i don't understand is how the headless mode can make streaming video display

ksooo:
Well, let’s agree to disagree here. 

> What i don't understand is how the headless mode can make streaming video display

That’s the whole point of headless. It would display nothing. It would act as a server, to deliver streams, maybe after transcoding original stream, stuff like that.

CastagnaIT:
> would act as a server, to deliver streams, maybe after transcoding original stream, stuff like that.

DRM videos work by having direct access to the HDCP hardware monitor and assume that the signal should never be intercepted in the middle, what you want to do means intercepting the a/v signal in some way with hacks
so you're browsing in the realm of illegality... this will allow a flood of illegal DRM video copies to spread

By doing this you can say goodbye to the Kodi software, the project will be closed by the next lawyer...
so i would avoid make headless mode running with these (this) addons

CastagnaIT:
add future todos to roadmap https://github.com/orgs/xbmc/projects/7?pane=issue&itemId=107433484

flubshi:
@CastagnaIT wow, nice addition!! I think it helps a lot of users to understand what is wrong. 

One of the most frequently issue users of pvr.waipu are reporting to me is "it is not working", which is usually caused by missing/wrong Widevine installation. If users now see the error messages in UI, they can hopefully help themselves. 
Great! Thank you!

ksooo:
> wow, nice addition!! I think it helps a lot of users to understand what is wrong.

There is light and there is dark, with this. Now, the user gets presented two error message boxes (at least this happened to me yesterday). First box is raised by the addon, second raised by Kodi (the usual "some blabla failed to play. See log for blabla...")

And, for me the addon error message was like "failed to parse manifest". Not sure a normal user understands that very technical stuff. For me, those messages are rather something for a debug/error entry in the log file.

CastagnaIT:
> There is light and there is dark, with this. Now, the user gets presented two error message boxes (at least this happened to me yesterday). First box is raised by the addon, second raised by Kodi (the usual "some blabla failed to play. See log for blabla...")

this is quite random, not always kodi core shown a kind of "error" message when playback fails
sometimes return to video list without any message
and sometimes show meaningless message such as "one or more items failed to play"

Edit: also one case is, when you start playback but ISA for x reasons do not provide packets to the demuxer, and so VP stops without any message, VP thinks the playback is finished but behind there is another problem if ISA dont provide packets

despite sometimes there is a chance that two messages will be shown, it is always better now than before

if for you this is very annoying
i can add a new addon setting that allow to disable GUI messages

> And, for me the addon error message was like "failed to parse manifest". Not sure a normal user understands that very technical stuff. For me, those messages are rather something for a debug/error entry in the log file.

yes some errors are more technical, but for some of them i could use a more general message that suggest to read the log for more details about the problem

ksooo:
> if for you this is very annoying
> i can add a new addon setting that allow to disable GUI messages

Thanks, no need for that. Gladly, I encounter isa errors really very seldom.

PR #1852, opened 2025-05-04
gcc 15 build fixes

BlackIkeEagle:
```
error: ‘uint16_t’ does not name a type
```

add `#include <cstdint>`

## Description

gcc 15 build fix cherry-pick #1851 for Piers

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveUtils.h
+++ b/src/common/AdaptiveUtils.h
@@ -13,6 +13,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <cstdint>
 
 // forwards
 class AP4_Movie;


CastagnaIT:
the fix its ok but doesnt make much sense to do it in this way
would be a continuous nonsense cat-and-mouse
no one will take responsibility for testing whether each new PR will be compatible or not with GCC15 before the merging

GCC15 has just been released for less than a month and currently i dont see prebuild packages for OSs in particular for Ubuntu where our CI live, i would avoid having the task of maintaining other custom CIs for being too far ahead of the times

before introducing any fix for GCC15 i suggest to make sure that we can implement an appropriate CI with officially provided packages, when they will be available

Sidenote:
you marked that you read Contributing guidelines, but did not follow them correctly...
- The commit title must reflect the code change and not the general context of the reasons behind it, it also lacks the appropriate prefix
- A PR should be created initially for the development "master" branch, and later time a backport if allowed, NEVER in reverse order so avoid create a PR for the development master branch based from old branches! this is not a thing to do...

so for now i will put this PR on hold

PR #1859, opened 2025-05-09
[Cleanup] Make GUI error messages less technical

CastagnaIT:
## Description

Make GUI error messages less technical
## Motivation and context

as suggested from comment: https://github.com/xbmc/inputstream.adaptive/pull/1828#issuecomment-2844444448

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -100,7 +100,7 @@ SResult SESSION::CSession::Initialize(std::string manifestUrl)
 
   CURL::HTTPResponse manifestResp;
   if (!CURL::DownloadFile(manifestUrl, manifestHeaders, {"etag", "last-modified"}, manifestResp))
-    return SResult::Error("Cannot download the manifest file.");
+    return SResult::Error(GUI::GetLocalizedString(30307));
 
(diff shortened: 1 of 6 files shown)


CastagnaIT:
CC @ksooo

src/Session.cpp
```diff
@@ -115,12 +115,18 @@ SResult SESSION::CSession::Initialize(std::string manifestUrl)
 
   m_adaptiveTree = PLAYLIST_FACTORY::CreateAdaptiveTree(manifestResp);
   if (!m_adaptiveTree)
-    return SResult::Error("Unable to determine type of manifest file.");
+  {
+    LOG::LogF(LOGERROR, "Unable to determine type of manifest file.");
+    return SResult::Error("A problem occurred in the stream initialization. See the log for details.");
```
ksooo: 1) If you want to make the messages less technical, then pointing users to the log is not a good approach. Most of the users have no idea how to read the log, and many of them do not even know where to find the log. For Kodi 22 we removed the "see log for details" part of many error message strings., for exactly this reason

2) If this is for end users, these string should be localizable.

CastagnaIT: done

Issue #1495, opened 2024-02-25
[K21][MPD] [live stream] playback stopped after 16 sec.


mtr81:
### Describe the bug
Here is a clear and concise description of what the problem is:

Playback started and stopped after 16 sec.  Continuation issue https://github.com/xbmc/inputstream.adaptive/issues/1446
In Kodi 20 playback didn't start (no fix "[Dash] Removed segment limit based on SIDX refs" for this Kodi version) https://github.com/xbmc/inputstream.adaptive/pull/1478 )

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

Playback live stream correctly.

## Actual Behavior

Playback stopped after 16 sec.

## Possible Fix

---

### To Reproduce

1.  Start playback manifest MPD

### Debuglog

The debuglog can be found here:

[log.txt](https://github.com/xbmc/inputstream.adaptive/files/14395096/log.txt)

### MPD/M3U8s/ISM

An example or copy of a manifest (or manifests for HLS - master and variants) can be found here:
Manifest MPD
[manifest.mpd.txt](https://github.com/xbmc/inputstream.adaptive/files/14395099/manifest.mpd.txt)

### Screenshots 
Here are some links or screenshots to help explain the problem:

![megogo_2024-02-25](https://github.com/xbmc/inputstream.adaptive/assets/86961563/0d1ecd1a-00b0-4bf0-a3ea-e3d89da97a77)

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

---

### Your Environment
Used Operating system:

 - [x] Android
 - [ ] iOS
 - [ ] tvOS
 - [ ] Linux
 - [ ] OSX
 - [x] Windows
 - [ ] Windows UWP

 - Operating system version/name: Windows 10
 - Kodi version: 21.0-BETA3 (20.90.830) Git:20240221-480476cb59
 - inputstream.adaptive version: 21.4.3 (from https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1492/4/artifacts)

*note: Once the issue is made we require you to update it with new information should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

CastagnaIT:
here lack of MPD manifest updates
by reading the manifest you have a unusual MPD data where its missing `minimumUpdatePeriod` attribute
[...]

PR #1867, opened 2025-05-23
[DashTree] Timeline improvements for live streams

### LINKED ISSUE #1769: [Bug] dash-stream: can't go past start point or back to live
### Describe the problem

It is not possible to move to "live" or e.g. +15 min after start point of the stream without pressing "stop" and starting again.
I'm using ISA (+ iptv-simple) together with some german dash-streams. (see below)

example below (How to reproduce):
I want to move past 20:00 (actual time 21:30). this is not possible.

I started the dash stream, went back 30 min, let it play for 3,5 min and tried to went forward 30 min to "live". But the stream only went forward 26,5 min to the point, where I started the stream. Shouldn't it to to "live"? It is a (live)stream ...

So ISA should update the endpoint (live)  in some way ...
Thx

Here is the forum-post: (put it together for this bug-report):
https://forum.kodi.tv/showthread.php?tid=378875&pid=3210333#pid3210333

### Possible fix

_No response_

### Steps to reproduce

In the stream, I can go back 2 hrs (from 20:00 to 18:00).
streamtime 18:00
actual time 20:00

If I go back 2h and watch the stream e.g. 1h30 min (at 19:30) and move forward because of advertising or end of the show, it is not possible to move to actual "live".
streamtime 19:30
actual time 21:30
I can only go to 20:00 streamtime and not past 20:00.
Kodi only goes forward to the start point of the stream (1h30 min ago) (20:00), where I moved backwards.

### Debug log

was too big for 1 part
part 1: https://paste.kodi.tv/kavomuzeja.kodi
part 2: https://paste.kodi.tv/oduzeyiqos.kodi
part 3: https://paste.kodi.tv/avisohokek.kodi
part 4: https://paste.kodi.tv/ujakayupim.kodi

### Stream manifest file(s)

The free (and legal) stream is geoblocked (german) from here: https://github.com/jnk22/kodinerds-iptv --> dash
Tested with ARD (Das Erste) and ZDF
ARD: https://paste.kodi.tv/ayavecoqeq
ZDF: https://paste.kodi.tv/azumucukaj

### Additional info

_No response_

### Operating system(s)

Android

Android 11 on Philips TV

21.5.9

### Kodi version(s)

21.1

CastagnaIT:
## Description

While investigating to fix issue i noticed various problems, some due to partial implementations that were never completed

in short
- implementation of TSB timeline update between periods for MPD that use SegmentTemplate without a defined TimeLine,
the code relative to TSB timeline update, will check also for expired segments that fall outside the TSB, then delete them all in realtime, this fix the #1769 issue because previously expired segments was never deleted and the video seek was pointing always to the same (old) segment, with the problem that was impossible seek back to "live"
- implementation of the removal of expired periods, where previously they were persistently kept in memory, with the side effect of a small "memory leak" as well
- add dynamic update of period/chapter duration, this will allow the passage of time to be displayed appropriately in the GUI seek bar
- Reverted dash segment-controlled live stream support (implemented by #1525), this is now not more needed since segments for templated representations are now inserted while in playback without any restrictions. This use case has been tested by user see comments below

## Motivation and context

fix #1769 

## How has this been tested?

- Live Dash with segment templated representations (no defined segments)
example on ISASamples addon test stream "dashif periods_20 multi-period"
- tested also live HLS having long timeshift

some users also tested

## Types of change

## Checklist:


CastagnaIT:
@mtr81 can you help me to do a test?
time ago you have opened this issue #1495, for a specific live stream that was not working,
have you the possibility to test that same live stream to see if it works with this PR? (kodi 22)

Test builds: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1867/3/artifacts

mtr81:
Hi
Playback stops after several seconds.
[log.txt](https://github.com/user-attachments/files/21200872/log.txt)

CastagnaIT:
@mtr81 i made a fix, new build:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1867/4/artifacts

mtr81:
Works fine. Returning to live mode after rewinding to previous segments also works as expected.

CastagnaIT:
perfect thanks

src/Session.cpp
```diff
@@ -719,6 +719,15 @@ void CSession::EnableStream(CStream* stream, bool enable)
...
 }
 
+uint64_t SESSION::CSession::GetTotalTimeMs() const
+{
+  // In live streaming do not take into account the live delay duration, because its not seekable
+  if (m_adaptiveTree->IsLive() && m_adaptiveTree->m_totalTime > m_adaptiveTree->m_liveDelay * 1000)
+    return m_adaptiveTree->m_totalTime - m_adaptiveTree->m_liveDelay * 1000;
+  else
+    return m_adaptiveTree->m_totalTime;
+}
```
CastagnaIT, line 729: PS: I wrote that the live delay part should not take into account, this is to prevent the user from attempting to seek in this part of the video (being the GetTotalTimeMs method is involved to the GUI bar to shown the duration of the video). This is because if you attempt to seek in to live delay part, ISA try limits this attempt on CSession::SeekTime, but Kodi core does not care about this limit and Kodi core keeps dropping packets until it gets the PTS it expects on packets, this can stall playback in buffering for several seconds and should be avoided

Issue #1874, opened 2025-06-03
Loading the LG supplied widevine lib on webOS


cscd98:
I have been looking at widevine and I've managed to get inputstream.adaptive to load the LG supplied widevine lib. webOS uses armv7a softfp so there is no other option.

It appears to be getting a license for the content:

```
2025-06-03 18:52:00.878 T:13286   debug <general>: AddOnLog: inputstream.adaptive: [WV-CDM-Library] CDM version: 17.1.1
2025-06-03 18:52:00.901 T:13286   debug <general>: AddOnLog: inputstream.adaptive: [WV-CDM-Library] CDM is initialized: true
2025-06-03 18:52:00.974 T:13289   debug <general>: AddOnLog: inputstream.adaptive: CDM message: type 1 arrived
2025-06-03 18:52:00.974 T:13289   debug <general>: AddOnLog: inputstream.adaptive: SetSession: Opened widevine session ID: xyz
2025-06-03 18:52:00.975 T:13286   debug <general>: CurlFile::Open - <https://cassie.channel5.com/api/v2/licences/widevine/491/xyz?expiry=1748983920&tag=blah>
2025-06-03 18:52:00.976 T:13286   debug <general>: easy_acquire - Created session to https://cassie.channel5.com/
2025-06-03 18:52:01.367 T:13289   debug <general>: AddOnLog: inputstream.adaptive: [WV-CDM-Library] OnSessionKeysChange: KID blah, Status: 0, System code: 0
2025-06-03 18:52:01.367 T:13289   debug <general>: AddOnLog: inputstream.adaptive: CDM message: type 3 arrived
2025-06-03 18:52:01.367 T:13289   debug <general>: AddOnLog: inputstream.adaptive: [WV-CDM-Library] OnSessionKeysChange: KID blah, Status: 0, System code: 0
2025-06-03 18:52:01.367 T:13289   debug <general>: AddOnLog: inputstream.adaptive: CDM message: type 3 arrived
2025-06-03 18:52:01.367 T:13286   debug <general>: AddOnLog: inputstream.adaptive: License update successful
```

and it seems to be fetching & decrypting data:
```
2025-06-03 22:25:34.315 T:4995    debug <general>: AddOnLog: inputstream.adaptive: CDM - ReadSample()
2025-06-03 22:25:34.315 T:4995    debug <general>: AddOnLog: inputstream.adaptive: CDM - CWVCencSingleSampleDecrypter::DecryptSampleData()
[...]

PR #1893, opened 2025-07-24
[Widevine] Add webOS support (L3)

cscd98:
## Description
This adds L3 Widevine support to the webOS port of Kodi.

## Motivation and context
webOS cannot use chromium supplied widevine libs as it still uses a 32-bit softfp tool chain. However, LG have their own widevine libraries installed anyway.

I am investigating the usage of secure video path which **might** give us L1 but in the mean time L3 is here, and webOS does not have anything at the moment so this is a major improvement.

partially solve #1874 (secure decoder still wip)

## How has this been tested?
Tested with Google Tears [CBCS] and real usage testing by [Uukrull](https://github.com/Uukrull) with Amazon Prime, Movistar+ and DAZN.

## How did I go about this?
https://opensource.lge.com/
Search for your TV (e.g. OLED65C5)
Download firmware source and extract chromium120 source.

In webOS they use an additional parameter to determine the type of stream:

src/media/cdm/cdm_adapter.cc
```
void CdmAdapter::Decrypt(StreamType stream_type,
                         scoped_refptr<DecoderBuffer> encrypted,
                         DecryptCB decrypt_cb) {
<snip>
#if defined(USE_NEVA_MEDIA)
  cdm::Status status = cdm_->Decrypt(input_buffer, decrypted_block.get(),
                                     ToCdmStreamType(stream_type));
#else
  cdm::Status status = cdm_->Decrypt(input_buffer, decrypted_block.get());
#endif
```

src/media/cdm/cdm_wrapper.h
```
  virtual cdm::Status Decrypt(const cdm::InputBuffer_2& encrypted_buffer,
#if defined(USE_NEVA_MEDIA)
                              cdm::DecryptedBlock* decrypted_buffer,
                              const cdm::StreamType decoder_type) = 0;
#else
                              cdm::DecryptedBlock* decrypted_buffer) = 0;
#endif
```

The presence of SVP headers was detected through HEX dumps of the decrypted packet.

As for libwidevine-wrapper and libwvcdm_shared, these are only found by SSH to the TV and change based on how old your webOS is.

## Types of change

## Checklist:

### DIFF
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -228,9 +228,11 @@ bool CdmAdapter::Initialize()
 #endif  // defined(OS_WIN)
 
   init_cdm_func();
-
+  
+#ifndef TARGET_WEBOS
   cdm12_ = static_cast<cdm::ContentDecryptionModule_12*>(create_cdm_func(
       12, key_system_.data(), static_cast<uint32_t>(key_system_.size()), GetCdmHost, this));
+#endif
 
   if (!cdm12_)
   {
@@ -441,9 +443,21 @@ cdm::Status CdmAdapter::Decrypt(const cdm::InputBuffer_2& encrypted_buffer,
   if (cdm12_)
     ret = cdm12_->Decrypt(encrypted_buffer, decrypted_buffer);
   else if (cdm11_)
+#ifdef TARGET_WEBOS
+    // we set this to kStreamTypeAudio, as this bypasses automatic SVP header/meta data injection
+    // this is lost in RepackSubsampleData
+    ret = cdm11_->Decrypt(encrypted_buffer, decrypted_buffer, cdm::StreamType::kStreamTypeAudio);
+#else
     ret = cdm11_->Decrypt(encrypted_buffer, decrypted_buffer);
+#endif
   else if (cdm10_)
+#ifdef TARGET_WEBOS
+    // we set this to kStreamTypeAudio, as this bypasses automatic SVP header/meta data injection
+    // this is lost in RepackSubsampleData
+    ret = cdm10_->Decrypt(encrypted_buffer, decrypted_buffer, cdm::StreamType::kStreamTypeAudio);
+#else
     ret = cdm10_->Decrypt(encrypted_buffer, decrypted_buffer);
+#endif
 
   active_buffer_ = 0;
   return ret;
--- a/.github/workflows/build-webos.yml
+++ b/.github/workflows/build-webos.yml
@@ -1,0 +1,73 @@
+name: Build for webOS
+on: [push, pull_request]
+env:
+  app_id: inputstream.adaptive
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        ZIP_NAME_SUFFIX: [webos]
+    steps:
+    - name: Install system cmake
+      run: |
+        sudo apt-get update
+        sudo apt-get install cmake
+        echo "System CMake installed at: $(which cmake)"
+        cmake --version
+    - name: Download webOS NDK
+      uses: robinraju/release-downloader@v1.11
+      with:
+        repository: "openlgtv/buildroot-nc4"
+        latest: true
+        fileName: "linux-intel-webos-sdk.tar.gz"
+        out-file-path: "/tmp"
+    - name: Extract webOS NDK
+      working-directory: /tmp
+      run: |
+        tar xzf linux-intel-webos-sdk.tar.gz
+        ./arm-webos-linux-gnueabi_sdk-buildroot/relocate-sdk.sh
+    - name: Checkout Kodi repo
+      uses: actions/checkout@v4
+      with:
+        repository: xbmc/xbmc
+        ref: master
+        path: xbmc
+    - name: Bootstrap depends
+      run: |
+        cd xbmc/tools/depends
+        ./bootstrap
+        ./configure --prefix=${{ github.workspace }}/xbmc-deps \
+                    --with-toolchain=/tmp/arm-webos-linux-gnueabi_sdk-buildroot --host=arm-webos-linux-gnueabi \
+                    --with-platform=webos --enable-debug=no
+    - name: Checkout add-on repo
+      uses: actions/checkout@v4
+      with:
+        path: ${{ env.app_id }}
+    - name: Configure
+      run: |
+        cd ${app_id}
+        mkdir -p build && cd build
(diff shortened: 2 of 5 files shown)


Uukrull:
@CastagnaIT I tried this code and the challenge sends _oem_crypto_security_patch_level 0_, compared to L3 on Windows that sends _oem_crypto_security_patch_level 02_.

Does oem_crypto_security_patch_level 0 mean L1? I have my Shield TV disconnected so I can't check the security level with it.

This is the challenge data:

```
company_name LG
model_name OLED48C44LA.DEUQDJP
model_year 2024
architecture_name aarch64
device_name o22n2
product_name webOS TV
build_info S17.1.1 | webOS | TV | aarch64 | CE CDM 17.1.1 | AArch32 | Logging Enabled | Release
widevine_cdm_version 17.1.1
oem_crypto_security_patch_level 0
oem_crypto_build_information Widevine OPK v17.0.2 WVUTA2
```

cscd98:
I think it probably means L3. We need to get SVP working and that might expose L1 but I'm not 100% until I see it.

lib/cdm/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -443,7 +443,13 @@ cdm::Status CdmAdapter::Decrypt(const cdm::InputBuffer_2& encrypted_buffer,
   else if (cdm11_)
     ret = cdm11_->Decrypt(encrypted_buffer, decrypted_buffer);
   else if (cdm10_)
+#ifdef TARGET_WEBOS
+    // we set this to kStreamTypeAudio, as this bypasses automatic SVP header/meta data injection
+    // this is data that is lost when entering the media pipeline
+    ret = cdm10_->Decrypt(encrypted_buffer, decrypted_buffer, cdm::StreamType::kStreamTypeAudio);
```
CastagnaIT: Clearly a hack,
is it worth waiting to merge the PR to see if you make progress towards a more appropriate solution?

and in case IMO it would make sense to apply this also to cdm11 and cdm12 as well,
without having to do another PR in the future to avoid left users in a broken state

CastagnaIT:
It would be appreciated if you could include in the PR description also the reference links (on source code or other) on which you based your studies on these changes, ofc if freely available
this would help future developers

One thing we are missing now is at least a CI for webOS, as no one here is able to test the code

lib/cdm/cdm/media/cdm/cdm_adapter.cc
cscd98: @CastagnaIT I have been working on this, but so far haven't gotten it to work. I have managed to alter the pipeline to pass the streamType all the way to this function, but passing it correctly as Video means it adds SVP headers and meta data.

But this information then gets lost during repackSubsampleData.

The SVP headers are however not required for playback anyway in L3 mode so they are totally fruitless unless your in secure video path mode.

webOS does not use cdm11/12, I can add it if required.

cscd98:
> It would be appreciated if you could include in the PR description also the reference links (on source code or other) on which you based your studies on these changes, ofc if freely available this would help future developers
> 
> One thing we are missing now is at least a CI for webOS, as no one here is able to test the code

I've added some information to the PR description under "How did I go about this?"

@wsnipex has an LG TV if hes available to do so.

cscd98:
@CastagnaIT Here is an example of a CI I've added in the past for retroarch:

```      - name: Download webOS NDK
        uses: robinraju/release-downloader@v1.11
        with:
          repository: "openlgtv/buildroot-nc4"
          latest: true
          fileName: "linux-intel-webos-sdk.tar.gz"
          out-file-path: "/tmp"

      - name: Extract webOS NDK
        shell: bash
        working-directory: /tmp
        run: |
          tar xzf linux-intel-webos-sdk.tar.gz
          ./arm-webos-linux-gnueabi_sdk-buildroot/relocate-sdk.sh

      - name: Compile RA
        shell: bash
        run: |
          . /tmp/arm-webos-linux-gnueabi_sdk-buildroot/environment-setup
          make -f Makefile.webos ipk PACKAGE_NAME=${PACKAGE_NAME} ADD_SDL2_LIB=1 -j$(getconf _NPROCESSORS_ONLN)
        env:
          DEBUG: ${{ github.event_name == 'release' && '0' || '1' }}
```

We could modify this for IA?

lib/cdm/cdm/media/cdm/cdm_adapter.cc
CastagnaIT: i have take a look at LG sources, they make use of custom Decrypt method also for cdm 11 interface,
although they may not be using it today, it is ready for use
so yes please add the same change on cdm 11 interface

The only unknown factor is that they currently still use an old version of content_decryption_module.h,
so we cannot be certain what they will do when they update to introduce support to cdm12 interface

I think the best thing to do for now is to avoid checking for cdm12 interface for webOS case
so i suggest to exclude following code for webOS by ifdeffing it
https://github.com/xbmc/inputstream.adaptive/blob/f20627945db08a02299eb494c648b8b3e9ba7ad5/lib/cdm/cdm/media/cdm/cdm_adapter.cc#L232-L233
until they update the interface code

cscd98: I have added it to cdm11 as well as added a check for cdm12 to bypass.

I've also wrote a github actions for webOS

CastagnaIT:
> I have added it to cdm11 as well as added a check for cdm12 to bypass.
> I've also wrote a github actions for webOS

thank you great work!
do you want that i merge & release, or do you prefer wait for further improvements?

cscd98:
@CastagnaIT Go ahead and release it :+1: I don't know how long it might take to get secure video path mode working properly.

CastagnaIT:
look like there is a problem with webOS CI about NDK file
https://github.com/xbmc/inputstream.adaptive/actions/runs/16563242800/job/46837529631
maybe temporary problem? can you take a look? 
immagine

Uukrull:
[script.module.inputstreamhelper](https://github.com/emilsvennesson/script.module.inputstreamhelper) must be updated so it doesn't ask about installing the Widevine library like on Android. I have a working script that does that.

cscd98:
@CastagnaIT The naming scheme was changed 13 hours ago :+1: I'm doing a PR now to reflect the new name

cscd98:
> [script.module.inputstreamhelper](https://github.com/emilsvennesson/script.module.inputstreamhelper) must be updated so it doesn't ask about installing the Widevine library like on Android. I have a working script that does that.

Could you do a PR for it?

Uukrull:
> Could you do a PR for it?

Sure, no problem.

Uukrull:
Done: https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/592

CastagnaIT:
Thank you both

PR #1900, opened 2025-08-21
webOS: Implement L1 secure path decoding

cscd98:
## Description
webOS supports L1 decoding - so we add this support for 4K

## Motivation and context
**Implements 4K UHD and DV support for Disney+, Prime.** 

The way this works is that the secure packet is routed through the CDM. This does not decrypt the packet, it is still encrypted, instead it is wrapped in SVP headers and then sent encrypted into the media pipeline.

It defaults to L1 for webOS from now on.

Other changes in this PR:

- Fixes Annex B conversion, before it was over writing bytes causing a blank screen
- Makes use of unimplemented call backs for cdm initialization and session creation in case of race condition
- Adds Dolby Vision meta data parsing and pass through to Kodi using Inputstream interface (moved to a separate PR: https://github.com/xbmc/inputstream.adaptive/pull/1920 due to breaking change in the interface)

## How has this been tested?
Tested on webOS 24 by myself using Disney+ and Prime addons.
Tested by @Uukrull also
Also tested on ITV X (which is 1080p max and no DV).

## Types of change

## Checklist:

### DIFF
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -61,6 +61,8 @@ const char PATH_SEPARATOR = '\\';
 const char PATH_SEPARATOR = '/';
 #endif
 
+constexpr std::chrono::seconds INIT_FUTURE_TIMEOUT_SEC{3};
+
 void* GetCdmHost(int host_interface_version, void* user_data)
 {
   if (!user_data)
@@ -216,6 +218,9 @@ bool CdmAdapter::LoadCDM()
 
 bool CdmAdapter::Initialize()
 {
+  m_initPromise = std::promise<void>{};
+  m_initFuture = m_initPromise.get_future();
+  m_provisioningCompleteOrStarted = false;
   m_isClosingSession = false;
 
   Log(LogLevel::DEBUG, "CDM version: %s", GetVersion().c_str());
@@ -263,6 +268,22 @@ bool CdmAdapter::Initialize()
       cdm10_->Initialize(cdm_config_.allow_distinctive_identifier,
                          cdm_config_.allow_persistent_state, false);
     }
+
+    // Wait for the CDM to be initialized
+    // Add a maximum timeout in case we never hear back!
+    if (m_initFuture.valid() &&
+        m_initFuture.wait_for(INIT_FUTURE_TIMEOUT_SEC) != std::future_status::ready)
+    {
+      Log(LogLevel::ERROR, "CDM initialization timed out");
+      return false;
+    }
+
+    if (!m_provisioningCompleteOrStarted)
+    {
+      Log(LogLevel::ERROR, "CDM initialization failed or not started");
+      return false;
+    }
+
     return true;
   }
 
@@ -786,6 +807,15 @@ void CdmAdapter::ReportMetrics(cdm::MetricName metric_name, uint64_t value)
 
 void CdmAdapter::OnInitialized(bool success)
 {
+  // Notify the client that the CDM is initialized
+  if (success)
+  {
+    m_provisioningCompleteOrStarted = true;
+  }
+
+  // Set the promise value so that execution can continue
+  m_initPromise.set_value();
+
   Log(LogLevel::DEBUG, "CDM is initialized: %s", success ? "true" : "false");
 }
 
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -613,10 +613,32 @@ void CdmAdapter::OnResolvePromise(uint32_t promise_id)
 {
 }
 
-void CdmAdapter::OnResolveNewSessionPromise(uint32_t promise_id,
-                      const char* session_id,
-                      uint32_t session_id_size)
+std::future<std::string> CdmAdapter::PrepareSessionFuture(uint32_t promiseId)
 {
+  std::lock_guard<std::mutex> lock(m_sessionMx);
+  std::promise<std::string> promise;
+  auto future = promise.get_future();
+  m_sessionPromises[promiseId] = std::move(promise);
+  return future;
+}
+
+void CdmAdapter::OnResolveNewSessionPromise(uint32_t promise_id,
+                                            const char* session_id,
+                                            uint32_t session_id_size)
+{
+  std::string sessionStr(session_id, session_id_size);
+
+  std::lock_guard<std::mutex> lock(m_sessionMx);
+  auto it = m_sessionPromises.find(promise_id);
+  if (it != m_sessionPromises.end())
+  {
+    it->second.set_value(sessionStr);
+    m_sessionPromises.erase(it);
+  }
+  else
+  {
+    Log(LogLevel::ERROR, "Promise ID %u not found", promise_id);
+  }
 }
 
 void CdmAdapter::OnSessionKeysChange(const char* session_id,
@@ -709,9 +731,37 @@ void CdmAdapter::OnResolveKeyStatusPromise(uint32_t promise_id, cdm::KeyStatus_2
 {
 }
 
-void CdmAdapter::OnRejectPromise(uint32_t promise_id, cdm::Exception exception,
-  uint32_t system_code, const char* error_message, uint32_t error_message_size)
+void CdmAdapter::OnRejectPromise(uint32_t promise_id,
+                                 cdm::Exception exception,
+                                 uint32_t system_code,
+                                 const char* error_message,
+                                 uint32_t error_message_size)
 {
+  std::lock_guard<std::mutex> lk(m_sessionMx);
+  auto it = m_sessionPromises.find(promise_id);
+  if (it == m_sessionPromises.end())
+    return;
+
+  try
+  {
+    std::string msg;
+    if (error_message && error_message_size)
+    {
+      msg.assign(error_message, error_message_size);
+    }
+    else
+    {
+      msg = "CDM reject: " + std::to_string(static_cast<int>(exception)) +
+            " sys=" + std::to_string(system_code);
+    }
+
+    it->second.set_exception(std::make_exception_ptr(std::runtime_error(msg)));
+  }
+  catch (...)
+  {
+  }
+
+  m_sessionPromises.erase(it);
 }
 
 void CdmAdapter::OnSessionMessage(const char* session_id, uint32_t session_id_size,
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -429,7 +429,8 @@ void CdmAdapter::TimerExpired(void* context)
 }
 
 cdm::Status CdmAdapter::Decrypt(const cdm::InputBuffer_2& encrypted_buffer,
-  cdm::DecryptedBlock* decrypted_buffer)
+                                cdm::DecryptedBlock* decrypted_buffer,
+                                cdm::StreamType streamType)
 {
   //We need this wait here for fast systems, during buffering
   //widewine stopps if some seconds (5??) are fetched too fast
@@ -444,17 +445,13 @@ cdm::Status CdmAdapter::Decrypt(const cdm::InputBuffer_2& encrypted_buffer,
     ret = cdm12_->Decrypt(encrypted_buffer, decrypted_buffer);
   else if (cdm11_)
 #ifdef TARGET_WEBOS
-    // we set this to kStreamTypeAudio, as this bypasses automatic SVP header/meta data injection
-    // this is lost in RepackSubsampleData
-    ret = cdm11_->Decrypt(encrypted_buffer, decrypted_buffer, cdm::StreamType::kStreamTypeAudio);
+    ret = cdm11_->Decrypt(encrypted_buffer, decrypted_buffer, streamType);
 #else
     ret = cdm11_->Decrypt(encrypted_buffer, decrypted_buffer);
 #endif
   else if (cdm10_)
 #ifdef TARGET_WEBOS
-    // we set this to kStreamTypeAudio, as this bypasses automatic SVP header/meta data injection
-    // this is lost in RepackSubsampleData
-    ret = cdm10_->Decrypt(encrypted_buffer, decrypted_buffer, cdm::StreamType::kStreamTypeAudio);
+    ret = cdm10_->Decrypt(encrypted_buffer, decrypted_buffer, streamType);
 #else
     ret = cdm10_->Decrypt(encrypted_buffer, decrypted_buffer);
 #endif
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.cc
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.cc
@@ -256,17 +256,17 @@ bool CdmAdapter::Initialize()
     if (cdm12_)
     {
       cdm12_->Initialize(cdm_config_.allow_distinctive_identifier,
-                         cdm_config_.allow_persistent_state, false);      
+                         cdm_config_.allow_persistent_state, cdm_config_.use_hw_secure_codecs);
     }
     else if (cdm11_)
     {
       cdm11_->Initialize(cdm_config_.allow_distinctive_identifier,
-                         cdm_config_.allow_persistent_state, false);
+                         cdm_config_.allow_persistent_state, cdm_config_.use_hw_secure_codecs);
     }
     else if (cdm10_)
     {
       cdm10_->Initialize(cdm_config_.allow_distinctive_identifier,
-                         cdm_config_.allow_persistent_state, false);
+                         cdm_config_.allow_persistent_state, cdm_config_.use_hw_secure_codecs);
     }
 
     // Wait for the CDM to be initialized
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.h
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.h
@@ -221,7 +221,6 @@ class CdmAdapter : public std::enable_shared_from_this<CdmAdapter>,
 
   void OnInitialized(bool success) override;
 
-
   std::future<std::string> PrepareSessionFuture(uint32_t promiseId);
 
   //Misc
@@ -269,6 +268,10 @@ private:
   CdmConfig cdm_config_;
 
   cdm::Buffer* active_buffer_{nullptr};
+
+  std::promise<void> m_initPromise;
+  std::future<void> m_initFuture;
+  std::atomic<bool> m_provisioningCompleteOrStarted;
 
   cdm::ContentDecryptionModule_10* cdm10_{nullptr};
   cdm::ContentDecryptionModule_11* cdm11_{nullptr};
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.h
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.h
@@ -9,19 +9,20 @@
 #ifndef MEDIA_CDM_CDM_ADAPTER_H_
 #define MEDIA_CDM_CDM_ADAPTER_H_
 
-#include <string>
-#include <vector>
+#include "../../base/compiler_specific.h"
+#include "../../base/macros.h"
+#include "../../base/native_library.h"
+#include "../base/cdm_config.h"
+#include "api/content_decryption_module.h"
+
+#include <atomic>
+#include <future>
 #include <inttypes.h>
 #include <memory>
 #include <mutex>
-#include <atomic>
-#include <future>
-
-#include "../../base/native_library.h"
-#include "../../base/compiler_specific.h"
-#include "../../base/macros.h"
-#include "api/content_decryption_module.h"
-#include "../base/cdm_config.h"
+#include <string>
+#include <unordered_map>
+#include <vector>
 
 namespace media {
 
@@ -221,11 +222,17 @@ class CdmAdapter : public std::enable_shared_from_this<CdmAdapter>,
   void OnInitialized(bool success) override;
 
 
+  std::future<std::string> PrepareSessionFuture(uint32_t promiseId);
+
   //Misc
   ~CdmAdapter();
   bool LoadCDM();
   bool Initialize();
   std::string GetVersion() const;
+
+protected:
+  std::mutex m_sessionMx;
+  std::unordered_map<uint32_t, std::promise<std::string>> m_sessionPromises;
 
 private:
   void UnloadCDM();
--- a/lib/cdm/cdm/media/cdm/cdm_adapter.h
+++ b/lib/cdm/cdm/media/cdm/cdm_adapter.h
@@ -122,7 +122,8 @@ class CdmAdapter : public std::enable_shared_from_this<CdmAdapter>,
 	void TimerExpired(void* context);
 
 	cdm::Status Decrypt(const cdm::InputBuffer_2& encrypted_buffer,
-		cdm::DecryptedBlock* decrypted_buffer);
+                      cdm::DecryptedBlock* decrypted_buffer,
+                      cdm::StreamType streamType = cdm::StreamType::kStreamTypeVideo);
 
   cdm::Status InitializeAudioDecoder(
 		const cdm::AudioDecoderConfig_2& audio_decoder_config);
--- a/src/decrypters/widevine/WVCdmAdapter.cpp
+++ b/src/decrypters/widevine/WVCdmAdapter.cpp
@@ -88,7 +88,9 @@ SResult CWVCdmAdapter::Initialize(const DRM::Config& config, CWVDecrypter* host)
     return SResultCode::ERROR;
   }
   std::string cdmPath;
+  bool useHwSecureCodecs = false;
 #ifdef TARGET_WEBOS
+  useHwSecureCodecs = true;
   auto it = std::ranges::find_if(candidatePaths, [](std::string_view sv)
                                  { return std::filesystem::exists(std::filesystem::path{sv}); });
 
@@ -118,10 +120,10 @@ SResult CWVCdmAdapter::Initialize(const DRM::Config& config, CWVDecrypter* host)
   basePath = FILESYS::PathCombine(basePath, DRM::GenerateUrlDomainHash(licUrl));
   basePath += FILESYS::SEPARATOR;
 
-  auto cdmAdapter =
-      std::make_shared<media::CdmAdapter>("com.widevine.alpha", cdmPath, basePath,
-                                          media::CdmConfig(false, m_config.isPersistentStorage),
-                                          dynamic_cast<media::CdmAdapterClient*>(this));
+  auto cdmAdapter = std::make_shared<media::CdmAdapter>(
+      "com.widevine.alpha", cdmPath, basePath,
+      media::CdmConfig(false, m_config.isPersistentStorage, useHwSecureCodecs),
+      dynamic_cast<media::CdmAdapterClient*>(this));
 
   if (!cdmAdapter->LoadCDM())
   {
--- a/src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
+++ b/src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
@@ -31,6 +31,10 @@ using namespace UTILS;
 
 namespace
 {
+#ifdef TARGET_WEBOS
+constexpr size_t DECRYPTION_SVP_MARGIN = 64;
+#endif
+
 cdm::StreamType ToCdmStreamType(DRM::DRMMediaType stream_type)
 {
   switch (stream_type)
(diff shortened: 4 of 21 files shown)


lib/cdm/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -10,6 +10,8 @@
 
 #include "../../debug.h"
 #include "../base/limits.h"
+#include "SrvBroker.h"
+#include "CompSettings.h"
```
CastagnaIT: just a comment for now,
cdm library (or any other library) must not be coupled with ISA components so keep them stand-alone
what i mean is find another way to avoid use service broker here

you have made many changes will take some days to review
let me know when PR is ready

cscd98: Resolved that issue. Is it ok to make use of cdm::streamType and #include "cdm/media/cdm/api/content_decryption_module.h" in src/common/AdaptiveCencSampleDecrypter.h etc?

CastagnaIT: definitely a bad thing you are coupling many components to CDM library
where the intention is keep all decoupled, widevine (and the CDM form) is only one of many DRM's
to consider that cdm library cannot be builded on all OS's

I see many changes that carry streamType/mediatype, where i in the past for cleaning purposes i removed
I don't have the full picture right now because i have to download your branch and take a depth look,
at first glance, it seems that in some places the code maybe could be optimized to omit carry streamType/mediatype

I would like also to remind you that the use of defines should only be used when excusively necessary

cscd98: I've decoupled some of the CDM stuff and used DRM media type instead. I agree completely about passing around streamType, but at the time it was the only solution I could come up with as I needed to pass this all the way to the CDM.

I have removed a couple of #ifdefs, particularly around secure decode option, as it only shows on webOS anyway. It's possible to remove the L3 fallback option.

cscd98:
@CastagnaIT I've made some changes if you can take a look

lib/cdm/cdm/media/cdm/cdm_adapter.h
```diff
@@ -262,6 +270,10 @@ class CdmAdapter : public std::enable_shared_from_this<CdmAdapter>,
 
   cdm::Buffer* active_buffer_{nullptr};
 
+  std::promise<void> initPromise;
+  std::future<void> initFuture = initPromise.get_future();
+  std::atomic<bool> provisioningCompleteOrStarted{false};
```
CastagnaIT: to members use prefix `m_`

lib/cdm/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -61,6 +61,8 @@ const char PATH_SEPARATOR = '\\';
 const char PATH_SEPARATOR = '/';
 #endif
 
+constexpr unsigned int INIT_FUTURE_TIMEOUT_SEC = 3;
```
CastagnaIT: why not avoid conversions with
`constexpr std::chrono::seconds INIT_FUTURE_TIMEOUT{3};`

lib/cdm/cdm/media/cdm/cdm_adapter.cc
```diff
@@ -251,18 +253,32 @@ bool CdmAdapter::Initialize()
...
     else if (cdm10_)
     {
       cdm10_->Initialize(cdm_config_.allow_distinctive_identifier,
-                         cdm_config_.allow_persistent_state, false);
+                         cdm_config_.allow_persistent_state, cdm_config_.use_hw_secure_codecs);
+    }
+
+    // Wait for the CDM to be initialized
+    // Add a maximum timeout in case we never hear back!
+    if (initFuture.wait_for(std::chrono::seconds(INIT_FUTURE_TIMEOUT_SEC)) != std::future_status::ready) {
```
CastagnaIT: if for any x reason the cdm will be initializated again, a new call to `wait_for` will cause an exception due to bad future state, and so kodi will crash
the consumend promise should be re-set in some way

src/Session.cpp
```diff
@@ -684,6 +684,14 @@ bool SESSION::CSession::PrepareStream(CStream* stream, uint64_t startPts)
...
   }
 
+#ifdef TARGET_WEBOS
+  if (adp->GetStreamType() == StreamType::AUDIO)
+  {
+    m_reprChooser->SetSecureSession(session && session->capabilities.flags &
+                                                   DRM::DecrypterCapabilites::SSD_SECURE_PATH);
+  }
+#endif
+
```
CastagnaIT: this can be fully removed, maybe its not so clear but the only reason that "representation choosers" need to know this is for video purpose not audio,
this is related to setting "maximum resolution for drm videos" its a workaround for RPI & similars, low-powered CPUs cannot process correctly high resolutions (stuttering)

i would like in a future full remove this workaround setting, but idk...

src/decrypters/widevine/WVCdmAdapter.cpp
```diff
@@ -88,7 +88,9 @@ SResult CWVCdmAdapter::Initialize(const DRM::Config& config, CWVDecrypter* host)
     return SResultCode::ERROR;
   }
   std::string cdmPath;
+  bool use_hw_secure_codecs = false;
```
CastagnaIT: please avoid special underscore chars on new variables

lib/cdm/cdm/media/cdm/cdm_adapter.cc
CastagnaIT: also please apply clang formatting

src/Session.cpp
```diff
@@ -191,7 +191,7 @@ bool SESSION::CSession::CheckPlayableStreams(PLAYLIST::CPeriod* period)
                 //! the task to initialize DRM session only to the requested streams.
                 //! It can be tested e.g. with some Am@zon videos
                 auto& caps = session->capabilities;
-
+#ifndef TARGET_WEBOS
```
CastagnaIT: you can move this define into the appropriate `IsSecureDecoderAudioSupported` method
here there are a lot of workaround due to this missing secure audio decoder feature

CastagnaIT (commented):
i placed some comments, but i need much more time to review all widevine cdm changes

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
```diff
@@ -71,14 +71,25 @@ CWVCencSingleSampleDecrypter::CWVCencSingleSampleDecrypter(
...
-      m_promiseId++, cdm::SessionType::kTemporary, cdm::InitDataType::kCenc, m_pssh.data(),
-      static_cast<uint32_t>(m_pssh.size()));
+  const uint32_t pid = m_promiseId++;
 
-  //! @todo: loop with thread sleep should be removed, use callbacks
-  int retrycount = 0;
-  while (m_strSession.empty() && ++retrycount < 100)
-    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  // prepare the promise before calling into CDM
+  std::future<std::string> fut = m_cdmAdapter->PrepareSessionFuture(pid);
```
CastagnaIT: a better var naming example
```suggestion
  std::future<std::string> sessionFut = m_cdmAdapter->PrepareSessionFuture(pid);
```

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
```diff
@@ -903,3 +1033,16 @@ void CWVCencSingleSampleDecrypter::AddKeyId(const std::vector<uint8_t>& keyId)
     m_keys.push_back(key);
   }
 }
+
+cdm::StreamType CWVCencSingleSampleDecrypter::ToCdmStreamType(DRM::DRMMediaType stream_type)
```
CastagnaIT: I don't think this utility needs to be a member of the class
please move on top in an unnamed namespace

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
```diff
@@ -445,7 +469,8 @@ void CWVCencSingleSampleDecrypter::RepackSubsampleData(AP4_DataBuffer& dataIn,
                                                        size_t& cipherPos,
                                                        const size_t subsamplePos,
                                                        const AP4_UI16* bytesOfCleartextData,
-                                                       const AP4_UI32* bytesOfEncryptedData)
+                                                       const AP4_UI32* bytesOfEncryptedData,
+                                                       DRM::DRMMediaType streamType)
```
CastagnaIT: streamType param has been added but unused, suggest you can remove it

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
```diff
@@ -514,123 +540,99 @@ AP4_Result CWVCencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 poolId,
...
-        if (iv)
-        {
-          // Update the byte containing the data size of current subsample referred to clear bytes array
-          AP4_UI16* clrb_out = reinterpret_cast<AP4_UI16*>(dataOut.UseData() + clrDataBytePos);
-          *clrb_out += (4 - fragInfo.m_nalLengthSize);
-        }
-        
-        // ++nalUnitCount;
+    const size_t header_cushion = 64;
+    m_decryptOut.Reserve(payload.GetDataSize() + dataOut.GetDataSize() + header_cushion);
```
CastagnaIT: have you some info about this `header_cushion`, cushion?
if so, please choose a better var name,
and move this inside an unnamed namespace on top by using constexpr, or otherwise at least should be `static const`

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
```diff
@@ -755,6 +773,134 @@ AP4_Result CWVCencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 poolId,
...
+    for (unsigned i = 0; i < subsampleCount; ++i)
+    {
+      cdm::SubsampleEntry e;
+      e.clear_bytes = bytesOfCleartextData[i];
+      e.cipher_bytes = bytesOfEncryptedData[i];
+      rebuiltSubs.push_back(e);
+    }
+  }
+
+  unsigned curSub = 0;
```
CastagnaIT:
```suggestion
  size_t curSub = 0;
```

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
```diff
@@ -755,6 +773,134 @@ AP4_Result CWVCencSingleSampleDecrypter::DecryptSampleData(AP4_UI32 poolId,
...
+
+    if (!iv) {
+      nalUnitSum = 0;
+    }
+    else if (nalSize + fragInfo.m_nalLengthSize + nalUnitSum >=
+             *bytesOfCleartextData + *bytesOfEncryptedData)
+    {
+      // We may cross one or more subsample boundaries; advance pointers and curSub in lockstep
+      AP4_UI32 summedBytes = 0;
+      unsigned advanced = 0;
```
CastagnaIT:
```suggestion
      size_t advanced = 0;
```

CastagnaIT (commented):
i will stop here for today, i think cdm/CWVCencSingleSampleDecrypter changes should be ok but i need at least do some tests

CastagnaIT:
just to know have you tested only on webos or also on other type of OS?

src/decrypters/widevine/WVCencSingleSampleDecrypter.cpp
cscd98: The CDM will add svp headers and meta data, so we set a safety margin

cscd98:
Just webOS. My intention was to make only changes for that but my changes annex b function may affect other platforms if they use secure path (not android though as it has a different class), but the callbacks/futures and dvcc parsing benefit other platforms.

src/CompSettings.cpp
```diff
@@ -87,6 +87,11 @@ uint32_t ADP::SETTINGS::CCompSettings::GetBandwidthMax() const
 
 bool ADP::SETTINGS::CCompSettings::IsIgnoreScreenRes() const
 {
+  // this setting has no effect on webOS as 1920x1080 is always reported
+  // return always true as this prevents 4k playback otherwise
+#ifdef TARGET_WEBOS
```
CastagnaIT: if kodi cant get 4k means that webos interface is capped to 1080p resolution (such as some android tv)
so if you play 4k will result that the images will be downscaled to 1080p and re-upscaled to 4k by display, not so ideal but better than nothing.

But remember: if you force this permanently you assume that all LG tv's are 4k(?)
(or if you play a 16k stream you assume that all TV are 16k capable, and so on...)
this way 1080p screens will always play 4K/16k/... even if they are not supported by display,
this will affect network capacity (bandwidth) and also playback performances

I suggest you consider whether it is right to force this on all TV models,
or if there may be better alternatives

cscd98: From what I understand the GUI is 1080p but then 4k is played natively at that resolution. Or are you saying that IA/Kodi is rendering at 1080p because of this and the TV is now upscaling?

I was also under the impression that IA only switches to higher quality streams when it can, which is why it starts in 720 and then goes to 4k on D+ after 30 seconds?

We only support 2018+ TVs which ASFAIK are all 4k *some being 8k*

sundermann: The gui layer (wayland) is 1080p / 720p. Almost all TVs are 4k with some exceptions being 1080p. There is a hardware compositor which merges GUI and video planes together so videos are played back in native resolution and the GUI is upscaled.

CastagnaIT: > Or are you saying that IA/Kodi is rendering at 1080p because of this and the TV is now upscaling?

yes due to the OS (assuming you dont have something wrong configured on VP) kodi is capped to render at max to 1080p, and the tv display do a second pass to upscale it to 4k

otherwise kodi should provide correctly max resolution:
https://github.com/xbmc/inputstream.adaptive/blob/f814b0df9cab5de01f01b742f43813f1878e4256/src/main.cpp#L411-L414
(max resolution == max supported resolution of tv display)
then when you set kodi setting `adjust display refresh rate` enabled, the stream start directly at 4k and then switch the display resolution to 4k

sundermann: No, not on webOS. The GUI is (usually) fixed at 1080p@60Hz. `adjust display refresh rate` is non functional because there is only one available resolution. Kodi only knows about the resolution of the GUI layer but on webOS the video layer is separate from that and can go beyond the GUI resolution.

CastagnaIT: then this is the root of problem that should be fixed
because what you have done here is a workaround not so good solution,
"adaptive streaming" need to know the max resolution in order to work correctly

sundermann: It's not easy to fix. There is no separation between GUI and video resolution in Kodi. The GUI can't use 4k and the video is not rendered on a wayland plane. The GUI does not know about the video plane at all. That's why we cannot easily add video resolutions to the wayland windowing. It would just break the GUI.

CastagnaIT: there is a way to query webos to know the max supported resolution?

CastagnaIT: because if so maybe you can find a way to override this
https://github.com/xbmc/xbmc/blob/e49295490c2e46b9d0aff28bc9f8dbe0c7e347aa/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp#L804-L818

cscd98: I will ask my resident webOS experts. But, I can remove this hard coded setting. The user just needs to switch to 'on' otherwise adaptive stream will only goto 1080p (just confirmed this again for clarity).

When off -> 720p -> becomes 1080p video stream
When on -> 720p -> becomes 4k video stream

sundermann: Yes there is `smp::util::getMaxVideoResolution` but it would require pulling in some additional dependencies. As a rule of thumb 1080p GUI -> 4k video, 720p GUI -> 1080p video. But the video decoder usually can go a bit more than just the native resolution.

src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -261,6 +266,12 @@ bool CFragmentedSampleReader::GetInformation(kodi::addon::InputstreamInfo& info)
 
   isChanged |= m_codecHandler->GetInformation(info);
 
+  // store dvcc metadata
+  if (auto dvcc = ParseDvcc())
+  {
+    PopulateDvccMetadata(info, dvcc);
+  }
```
CastagnaIT: when an info will be changed `isChanged` must be set to true, otherwise there is the risk that these new info will be not updated on kodi side

if you dont want to compare each DV info at least check if previously this info was missing, if so, and if you have this info to set, set true isChanged

CastagnaIT: another thing, has been created each Codechandler implementations to organize the code by codec,
instead you have placed dvcc parsing/set globally on the samplereader

more likely a better place is into `CodecHandler.h`
then populate this from the appropriate codechandler's GetInformation methods

src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -583,3 +594,61 @@ void CFragmentedSampleReader::ParseTrafTfrf(AP4_UuidAtom* uuidAtom)
...
+  meta->m_dvVersionMajor = dvcc->GetDvVersionMajor();
+  meta->m_dvVersionMinor = dvcc->GetDvVersionMinor();
+  meta->m_dvProfile = dvcc->GetDvProfile();
+  meta->m_dvLevel = dvcc->GetDvLevel();
+  meta->m_rpuPresentFlag = dvcc->GetRpuPresentFlag();
+  meta->m_elPresentFlag = dvcc->GetElPresentFlag();
+  meta->m_blPresentFlag = dvcc->GetBlPresentFlag();
+  meta->m_dvBlSignalCompatibilityId = dvcc->GetDvBlSignalCompatibilityID();
+
+  info.SetDvccMetadata(meta);
```
CastagnaIT: possible memleak and more likely this should be implemented differently not as raw pointer
see comment https://github.com/xbmc/xbmc/pull/27136#discussion_r2300945274

src/samplereader/FragmentedSampleReader.cpp
cscd98: I have added isChanged. I did originally put this in CodecHandler, but some for some reason I could not get the dvcc info from there and ended up parsing m_track.

src/CompSettings.cpp
cscd98: So do we need to override https://github.com/xbmc/xbmc/blob/e49295490c2e46b9d0aff28bc9f8dbe0c7e347aa/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp#L804-L818 or this https://github.com/xbmc/inputstream.adaptive/blob/f814b0df9cab5de01f01b742f43813f1878e4256/src/main.cpp#L411-L414 ?

The former is easier, as it already has some of the webOS APIs available (not necessarily the one we need but some)

CastagnaIT: yes i mean in the https://github.com/xbmc/xbmc/blob/e49295490c2e46b9d0aff28bc9f8dbe0c7e347aa/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp#L804-L818

but now that i m thinking again, you need to set the max video resolution to both pair of resolution (videoXY and maxXY) of `SetVideoResolution(videoWidth, videoHeight, maxWidth, maxHeight);` call
this because since  `adjust display refresh rate` is non functional, i assume its always disabled (and/or hiddened)
since ISA take in account also of this kodi setting, if you dont set (in this case) 4k to "current resolution" will have no effect when ISA perform the stream res selection

also i suggest you to add a code comment by explaining the underlying reason about this "override" on kodi side
something like "webOS have fixed GUI layer resolution, decoupled from video resolution that can become higher, and to allow IS addons to provide high resolutions streams you need to provide the appropriate max display resolution"
or a similar explanation
so as to explain to those who will get involved in future changes

src/samplereader/FragmentedSampleReader.cpp
CastagnaIT: you can pass m_track to codechandler constructor (to store var also on it) imo no problem
codechandlers live inside CFragmentedSampleReader and m_track dont change during his lifetime

src/Session.cpp
```diff
@@ -191,7 +191,6 @@ bool SESSION::CSession::CheckPlayableStreams(PLAYLIST::CPeriod* period)
                 //! the task to initialize DRM session only to the requested streams.
                 //! It can be tested e.g. with some Am@zon videos
                 auto& caps = session->capabilities;
-
```
CastagnaIT, line 194: please restore blank line otherwise generate unneeded code change history,
also below

src/CompSettings.cpp
cscd98: As @sundermann mentioned we can use smp::util::getMaxVideoResolution but this requires the codec name being passed to this function.

DetectScreenResolution() - do we know what codec is in use by this time? (as determined by the stream choice?)

There is some other APIs but they are not readily available and will take some more reverse engineering. e.g. https://webostv.developer.lge.com/develop/references/webostvjs-webos#deviceinfo

CastagnaIT: >  do we know what codec is in use by this time?

no because SetVideoResolution callback is not directly related to the playback
(to take in account that some type of live streaming can use different video codec between "chapters")
sound weird that to knowing the max res of display must be involved the codec

CastagnaIT: I don't understand why the max display resolution should change by codec. perhaps they limit the display resolution per codec due to poor CPU performances?

sundermann: It's the maximum decodable resolution. Sometimes you find 4k content with a resolution of 4096x2160. That is beyond the normal screen resolution of 3840x2160 but still in that case would prefer using 4096x2160 instead of falling back to 1080p.

That said I don't think it's worth much implementing it properly. Just go for 720p -> return 1080p max resolution for video and 1080p -> 4k video. 

`adjust display refresh rate` is hidden by default and even if enabled doesn't have any effect as there is only one resolution there is simply nothing to adjust to.

CastagnaIT: As you prefer, you know the best ways, i can't figure out what's is possible todo on webOS,
anyway imo this override should be done on kodi interface
and not by adding a workaround on each individual video binary add-on

cscd98: Well if I do this:

```
    <snip>
    CResolutionUtils::GetMaxAllowedScreenResolution(maxWidth, maxHeight);

#ifdef TARGET_WEBOS
    if(videoWidth >= 1920 || videoHeight >= 1080)
    {
      maxWidth = 3840;
      maxHeight = 2160;
    }
#endif
```

SetVideoResolution is called once, setting it to: 1920x1080, 3840x2160.
m_player->GetVideoResolution always returns to 1920x1080, so this code never fires again.

It will now go straight to a 4K UHD stream instead of going 720p->4k delay.

Is this now correctly rendering 4k? and not some butchered 4k->1080p->4k? Is there a more friendly less #ifdef way of doing so? e.g. GetMaxAllowedScreenResolution or GetVideoResolution

sundermann: Just add an override like this in Kodi:

```c++
void CWinSystemWaylandWebOS::UpdateResolutions()
{
  CWinSystemWayland::UpdateResolutions();

  RESOLUTION_INFO& res = CDisplaySettings::GetInstance().GetResolutionInfo(RES_DESKTOP);
  if (res.iWidth == 1920 && res.iHeight == 1080)
  {
    res.iScreenHeight = 2160;
    res.iScreenWidth = 3840;
  }
  else if (res.iWidth == 1280 && res.iHeight == 720)
  {
    res.iScreenHeight = 1920;
    res.iScreenWidth = 1080;
  }
}
```

After checking, Kodi differentiates between gui size and screen size. So doing this override should be safe.

CastagnaIT: > SetVideoResolution is called once, setting it to: 1920x1080, 3840x2160.
> m_player->GetVideoResolution always returns to 1920x1080, so this code never fires again.

as said above you have to make sure that "overrided" res is set on both parameters pair, not only on the "max" pair
PS. Maybe i misunderstood the sentence, but i think you understood

cscd98: @sundermann That is what I had in mind but I thought it might break something. But appears not.

xbmc/windowing/wayland/WinSystemWaylandWebOS.cpp:
```2025-08-29 08:50:32.306 T:4341     info <general>: virtual void KODI::WINDOWING::WAYLAND::CWinSystemWayland::UpdateResolutions(): - 1920x1080 @60.000 Hz pixel ratio 1.000 current
2025-08-29 08:50:32.306 T:4341    debug <general>: virtual void KODI::WINDOWING::WAYLAND::CWinSystemWaylandWebOS::UpdateResolutions(): UpdateResolutions setting screen resolution to 3840x2160
```

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp:
```
2025-08-29 08:50:55.161 T:4494    debug <general>: CInputStreamAddon::DetectScreenResolution - Detect screen resolution
2025-08-29 08:50:55.161 T:4494    debug <general>: CInputStreamAddon::DetectScreenResolution - Detected screen resolution 1920x1080
2025-08-29 08:50:55.161 T:4494    debug <general>: CInputStreamAddon::DetectScreenResolution - Set video resolution to 1920x1080, max allowed 3840x2160
2025-08-29 08:50:55.162 T:4494    debug <general>: CInputStreamAddon::DetectScreenResolution - Set video resolution to 1920x1080, max allowed 3840x2160 - done
```

However, repeated logging in xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp shows:
```
2025-08-29 08:51:16.187 T:4494    debug <general>: CInputStreamAddon::DetectScreenResolution - Detect screen resolution
2025-08-29 08:51:16.187 T:4494    debug <general>: CInputStreamAddon::DetectScreenResolution - Detected screen resolution 1920x1080
```

However the video output stream is 4k straight away with no 720p->4k phase.

So max is set to 3840x2160, but video resolution is still reported as 1920x1080.

cscd98: I think we have to override as well:

```
void CVideoPlayer::GetVideoResolution(unsigned int &width, unsigned int &height)
{
  CLog::LogF(LOGDEBUG, "CVideoPlayer::{} - returning hardcoded 3840x2160");
  RESOLUTION_INFO res = CServiceBroker::GetWinSystem()->GetGfxContext().GetResInfo();
  width = 3840;
  height = 2160;
  //width = res.iWidth;
  //height = res.iHeight;
}
```

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp:
```
2025-08-29 09:12:05.752 T:7868    debug <general>: CInputStreamAddon::DetectScreenResolution - Detect screen resolution
2025-08-29 09:12:05.752 T:7868    debug <general>: CInputStreamAddon::DetectScreenResolution - Detected screen resolution 3840x2160
2025-08-29 09:12:05.752 T:7868    debug <general>: CInputStreamAddon::DetectScreenResolution - Set video resolution to 3840x2160, max allowed 3840x2160
2025-08-29 09:12:05.752 T:7868    debug <general>: CInputStreamAddon::DetectScreenResolution - Set video resolution to 3840x2160, max allowed 3840x2160 - done
```

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp:
```
2025-08-29 09:12:12.515 T:7868    debug <general>: CInputStreamAddon::DetectScreenResolution - Detect screen resolution
2025-08-29 09:12:12.515 T:7868    debug <general>: CInputStreamAddon::DetectScreenResolution - Detected screen resolution 3840x2160
```

Seems to work fine for me playback of a 4K and 1080p stream

sundermann: You can override `CVideoPlayer::GetVideoResolution` but when you do retrieve the max screen size from the displayinfo and only when we are using the webOS decoder (m_mediaPipeline is not null).

cscd98: I tried this but its not picking up my overrode function:

```
class CVideoPlayerVideoWebOS final : public IDVDStreamPlayerVideo, IVideoPlayer
{
public:
  void GetVideoResolution(unsigned int &width, unsigned int &height) override;
```

```
void CVideoPlayerVideoWebOS::GetVideoResolution(unsigned int &width, unsigned int &height)
{
  CLog::Log(LOGDEBUG, "CVideoPlayerVideoWebOS");

  RESOLUTION_INFO res = CServiceBroker::GetWinSystem()->GetGfxContext().GetResInfo();

  if(width < MAX_VIDEO_WIDTH && height < MAX_VIDEO_HEIGHT)
  {
    width = MAX_VIDEO_WIDTH;
    height = MAX_VIDEO_HEIGHT;
    CLog::Log(LOGDEBUG, "CVideoPlayerVideoWebOS::{} - Adjusting resolution to {}x{}", __FUNCTION__,
              width, height);

    return;
  }

  width = res.iWidth;
  height = res.iHeight;
}
```

src/samplereader/FragmentedSampleReader.h
```diff
@@ -54,7 +54,6 @@ class ATTR_DLL_LOCAL CFragmentedSampleReader : public ISampleReader, public AP4_
 private:
   void UpdateSampleDescription();
   void ParseTrafTfrf(AP4_UuidAtom* uuidAtom);
-
```
CastagnaIT, line 57: just cleanup this, please restore empty line

CastagnaIT (approved):
i think its ok,
i tested under linux some ISASamples but atm i have no streams that use secure decoder to test so hope it also its ok

PS. for next pr's it would be better for each commit to have the prefix of the component it modifies (or appropriate context)

cscd98:
I've moved the dvcc metadata parsing to a separate PR - https://github.com/xbmc/inputstream.adaptive/pull/1920 as that PR is waiting for upstream xbmc changes to be merged which is not necessary for the rest of this PR to be merged.

src/decrypters/widevineandroid/WVDecrypter.h
```diff
@@ -55,7 +55,8 @@ class ATTR_DLL_LOCAL CWVDecrypterA : public DRM::IDecrypter
 
   virtual void GetCapabilities(std::shared_ptr<Adaptive_CencSingleSampleDecrypter> decrypter,
                                const std::vector<uint8_t>& keyId,
-                               DRM::DecrypterCapabilites& caps) override;
+                               DRM::DecrypterCapabilites& caps,
+                               DRMMediaType mediaType) override;
```
CastagnaIT: android compilers fails
this due to inappropriate/missing changes, this should be:

```suggestion
                               DRM::DRMMediaType mediaType) override;
```

and also its missing relative method code change on .cpp implementation

CastagnaIT: after fix this, let me know if you want to merge PR now or wait

cscd98:
@CastagnaIT Changes pushed. If your happy I'd like to get this merged in

CastagnaIT:
there are some other missing or wrong changes on android side take a look on jenkins build failures

cscd98:
@CastagnaIT OK now it should be working!

src/decrypters/HelperWv.h
```diff
@@ -78,6 +79,8 @@ class ATTR_DLL_LOCAL IWVCdmAdapter : public IWVSubject
   //! @todo: add here this method for convenience needed investigate better to better cleanup,
   //! also Load/Save certificate methods need a full code cleanup
   virtual void SaveServiceCertificate() {}
+
+  virtual std::future<std::string> PrepareSessionFuture(uint32_t promiseId) = 0;
```
CastagnaIT: i was taking a look at previous jenkins build errors about this method add to interface,
imo this way to use PrepareSessionFuture from here add a not needed round of "foward" implementation that can be removed

since `PrepareSessionFuture` is specific to widevine CDM should not involve other interfaces
so imo, you can call the CDM interface directly as already done on code of `CWVCencSingleSampleDecrypter`

so on `CWVCencSingleSampleDecrypter::CWVCencSingleSampleDecrypter` method
you can replace:
`std::future<std::string> sessionFut = m_cdmAdapter->PrepareSessionFuture(pid);`
with:
`std::future<std::string> sessionFut = m_cdmAdapter->GetCDM()->PrepareSessionFuture(pid);`
then you can remove `IWVCdmAdapter PrepareSessionFuture` overridable method
this will allow to keep cleanup android interface without having to add an unused implementation of PrepareSessionFuture method

cscd98:
@CastagnaIT updated

src/decrypters/widevine/WVCdmAdapter.cpp
```diff
@@ -231,3 +233,9 @@ void CWVCdmAdapter::NotifyObservers(const CdmMessage& message)
       observer->OnNotify(message);
   }
 }
+
+std::future<std::string> CWVCdmAdapter::PrepareSessionFuture(uint32_t promiseId)
+{
+  // Forward to the actual implementation
+  return m_cdmAdapter->PrepareSessionFuture(promiseId);
+}
```
CastagnaIT: why not removed? should be not used now or not?

src/decrypters/widevine/WVCdmAdapter.h
```diff
@@ -56,6 +56,8 @@ class ATTR_DLL_LOCAL CWVCdmAdapter : public media::CdmAdapterClient,
   void DetachObserver(IWVObserver* observer) override;
   void NotifyObservers(const CdmMessage& message) override;
 
+  std::future<std::string> PrepareSessionFuture(uint32_t promiseId);
```
CastagnaIT: same as above

src/decrypters/widevineandroid/WVCdmAdapter.h
```diff
@@ -23,6 +23,7 @@
 
 #include <list>
 #include <mutex>
+#include <future>
```
CastagnaIT: should no longer be necessary

cscd98: removed

src/CompSettings.cpp
```diff
@@ -87,6 +87,11 @@ uint32_t ADP::SETTINGS::CCompSettings::GetBandwidthMax() const
...
+  return true;
+#endif
```
CastagnaIT: last thing,
this problem has not been resolved by the PR https://github.com/xbmc/xbmc/pull/27198 ?

cscd98: Doesn't seem to be required when I tested it just, so removed it.

CastagnaIT:
Thank you for your patience in reaching the end :)

PR #1947, opened 2025-11-10
[FragmentedSampleReader] Fix switching between unencrypted/crypted content

### LINKED ISSUE #1865: [Bug] Netflix H264 not starting properly starting with ISA Omega V 21.5.10
### Describe the problem

Starting with ISA .10 Netflix h.264 Streams doesn't start properly. switching back to .9 is working like expected. Workaround ist to switch the codec to VP9, then there is no problem or switch to Fullscreen/Video and jump 10 Minutes then it is likly to get picture with sound. Language switching sometimes also start the picture. 

Amazon streams doesn't seems to be effected. But due clear reproducing the error, i think it is more an ISA issue then a Netflix-Addon error. 

My System LE 12/ Raspberry Pi 4 

### Possible fix

i don't know, i had also a look into the changes, but don't know to much about wv. 

### Steps to reproduce

Using Omega ISA >= .10
Start netflix stream with h.264. 
System is doing everything like display switch to suitable frequency and spinner is showing. 
Stream is not starting visibly. 
press return, spinner goes away, switch to video/fullscreen
see wild osd, 
jump 10 minutes + 

### Debug log

had no time, maybe not needed because of the small changes, but will deliver one, if necessary

### Stream manifest file(s)

same here

### Additional info

_No response_

### Operating system(s)

Linux

LE 12.0.2

>= 21.5.10

### Kodi version(s)

Omega

CastagnaIT:
## Description

followup of PR #1916

Here is a partial reworking of how `CFragmentedSampleReader::ReadSample` chooses which decryptor to use
still not so clear to me the reason why there was so messy variables in that method before,
perhaps some manifests use cases that i am unaware of and that should be documented,
or past attempts to fix other issues that are now unnecessary...who knowns

at least for n€tflix use case, the stream start with unencrypted packets, and then switch to encrypted content
so this lead to initially use `useDecryptingDecoder` condition for unencrypted content and then use `m_decrypter` condition for encrypted content,
so in any case also the unencrypted packets need to be passed to the WV decrypter

## Motivation and context

final fix #1865
## How has this been tested?

user of issue confirm to works

i also tested another similar (but not equal) stream use case from ISA Samples:
`Dash -> Google Tears [CBCS]`

## Types of change

## Checklist:

### DIFF
--- a/src/decrypters/widevine/CdmTypeConversion.cpp
+++ b/src/decrypters/widevine/CdmTypeConversion.cpp
@@ -203,37 +203,44 @@ void media::ToCdmInputBuffer(const DEMUX_PACKET* encryptedBuffer,
   inputBuffer->data = encryptedBuffer->pData;
   inputBuffer->data_size = encryptedBuffer->iSize;
   inputBuffer->timestamp = static_cast<int64_t>(encryptedBuffer->pts * 1000000 / STREAM_TIME_BASE);
-  inputBuffer->key_id = encryptedBuffer->cryptoInfo->kid;
-  inputBuffer->key_id_size = 16;
-  inputBuffer->iv = encryptedBuffer->cryptoInfo->iv;
-  inputBuffer->iv_size = 16;
-
-  const uint16_t numSubsamples =
-      encryptedBuffer->cryptoInfo ? encryptedBuffer->cryptoInfo->numSubSamples : 0;
-  if (numSubsamples > 0)
-  {
-    subsamples->reserve(numSubsamples);
-    for (uint16_t i = 0; i < numSubsamples; i++)
-    {
-      subsamples->push_back({encryptedBuffer->cryptoInfo->clearBytes[i],
-                             encryptedBuffer->cryptoInfo->cipherBytes[i]});
-    }
-  }
-
-  inputBuffer->subsamples = subsamples->data();
-  inputBuffer->num_subsamples = numSubsamples;
 
   if (encryptedBuffer->cryptoInfo)
   {
+    DEMUX_CRYPTO_INFO* cryptoInfo = encryptedBuffer->cryptoInfo;
+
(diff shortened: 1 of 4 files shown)


CastagnaIT:
i ping you here 
@Sandmann79
@matthuisman 

I would like to involve you in testing this change,
since its a critical component i would like to avoid breaking anything

what i need is test am@zon and D+ on platforms:
Linux or Windows, and android (if you can)

let me know if playback still go well (try to play some minutes, and video seek just to be sure)

Test builds here:
https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.adaptive/detail/PR-1947/4/artifacts
(link updated)

CastagnaIT:
@claude please review again

CastagnaIT:
about "Potential Concerns" N°1:
I think it's right as it is, since `IsEncrypted` method create an encrypted demux packet that should be used only with encrypted content for secure decoder, the `useDecryptingDecoder` case (in the `CFragmentedSampleReader::ReadSample`) should not use it
unless i'm missing something...

about "Potential Concerns" N°2:
i will investigate (UPDATE: included fix)

about "Potential Concerns" N°3:
should not cause problems since in that line its used only to get the KID,
sure in future more likely should be improved since the KID potentially could change between packets

the-dreamer:
> what i need is test am@zon and D+ on platforms: Linux or Windows, and android (if you can)


I can confirm at least disney+ ist still working. Amazon i had not tested recently. (i am talking about the test patch not this specific patch. )

Sandmann79:
Checked it on Windows and Android (aarch64) with Amazon and couldn't find any errors.
Playback / seekíng / multiple video starts in quick succession – worked as expected.

Testet with Commit 799794d

CastagnaIT:
thanks so good to go

PR #1959, opened 2025-12-04
[AdaptiveStream] Fix/simplify obtaining segment from PTS

CastagnaIT:
## Description

From recent PR #1955 i removed code commented as "Assume that we have I-Frames only at segment start"
commit https://github.com/xbmc/inputstream.adaptive/commit/94f6e40c8cb720f9d918084c9ab3771b77df0d5f

but i notices a crash on video seek from some streams
for example following HLS AVC https://www.dailymotion.com/video/x9u25zs 

seem that since i removed the above linked code on AdaptiveStream::seek_time
now select a segment where the PTS is not in the range (start/end PTS) of the segment
this PTS discrepancy seems to cause Bento4 to crash

the crash come from a call of `SeekSample` from `CFragmentedSampleReader::TimeSeek`
(seeksample method is a our custom implementation so i wonder if there is something wrong on it)
https://github.com/xbmc/Bento4/blob/f5eed3d7a4ec39bab8ed43332e58a0e6337fa189/Source/C%2B%2B/Core/Ap4LinearReader.cpp#L688-L692
seem that when the PTS is not in the range, 
last call to `Advance` method in `AP4_LinearReader::SeekSample` cause to delete the SampleTable of the tracker
then when call `SetSampleIndex` crash due to missing SampleTable

my idea is rework all related code to find more precisely the segment
by taking in account start/end PTS of each segment in the timeline

Unfortunately i haven't yet been able to fully understand how that bento4 method works, and atm i can't spend many hours on it
i'm afraid that it might contain old standing bugs

## Motivation and context

while working on WIP PR i found this problem

## How has this been tested?

https://www.dailymotion.com/video/x9u25zs 

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveStream.cpp
+++ b/src/common/AdaptiveStream.cpp
@@ -1168,44 +1168,17 @@ bool adaptive::AdaptiveStream::seek_time(double seek_seconds, bool preceeding, b
 
   std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(m_tree->GetTreeUpdMutex());
 
-  uint64_t sec_in_ts = static_cast<uint64_t>(seek_seconds * current_rep_->GetTimescale());
+  const uint64_t pts = static_cast<uint64_t>(seek_seconds * current_rep_->GetTimescale());
 
-  //Skip initialization
-  size_t choosen_seg{0};
-
-  while (choosen_seg < current_rep_->Timeline().GetSize() &&
-         sec_in_ts > current_rep_->Timeline().Get(choosen_seg)->startPTS_)
-  {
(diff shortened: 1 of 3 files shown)


CastagnaIT:
@claude review again

CastagnaIT:
about Potential Issues:
1) could be an idea
2) we assume that segments in timeline are always sorted by PTS (despite currently there is no check for it)
3) atm im not aware of overlapping segments, in case i think there could be problems also on many more code parts
4) I don't think it's worth it

CastagnaIT:
@claude i changed code to search segment
i think it's better to try to get the next segment than to jump to the last available one, as the previous change did

PR #1990, opened 2026-01-31
Removed last PTS video seek workaround on subtitles

CastagnaIT:
## Description

It was used because Kodi core did not stream seek on subtitles when subtitles were enabled/disabled or when changing tracks (which is essentially similar)

## Motivation and context

has been proposed a fix on kodi core that can cover also this lack
PR: https://github.com/xbmc/xbmc/pull/27770
so this PR depends from it

## How has this been tested?

stream with multiple subtitles (segmented cases):
-> enable / disable subtitles from different point of playback
-> change subtitles during playback

in both cases, subtiltes must result in-sync with video playback

tested also on multi-chapters (multi-periods) stream

## Types of change

## Checklist:

### DIFF
--- a/src/common/AdaptiveUtils.h
+++ b/src/common/AdaptiveUtils.h
@@ -37,9 +37,6 @@ constexpr uint64_t NO_PTS_VALUE = std::numeric_limits<uint64_t>::max();
 // Marker for undefined value
 constexpr uint64_t NO_VALUE = std::numeric_limits<uint64_t>::max();
 
-// Kodi VideoPlayer internal buffer
-constexpr uint64_t KODI_VP_BUFFER_SECS = 8;
-
 enum class EncryptionState
 {
   UNENCRYPTED,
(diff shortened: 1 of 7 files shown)


src/common/AdaptiveUtils.h
```diff
@@ -37,9 +37,6 @@ constexpr uint64_t NO_PTS_VALUE = std::numeric_limits<uint64_t>::max();
 // Marker for undefined value
 constexpr uint64_t NO_VALUE = std::numeric_limits<uint64_t>::max();
 
-// Kodi VideoPlayer internal buffer
-constexpr uint64_t KODI_VP_BUFFER_SECS = 8;
```
thexai, line 41: Also this was wrong currently because video player buffer (queue) is not 8 seconds anymore from https://github.com/xbmc/xbmc/pull/25360

Now is configurable and the default is 4s.

CastagnaIT:
closed because PR https://github.com/xbmc/xbmc/pull/27770 dont fix anymore demuxers callbacks

src/common/AdaptiveUtils.h
CastagnaIT: i'm not interested in changing it because the root of problem is in the VP side
anyway better a higher value than a low one since there is no way to obtain the current used value

PR #2005, opened 2026-03-03
Live stream jump

oliv3r:
## Description
To be able to jump forward (to the edge, e.g. the live moment) and jump backward (e.g. the start of the stream) on a live stream (when properly exposed/handled by the content provider/API, we need to cherry-pick commit b2da233e (Cleanups to chapter methods) from Piers which fixes the fact that we do not properly report the chapter count. Then add a commit to add a 'virtual' second chapter matching 'the end' which then gives us 2 chapters; ch1 'the start' and ch2 'now'. A third commit is to help with the start offset, like live_delay, we can add a bit of time to the start, where content providers love to add a conservative padding, to ensure they didn't miss the start of the program, or to make sure we get anything that was shown before the program (like an ad).

## Motivation and context
Being able to 'jump to start' (using |< for example) or 'jump to live/now' using >| which relies on proper chapters to trigger.

## How has this been tested?
Using retrospect and NLZIET we have livestreams from dutch television programs, moving this functionality to the |< and >| buttons allowed us to jump back/forward to the start/edge of a live stream reliably, with offset (where progress bar was showing a 'dot' of the offsetted jump point.

## Types of change

## Checklist:

### DIFF
--- a/inputstream.adaptive/addon.xml.in
+++ b/inputstream.adaptive/addon.xml.in
@@ -10,7 +10,7 @@
     name="adaptive"
     extension=""
     tags="true"
-    listitemprops="drm_legacy|license_type|license_key|license_url|license_url_append|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_upd_params|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max|live_delay|internal_cookies|config|manifest_config"
+    listitemprops="drm_legacy|license_type|license_key|license_url|license_url_append|license_data|license_flags|manifest_type|server_certificate|manifest_update_parameter|manifest_upd_params|manifest_params|manifest_headers|stream_params|stream_headers|original_audio_language|play_timeshift_buffer|pre_init_data|stream_selection_type|chooser_bandwidth_max|chooser_resolution_max|chooser_resolution_secure_max|live_delay|live_offset|internal_cookies|config|manifest_config"
     library_@PLATFORM@="@LIBRARY_FILENAME@"/>
   <extension point="xbmc.addon.metadata">
     <platform>@PLATFORM@</platform>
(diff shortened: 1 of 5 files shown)


oliv3r:
Since i don't have Piers, I couldn't run that, but the patches are trivial and should be very easy to port, i can push to a branch if needed/open a second MR if needed.

CastagnaIT:
First and foremost, as per docs/wiki always develops for the master branch and, if allowed in older branches. If for x reason you cannot do so, then do not even propose the PR on old branches (unless we are talking about backports, but each case is evaluated individually)

the new property `live_offset` its a clear workaround and i will not accept it

I won't even analyze the other changes because in the old branch i will only accept critical fixes that do not involve too many changes or dont require deeper testing that could potentially break other use cases

EDIT: In addition, given the complexity of live streams, you should provide free stream to test the use case. If this is not possible, at least attach one or more manifests to understand the dynamics of the stream

oliv3r:
> First and foremost, as per docs/wiki always develops for the master branch and, if allowed in older branches. If for x reason you cannot do so, then do not even propose the PR on old branches (unless we are talking about backports, but each case is evaluated individually)
> 
> the new property `live_offset` its a clear workaround and i will not accept it
> 
> I won't even analyze the other changes because in the old branch i will only accept critical fixes that do not involve too many changes or dont require deeper testing that could potentially break other use cases

harsh; but sure fair. i saw that Piers was the default branch, and as I only had omega available i wanted to at least start the discussion. So i'll re-do things on master no problem. But do explain how live_offset is just a hack? what would you propose? how is it different from live_delay?

@CastagnaIT much appreciated for your feedback so far, Thank you!

CastagnaIT:
Internal methods of the player should not be exposed to the user as setting, unless it is a problem that cannot be solved at the moment

You have not attached any sample stream to reproduce the problem, nor manifests to inspect, exists many different manifest structures of live streams and i have no idea of your use case

If you want to discuss about your playback problem i invite you to open an Issue thread and provide the necessary info

oliv3r:
For googlers finding this later; I've pushed a branch (doesn't show up here) to the same origin backporting it from piers.

PR #2006, opened 2026-03-04
Live stream jump piers

oliv3r:
## Description
To be able to jump forward (to the edge, e.g. the live moment) and jump backward (e.g. the start of the stream) on a live stream (when properly exposed/handled by the content provider/API, we need to cherry-pick commit b2da233e (Cleanups to chapter methods) from Piers which fixes the fact that we do not properly report the chapter count. Then add a commit to add a 'virtual' second chapter matching 'the end' which then gives us 2 chapters; ch1 'the start' and ch2 'now'. A third commit is to help with the start offset, like live_delay, we can add a bit of time to the start, where content providers love to add a conservative padding, to ensure they didn't miss the start of the program, or to make sure we get anything that was shown before the program (like an ad).

## Motivation and context
Being able to 'jump to start' (using |< for example) or 'jump to live/now' using >| which relies on proper chapters to trigger.

## How has this been tested?
Using retrospect and NLZIET we have livestreams from dutch television programs, moving this functionality to the |< and >| buttons allowed us to jump back/forward to the start/edge of a live stream reliably, with offset (where progress bar was showing a 'dot' of the offsetted jump point. Further more I used a test file as mentioned in #1517 and backported the changes to omega https://github.com/xbmc/inputstream.adaptive/pull/2005 but with fixes in place as per review comments.

## Types of change

## Checklist:

### DIFF
--- a/src/CompKodiProps.cpp
+++ b/src/CompKodiProps.cpp
@@ -402,6 +402,10 @@ void ADP::KODI_PROPS::CCompKodiProps::ParseManifestConfig(const std::string& dat
     {
       m_manifestConfig.liveDelay = jValue.get<uint64_t>();
     }
+    else if (configName == "live_offset" && jValue.is_number_unsigned())
+    {
+      m_manifestConfig.liveOffset = jValue.get<uint64_t>();
+    }
     else if (configName == "dash_utctiming" && jValue.is_object())
     {
       for (auto& [schemeId, jValue] : jValue.items()) // Iterate JSON dict
--- a/src/CompKodiProps.h
+++ b/src/CompKodiProps.h
@@ -70,6 +70,8 @@ struct ManifestConfig
   bool hlsFixDiscontSequence{false};
   // Custom delay from LIVE edge in seconds
   uint64_t liveDelay{0};
+  // Apply an offset in seconds from the live start (e.g. to skip pre-program buffered content)
+  uint64_t liveOffset{0};
   // Allows to set a custom UTC Timing for DASH live streams (schemeIdUri, value)
   std::optional<std::pair<std::string, std::string>> dashUTCTiming;
   // If true will ignore default KID provided from FMP4 init segment
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -1225,20 +1225,28 @@ int SESSION::CSession::GetChapterCount() const
   if (!m_adaptiveTree)
     return 0;
 
-  return static_cast<int>(m_adaptiveTree->m_periods.size());
+  int count = static_cast<int>(m_adaptiveTree->m_periods.size());
+  if (count > 0 && m_adaptiveTree->IsLive())
+    count += 1;
+  return count;
 }
 
 const char* SESSION::CSession::GetChapterName(int number) const
 {
+  if (!m_adaptiveTree)
+    return nullptr;
+
+  --number; // To convert chapter number to index
+  if (m_adaptiveTree->IsLive() && number == static_cast<int>(m_adaptiveTree->m_periods.size()))
+    return "Live";
+
   // Chapter name is shown on GUI info window
   // Manifests usually dont provide this info
   // so show the period ID for debugging purpose at request
-  if (CSrvBroker::GetSettings().IsDebugVerbose() && m_adaptiveTree)
+  if (CSrvBroker::GetSettings().IsDebugVerbose())
   {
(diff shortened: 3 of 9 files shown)


oliv3r:
marking as draft, just to see the pipeline work, as i've also added tests.

CastagnaIT:
now that i have had some time to read and understand it better
i will stop you right there before you continue wasting time
what you're doing is a hack, chapters must always match periods, and not generate fake/virtual chapters to use beyond their meaning

your proposal introduces complexities that lead to further complications in the maintenance of the add-on regards periods already difficult to maintain

managing periods is already difficult due to the multitude of different manifests that can not works well (e.g. ADS), and on HLS manifests the periods are not yet well managed

to note that already exists a property to allow start playback from start of TSB or his end
which unfortunately is not fully functional for all types of manifests due to the problems mentioned

we have more serious problems to solve in ISA before considering introducing strange things
so i cant accept this

IMO it would be better to have a shortcut key on Kodi to seek to the beginning or end of the stream, regardless of the chapters it would be a cleaner solution

oliv3r:
> now that i have had some time to read and understand it better i will stop you right there before you continue wasting time what you're doing is a hack, chapters must always match periods, and not generate fake/virtual chapters to use beyond their meaning
> 
> your proposal introduces complexities that lead to further complications in the maintenance of the add-on regards periods already difficult to maintain
> 
> managing periods is already difficult due to the multitude of different manifests that can not works well (e.g. ADS), and on HLS manifests the periods are not yet well managed
> 
> to note that already exists a property to allow start playback from start of TSB or his end which unfortunately is not fully functional for all types of manifests due to the problems mentioned
> 
> we have more serious problems to solve in ISA before considering introducing strange things so i cant accept this

that's sad of course, as the feature is quite useful from a UX point of view. I'll keep it for now in my local builds; but will look at:

> to note that already exists a property to allow start playback from start of TSB or his end which unfortunately is not fully functional for all types of manifests due to the problems mentioned 

to see if I could/should have used that instead and see if that can be fixed.

Having said that, there's 2 or 3 commits here worth looking at like:
* [AdaptiveStream] Make getMaxTimeMs() const
* [Test] Add tests for live_delay and live stream total time

> IMO it would be better to have a shortcut key on Kodi to seek to the beginning or end of the stream, regardless of the chapters it would be a cleaner solution

I can see how that could also work, though i'd do it slightly differently. |< and >| are only available in the player if there is a playlist or a chapter. I think this could be extended to 'or if there's a livestream which is seekable'. that would stay clean of ISA; and would only be a kodi change, if they are willing to accept it of course.

CastagnaIT:
> Having said that, there's 2 or 3 commits here worth looking at like:

> [AdaptiveStream] Make getMaxTimeMs() const

this cleanup makes sense but is not strictly needed, can be an idea to be done with other code cleanups

> [Test] Add tests for live_delay and live stream total time

- about additional "live delay" test, yes can be an idea, but i don't think it's worth adding code that will probably be reworked, the underlying problem is that there is much broader problem that has to do with HLS implementation, where live delay is not set, there are much more problems such as TSB thats generated only with the child manifest download, that lead to problem in managing play_timeshift_buffer property and others side effects. All is related/connected and this requires many code changes that i will not discuss here, so adding this test before solving all these problems is not very useful
- about "total time" test, is more complex that this, mediaPresentationDuration should be present but is not so on all dash manifest, as for example specs says that can period duration can be alternative value, and we currently fallback to calculated timeline to have more precision. if you add new test case should cover at least the most common manifest cases. to note that this variable is also used by other types of manifests, if the purpose is to verify the functioning of m_totalTime, it would be a half-measure since are missing the other cases. Is it worth knowing that specific manifest you are testing returns this value and not the other ones? Because, as Dash is implemented, if you add all the use cases supported are much broader, that depends on how its generated or provided the segments timeline...so i don't know if it's worth it at this point

> I can see how that could also work, though i'd do it slightly differently. |< and >| are only available in the player if there is a playlist or a chapter. I think this could be extended to 'or if there's a livestream which is seekable'. that would stay clean of ISA; and would only be a kodi change, if they are willing to accept it of course.

I think you should first open a feature request
https://forum.kodi.tv/forumdisplay.php?fid=9
and also ask for better alternatives to achieve what you want
doing what you ask on Kodi core will cover all playback uses, such as local files, PVR, InputStream, etc.... , not just ISA that is just a single InputStream addon

PR #2040, opened 2026-06-01
New parameters to "config" property to override streams media flags

### LINKED ISSUE #422: AdaptationSet default flag regression

CastagnaIT:
## Description

New parameters to `inputstream.adaptive.config`:
- `media_audio_langcode_default`: Set the "default" flag to the audio streams with specified language code
- `media_audio_langcode_original`: Set the "original" flag to the audio streams with specified language code
- `media_audio_stereo_pref`: Can be set to `true` when you want prefer stereo track as default instead of multi-channels (it will be applied to `media_audio_langcode_default` and `media_audio_langcode_original`)
- `media_audio_type_pref`: Can be set to `original` to prefer set as default the stream with original language, or `impaired` to prefer set as default the stream for impaired, otherwise left empty value for default behaviour.
- `media_subtitle_langcode_default`: Set the "default" flag to the subtitle streams with specified language code

about the language code format used should be the IETF RFC 5646 (BCP-47), but i suggest you to check also your manifest data

The property `inputstream.adaptive.original_audio_language` is now deprecated and will be removed from next kodi (major) version

NOTE FOR LANG CODE VARIANTS: At moment Kodi dont support any kind of track auto-selection fallback for language code that have variants e.g. "pt-BR" for Brazilian Portuguese vs "pt" for Portuguese, so here at moment its the same, if you set as default "pt-BR" but exists only a "pt" stream no stream will be set as default. This could be improved in future but for now we keep it simple.

## Motivation and context

Completed fix for #422 in more generic way

this will allow to override streams media flags in a generic way without being strictly limited to a single manifest type

the reason behind this is that Kodi stream media flags have not always have the same meaning of manifests attributes
and/or there are video services that dont follow exactly the specs and lead to wrong flags set to the audio/subtitle tracks

this will make the life easier for devs without the needs to implement a proxy on each addon to modify on the fly the manifest media flags

### HLS manifest side effect
The HLS manifests dont always provide the appropriate language codes and also dont always provide the right number of audio channels in the manifest, therefore the use of these new overrides may fails, you will need to check what your provider offers and act accordingly

## Types of change

## Checklist:

### DIFF
--- a/src/CompKodiProps.cpp
+++ b/src/CompKodiProps.cpp
@@ -52,7 +52,7 @@ constexpr std::string_view PROP_MANIFEST_CONFIG = "inputstream.adaptive.manifest
 constexpr std::string_view PROP_STREAM_PARAMS = "inputstream.adaptive.stream_params";
 constexpr std::string_view PROP_STREAM_HEADERS = "inputstream.adaptive.stream_headers";
 
-constexpr std::string_view PROP_AUDIO_LANG_ORIG = "inputstream.adaptive.original_audio_language";
+constexpr std::string_view PROP_AUDIO_LANG_ORIG = "inputstream.adaptive.original_audio_language"; //! @todo: deprecated, to be removed on Kodi 23
 constexpr std::string_view PROP_PLAY_TIMESHIFT_BUFFER = "inputstream.adaptive.play_timeshift_buffer";
 constexpr std::string_view PROP_LIVE_DELAY = "inputstream.adaptive.live_delay"; //! @todo: deprecated to be removed on Kodi 23
 constexpr std::string_view PROP_PRE_INIT_DATA = "inputstream.adaptive.pre_init_data"; //! @todo: deprecated to be removed on Kodi 23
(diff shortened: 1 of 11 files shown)


CastagnaIT:
@kodiai please review

CastagnaIT:
@kodiai updated please review

CastagnaIT:
@kodiai updated please review

CastagnaIT:
@matthuisman if iirc you was using `inputstream.adaptive.original_audio_language` on some add-on,
have you some time to test this new way?

PR #2064, opened 2026-06-20
[mpegts] Accept any PID number not only PAT with PID 0

### LINKED ISSUE #2010: [Feature request] Support in-correct IV that ffmpeg plays OK
### Describe the feature

In one of my add-ons, we generate a new higher quality HLS playlist by grabbing the 720p playlist and replacing the .ts urls with a known 1080p version.

ffmpeg plays this just fine - but inputstream.adaptive refuses too.
I suspect because the IV is in-correct and not known.

"AI" says that ffmpeg can handle that by just skipping the first few bytes and then all is OK.
Not sure if this is accurate or not

Anyway - it would be cool if IA could handle it too :)

Here is the sub manifest.
So this originally was pointing to -TEN-3000000.key but then replaced with -TEN-5000000.key
As well as all the .ts files
[sub-out.m3u8.txt](https://github.com/user-attachments/files/26079932/sub-out.m3u8.txt)
Note: ffmpeg seems to be able to play it even with IV not present at all as well

Here is the .key and the first two segments:
[0L-AUSsurvivor-SURV12_001_20260223-TEN-5000000-2.ts.txt](https://github.com/user-attachments/files/26080068/0L-AUSsurvivor-SURV12_001_20260223-TEN-5000000-2.ts.txt)
[0L-AUSsurvivor-SURV12_001_20260223-TEN-5000000-1.ts.txt](https://github.com/user-attachments/files/26080069/0L-AUSsurvivor-SURV12_001_20260223-TEN-5000000-1.ts.txt)
[0L-AUSsurvivor-SURV12_001_20260223-TEN-5000000.key.txt](https://github.com/user-attachments/files/26080067/0L-AUSsurvivor-SURV12_001_20260223-TEN-5000000.key.txt)

When inputstream adaptive tries to play - it just hangs loading forever and error on every segment.
Log attached
[ia.log.txt](https://github.com/user-attachments/files/26080167/ia.log.txt)

For now, I've told Kodi to just ignore IA and always playback with ffmpeg

### Motivation

playing back more content

No

No

CastagnaIT:
## Description

From what i understand on TS streams PAT should always be present and have PID 0
while PMT can have different PID numbers

a stream such as provided from the issue #2010 dont contains the PAT, but only PMT with PID 4097
as VLC shown:
```
ts debug: pid[4097] unknown
ts debug: pid[256] unknown
ts debug: pid[257] unknown
```

this is not a standard stream, but VLC is able to play it with some kind of fallback
so the reason why this video cant be played its not a bug, but its needed make the parser less strict

I found this solution with the help of AI, seem to works correctly without cause problems with regular TS streams

## Motivation and context

fix #2010

## How has this been tested?

manually tested with provided files from issue

## Types of change

## Checklist:

### DIFF
--- a/lib/mpegts/mpegts/tsDemuxer.cpp
+++ b/lib/mpegts/mpegts/tsDemuxer.cpp
@@ -471,20 +471,26 @@ int AVContext::ProcessTSPacket()
   }
 
   it = this->packets.find(this->pid);
-  if (it == this->packets.end())
+  if (it == this->packets.end()) // Not registered PID.
   {
-    // Not registred PID
-    // We are waiting for unit start of PID 0 else next packet is required
-    if (this->pid == 0 && this->payload_unit_start)
+    // Only auto-register if this payload_unit_start actually contains a PSI table
+    if (this->payload_unit_start && this->payload && this->payload_len > 1)
     {
-      // Registering PID 0
-      Packet pid0;
-      pid0.pid = this->pid;
-      pid0.packet_type = PACKET_TYPE_PSI;
-      pid0.continuity = continuity_counter;
-      it = this->packets.insert(it, std::make_pair(this->pid, pid0));
+      const uint8_t pointerField = av_rb8(this->payload); // pointer_field is at payload[0];
+      if (static_cast<size_t>(pointerField) + 1 < this->payload_len)
+      {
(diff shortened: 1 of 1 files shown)


CastagnaIT:
@kodiai review

PR #2086, opened 2026-06-29
[mpegts] Enlarge H.264 SPS/PPS raw_data buffers to 256 bytes

wmdebrito:
## Description

The per-NAL raw_data buffers used to build the codec extra data were only 64 bytes. Some SPS with full VUI/HRD parameters can exceed that (e.g. Wowza-originated HLS), so the SPS was silently dropped (raw_data_size reset to 0) and the H.264 extra data ended up empty.

Kodi then rejects the video stream with "Codec id 27 require extradata" and disables it, leaving an audio-only stream that immediately hits EOF.

Increase the SPS/PPS raw_data buffers to 256 bytes so that for some large SPS extra data is built correctly.

Also extra_data was increase from 512 to 1024 to accommodate rare situations.
NAL buffers are now 256 bytes each, so a single legitimate SPS+PPS can be up to 4+255+4+255 = 518 bytes. With only 512, the guard I added would correctly refuse to overflow — but it would then drop a valid PPS, silently reintroducing a decode failure. 1024 ensures the destination can hold whatever the 256-byte NAL buffers can legitimately produce. The guard prevents the crash; the 1024 prevents truncation.

## Motivation and context

Fix some streams that are not playing due to small buffer size.

## How has this been tested?

Tested on Kodi 21 with a custom IA build. The falling stream was now able to be played. No regression on other streams

Stream: https://streamtv.novotempo.com/CDN-TV-PT/smil:tvnovotempo.smil/playlist.m3u8
  12:07:59.315  CDVDDemuxClient::RequestStream(): added/updated stream 1001 with codec_id 27
  12:07:59.315  CDVDDemuxClient::RequestStream(): added/updated stream 1002 with codec_id 86018
  12:07:59.316  Opening stream: 1001 source: 256
  12:07:59.932  [WHITELIST] Searching the whitelist for: width: 1280, height: 720, fps: 29.970, 3D: false
  12:07:59.979  error: OpenStream: Codec id 27 require extradata.
  12:07:59.979  warning: OpenStream - Unsupported stream 1001. Stream disabled.
  12:07:59.979  Opening stream: 1002 source: 256
  12:08:00.004  CDVDAudioCodecFFmpeg::Open() Successful opened audio decoder aac
  12:08:00.034  Process - eof reading from demuxer

## Types of change

## Checklist:

### DIFF
--- a/lib/mpegts/mpegts/elementaryStream.h
+++ b/lib/mpegts/mpegts/elementaryStream.h
@@ -61,7 +61,7 @@ namespace TSDemux
     int                   bit_rate;
     int                   bits_per_sample;
     bool                  interlaced;
-    uint8_t               extra_data[512];
+    uint8_t               extra_data[1024];
     int                   extra_data_size;
   };
 
--- a/lib/mpegts/mpegts/ES_h264.cpp
+++ b/lib/mpegts/mpegts/ES_h264.cpp
@@ -158,18 +158,30 @@ void ES_h264::Parse(STREAM_PKT* pkt)
       {
         if (m_streamData.sps[0].raw_data_size)
         {
-          uint8_t *ed(stream_info.extra_data);
-          stream_info.extra_data_size = 4 + m_streamData.sps[0].raw_data_size;
-          ed[0] = ed[1] = ed[2] = 0, ed[3] = 1, ed += 4;
-          memcpy(ed, m_streamData.sps[0].raw_data, m_streamData.sps[0].raw_data_size), ed += m_streamData.sps[0].raw_data_size;
+          uint8_t* ed = stream_info.extra_data;
+          const uint8_t* edEnd = stream_info.extra_data + sizeof(stream_info.extra_data);
+          stream_info.extra_data_size = 0;
+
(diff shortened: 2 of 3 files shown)


lib/mpegts/mpegts/elementaryStream.h
```diff
@@ -61,7 +61,7 @@ namespace TSDemux
     int                   bit_rate;
     int                   bits_per_sample;
     bool                  interlaced;
-    uint8_t               extra_data[512];
+    uint8_t               extra_data[1024];
     int                   extra_data_size;
```
CastagnaIT, line 65: a better change should be replace both with a `std::vector<uint8_t>`
so that the size is not hardcoded but dynamic, you will solve the problem at the root
but its fine anyway if you want to keep it like this

CastagnaIT (approved):
thanks for the fix, please backport

lib/mpegts/mpegts/elementaryStream.h
wmdebrito: It is a good idea, I will create it in another PR to test it better.

PR #2095, opened 2026-08-09
[mpegts] Drop provisional PSI PID registration on invalid section

ksooo:
## Description

Commit 4ebb4ca ("[mpegts] Accept any PID number not only PAT with PID 0") replaced
the fixed `pid == 0` condition in `AVContext::ProcessTSPacket()` with a heuristic
that registers any unknown PID as `PACKET_TYPE_PSI` when the payload looks like it
starts with a `pointer_field` followed by `table_id` 0x00 (PAT) or 0x02 (PMT).

That heuristic has a false positive: a PES packet begins with the start code prefix
`00 00 01`, which reads as `pointer_field = 0x00` and therefore `table_id = 0x00`
(PAT). Any elementary stream PID whose first observed `payload_unit_start` packet
reaches the demuxer before the corresponding PAT/PMT is registered as PSI, and its
media payload is discarded from that point on.

This change keeps the heuristic but makes the registration provisional:

- `Packet` gains an `auto_registered` flag, set when a PID is registered by the
  heuristic rather than announced by a PAT/PMT.
- `parse_ts_psi()` clears the flag as soon as a section passes the CRC32 check
  added in be978e1, i.e. once the PID has proven it really carries PSI.
- `ProcessTSPayload()` drops the registration again if the section fails CRC
  validation or the parser returns `AVCONTEXT_TS_ERROR`, so the PID can be
  registered correctly by a later PMT.

The erase happens in `ProcessTSPayload()` rather than in `parse_ts_psi()` because
`this->packet` points into the `packets` map; `this->packet` is set to `NULL`
afterwards, which all accessors already handle. `AVCONTEXT_TS_ERROR` is normalised
to `AVCONTEXT_CONTINUE` in this path since the error was caused by our own bogus
registration and does not justify forcing the caller into a resync. A genuine PSI
PID whose section was merely damaged in transit is registered again on the next
section start, so the behaviour is self-healing.

## Motivation and context

Follow-up to be978e1 ("[mpegts] Validate PSI section CRC32 before parsing PAT/PMT"),
which fixed the crash caused by the same false positive but left the bogus PID
registration in place: on CRC failure the section buffer is reset and the function
returns, while the entry stays in `this->packets` as `PACKET_TYPE_PSI`.

The remaining symptom shows up on HLS/TS live streams with server-side ad insertion,
where every `#EXT-X-DISCONTINUITY` starts a separately muxed segment. When the video
PID is hijacked during the stream information scan of the first segment of a new
period, no SPS is parsed, `ES_h264` produces empty extra data, and Kodi rejects and
disables the video stream:

    error <general>: CVideoPlayerVideo::OpenStream(CDVDStreamInfo): Codec id 27 require extradata.
    warning <general>: OpenStream - Unsupported stream 3001. Stream disabled.
    debug <inputstream.adaptive>: EnableStream(3001: false)

Playback then continues audio-only until the decoder hits EOF and the stream is
reopened. This is the same failure class described in the commit message of
c9fc8ee, but it is not caused by undersized SPS/PPS buffers and therefore still
reproduces on 22.3.19-Piers.

Reverting 4ebb4ca also resolves it, but that would reintroduce the problem 4ebb4ca
was written for (streams whose PMT is not on PID 0 and that provide no PAT), so
this PR keeps the heuristic instead.

## How has this been tested?

Environment: Kodi 22.0-BETA1 master (Git:20260809-ad2b846cc6), macOS 26.6.1 on
Apple M3, debug build, FFmpeg 8.1.2-Kodi, inputstream.adaptive 22.3.19-Piers,
pvr.plutotv (Piers) on a German Pluto.tv channel — HLS/TS with AES-128 encrypted,
separately muxed ad segments and 6-7 periods per ad pod.

Runtime tests, counting occurrences in kodi.log:

| Build | Periods | `require extradata` / stream disabled |
| --- | --- | --- |
| 22.3.19-Piers unmodified | 8 | 1 |
| 22.3.19-Piers with 4ebb4ca reverted | 7 | 0 |
| 22.3.19-Piers with 4ebb4ca reverted, longer run | 21 (3 ad pods, ~1 h) | 0 |
| 22.3.19-Piers with this patch | 20 (3 ad pods, ~1 h) | 0 |

The revert of 4ebb4ca is listed only as a control to confirm the root cause; this
PR keeps the heuristic and achieves the same result.

Additionally exercised with a standalone harness driving the public `AVContext` API
over a synthetic transport stream that reproduces the ordering (PES packets of an
unregistered elementary stream PID arriving before the PAT/PMT of the segment). The
elementary stream is discovered correctly and the run is clean under
`-fsanitize=address,undefined`.

n/a

## Types of change

## Checklist:

Note on code guidelines: `lib/mpegts/` is imported third-party code with its own
style; the changes follow the surrounding conventions rather than the Kodi guidelines.
Happy to adjust if you prefer otherwise.

### DIFF
--- a/lib/mpegts/mpegts/tsDemuxer.cpp
+++ b/lib/mpegts/mpegts/tsDemuxer.cpp
@@ -39,6 +39,7 @@ AVContext::AVContext(TSDemuxer* const demux, uint64_t pos, uint16_t channel)
   , payload(NULL)
   , payload_len(0)
   , packet(NULL)
+  , invalid_psi_section(false)
 {
   m_demux = demux;
   memset(av_buf, 0, sizeof(av_buf));
@@ -58,6 +59,7 @@ void AVContext::Reset(void)
   payload_unit_pos = 0;
   prev_payload_unit_pos = 0;
   packet = NULL;
+  invalid_psi_section = false;
 }
 
 uint16_t AVContext::GetPID() const
@@ -486,6 +488,11 @@ int AVContext::ProcessTSPacket()
           p.pid = this->pid;
           p.packet_type = PACKET_TYPE_PSI;
           p.continuity = continuity_counter;
+          // The heuristic above cannot distinguish a PSI table from arbitrary
+          // payload, e.g. a PES packet starts with the start code prefix
+          // 00 00 01, which reads as pointer_field 0 and table_id 0x00 (PAT).
+          // Keep the registration provisional until a section validates.
+          p.auto_registered = true;
           it = this->packets.insert(it, std::make_pair(this->pid, p));
         }
       }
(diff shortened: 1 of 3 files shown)


CastagnaIT (approved):
tested a couple of problematic TS videos
LGTM thanks

PR #2097, opened 2026-08-15
[Session] Fix data race on m_periods in Kodi core chapter callbacks

ksooo:
## Description

`CSession::GetChapterPos()` and the other chapter accessors read
`m_adaptiveTree->m_periods` from the Kodi player thread without any
synchronisation, while the manifest update thread can add and remove periods at
the same time. This is the problem described by the existing `@todo` in
`GetChapterPos()`, which this PR removes.

The unsynchronised access was:

```cpp
sum += (m_adaptiveTree->m_periods[number - 1]->GetTlDuration() * STREAM_TIME_BASE) /
       m_adaptiveTree->m_periods[number - 1]->GetTimescale();
```

The update thread appends a period at every `#EXT-X-DISCONTINUITY`
(`HLSTree.cpp`, `m_periods.push_back(std::move(newPeriodPtr))`) and drops expired
ones (`m_periods.erase(...)`). When `push_back` reallocates the vector the
`unique_ptr` elements are move constructed into the new buffer, so a concurrent
reader can observe a null element. The bounds check above the access is racy as
well, since `size()` is read separately from the element access.

The manifest update lock (`GetTreeUpdMutex()`) is taken in `AdaptiveStream.cpp`
and `HLSTree.cpp`, but nowhere in `Session.cpp`.

This change serves the callbacks from an immutable snapshot instead:

- `AdaptiveTree` gains `ChaptersSnapshot` (period id, `tlDuration`, `timescale`,
  plus the index of the current period) held as a
  `std::shared_ptr<const ChaptersSnapshot>`.
- `RefreshChaptersSnapshot()` rebuilds it and is called from `PostOpen()` before
  the update thread starts, from the update worker right after
  `OnUpdateSegments()`, and from `CSession::InitializePeriod()` after a period
  change.
- All five accessors (`GetChapter()`, `GetChapterCount()`, `GetChapterName()`,
  `GetChapterPos()`, `GetChapterStartTime()`) now read only the snapshot and no
  longer touch `m_periods`.

The reader side copies the `shared_ptr` under a mutex that is never held while
doing any work, which keeps the player callback path cheap. Simply taking the
update lock in the accessors is not viable: `TreeUpdateThread::lock()` is
`Pause()` and can block on a manifest download, and these callbacks run on every
player loop iteration — the second half of the `@todo`.

`STREAM_TIME_BASE` scaling stays in `Session.cpp` because the constant comes from
the Kodi headers and is not necessarily visible in `AdaptiveTree.cpp`. The
arithmetic is otherwise unchanged, with an added guard against `timescale == 0`
that did not exist before. `GetChapterStartTime()` previously iterated
`m_periods` with no null or bounds check at all.

Two points I am not sure about and would like your opinion on:

- `InitializePeriod()` takes `std::lock_guard<TreeUpdateThread>` to rebuild the
  snapshot from a stable `m_periods`. This runs once per period change on the
  demux thread rather than on the callback path, and follows the pattern already
  used in `AdaptiveStream.cpp`, but it is the riskiest part of the change.
- `GetChapterName()` returns a `const char*`, so the session has to keep the
  snapshot alive across the return. That is a workaround for the signature rather
  than a clean solution, and I am happy to change the approach if you prefer.

## Motivation and context

Reproducible SIGSEGV on Android when a live HLS stream with server-side ad
insertion enters an ad break. Tombstone (NVIDIA Shield, Android 11, arm64):

```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x178
Cause: null pointer dereference
#00 pc 000000000012f698  libinputstream.adaptive.so
#01 pc 00000000023cd818  libkodi.so (non-virtual thunk to CInputStreamAddon::GetChapterPos(int)+28)
#02 pc 000000000242a980  libkodi.so (CVideoPlayer::UpdatePlayState(double)+1416)
#03 pc 000000000242b700  libkodi.so (CVideoPlayer::Process()+432)
```

The null dereference matches a reader observing a moved-from `unique_ptr` during
a `push_back` reallocation.

The kodi.log shows the period switch at 09:14:45.220 and the crash at
09:14:49.538. In between, `CDVDVideoCodecAndroidMediaCodec::Open - InstanceGuard
locked` delays the video codec reopen by about 4 seconds, so the reopen coincides
with a manifest update. The same stream on macOS shows an equivalent
VideoToolbox delay of 3.3–4.0 s across 20 period switches and does not crash —
the race is timing dependent and only lost on the slower device.

Unrelated to #2095; the crash occurs both with and without that patch.

## How has this been tested?

Quick runtime-test on macOS, no regressions found.
Extended runtime-test on Android (on the machine I encountered the crashes): tested more then 3 hours playback in row, at least 6 ad breaks, no crashes occurred. No regressions found.

n/a

## Types of change

## Checklist:

### DIFF
--- a/src/Session.cpp
+++ b/src/Session.cpp
@@ -295,6 +295,13 @@ void SESSION::CSession::InitializePeriod()
     // Complete the transition into the new period
     m_adaptiveTree->m_currentPeriod = m_adaptiveTree->m_nextPeriod;
     m_adaptiveTree->OnPeriodChange();
+    {
+      // Block manifest updates so the snapshot is built from a stable m_periods.
+      // This runs once per period change, not on the player callback path.
+      std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(
+          m_adaptiveTree->GetTreeUpdMutex());
+      m_adaptiveTree->RefreshChaptersSnapshot();
+    }
   }
 
   // Clean to create new SESSION::STREAM objects. One for each AdaptationSet/Representation
@@ -679,6 +686,14 @@ bool SESSION::CSession::PrepareStream(CStream& stream)
     {
       if (!m_adaptiveTree->PrepareRepresentation(period, adp, repr))
         return false;
+
+      {
+        // PrepareRepresentation can process a child manifest (HLS) and by that add and
+        // remove periods. This runs after InitializePeriod, so refresh the snapshot.
+        std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(
+            m_adaptiveTree->GetTreeUpdMutex());
+        m_adaptiveTree->RefreshChaptersSnapshot();
+      }
     }
   }
 
@@ -853,31 +868,42 @@ bool SESSION::CSession::SeekTime(double seekTime, bool& isError)
     seekTime = 0;
 
   // Check if we leave our current period
-  double chapterTime{0};
-  auto pi = m_adaptiveTree->m_periods.cbegin();
-
-  for (; pi != m_adaptiveTree->m_periods.cend(); pi++)
   {
-    chapterTime += double((*pi)->GetTlDuration()) / (*pi)->GetTimescale();
-    if (chapterTime > seekTime)
-      break;
+    // Block manifest updates, the update thread can add and remove periods at any time.
+    // Unlike the chapter callbacks this is not called on every player loop iteration,
+    // so blocking updates here is acceptable.
+    std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(
+        m_adaptiveTree->GetTreeUpdMutex());
+
+    if (m_adaptiveTree->m_periods.empty())
+      return false;
+
+    double chapterTime{0};
+    auto pi = m_adaptiveTree->m_periods.cbegin();
+
+    for (; pi != m_adaptiveTree->m_periods.cend(); pi++)
+    {
+      chapterTime += double((*pi)->GetTlDuration()) / (*pi)->GetTimescale();
+      if (chapterTime > seekTime)
+        break;
+    }
+
+    if (pi == m_adaptiveTree->m_periods.cend())
+      --pi;
+
+    chapterTime -= double((*pi)->GetTlDuration()) / (*pi)->GetTimescale();
+
+    if ((*pi).get() != m_adaptiveTree->m_currentPeriod)
+    {
+      LOG::Log(LOGDEBUG, "SeekTime: seeking into new chapter: %d",
+               static_cast<int>((pi - m_adaptiveTree->m_periods.begin()) + 1));
+      SeekChapter(static_cast<int>(pi - m_adaptiveTree->m_periods.begin()) + 1);
+      m_chapterSeekTime = seekTime;
+      return true;
+    }
+
+    seekTime -= chapterTime;
   }
-
-  if (pi == m_adaptiveTree->m_periods.cend())
-    --pi;
-
-  chapterTime -= double((*pi)->GetTlDuration()) / (*pi)->GetTimescale();
-
-  if ((*pi).get() != m_adaptiveTree->m_currentPeriod)
-  {
-    LOG::Log(LOGDEBUG, "SeekTime: seeking into new chapter: %d",
-             static_cast<int>((pi - m_adaptiveTree->m_periods.begin()) + 1));
-    SeekChapter(static_cast<int>(pi - m_adaptiveTree->m_periods.begin()) + 1);
-    m_chapterSeekTime = seekTime;
-    return true;
-  }
-
-  seekTime -= chapterTime;
 
   if (m_adaptiveTree->IsLive())
   {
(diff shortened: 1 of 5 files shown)


src/common/AdaptiveTree.h
```diff
@@ -326,6 +326,42 @@ class ATTR_DLL_LOCAL AdaptiveTree
...
+  {
+    std::vector<ChapterInfo> chapters;
+    int currentIndex{-1}; //!< 0 based index of the period being played, -1 if unknown
+  };
+
+  /*!
+   * \brief Get the current chapter snapshot. Safe to call from any thread.
+   * \return A snapshot, never nullptr.
+   */
+  std::shared_ptr<const ChaptersSnapshot> GetChaptersSnapshot() const;
```
CastagnaIT, line 356: its really needed the use of shared_ptr?
i see that the getter its already mutex protected and return a const object

src/Session.h
```diff
@@ -270,5 +270,8 @@ class ATTR_DLL_LOCAL CSession : public adaptive::AdaptiveStreamObserver
   uint64_t m_chapterStartTime{0}; // In STREAM_TIME_BASE
   double m_chapterSeekTime{0.0}; // In seconds
   uint8_t m_mediaTypeMask{0};
+
+  // Keeps the snapshot backing the string returned by GetChapterName() alive
+  mutable std::shared_ptr<const adaptive::AdaptiveTree::ChaptersSnapshot> m_chaptersSnapshot;
```
CastagnaIT: yeah this reintroduce a workaround, the problem is the bad inputstream API method
that should be reworked in a different way

if you want keep as is, please add a TODO comment about the problem to be fixed

src/common/AdaptiveStream.cpp
```diff
@@ -922,6 +926,9 @@ bool adaptive::AdaptiveStream::ensureSegment()
             m_tree->OnStreamChange(current_period_, current_adp_, prevRep, newRep);
             // Try aligning the segment to ensure that it exists on the changed representation
             m_tree->OnAlignSegment(current_period_, current_adp_, prevRep, newRep, queueSegment);
+            // OnAlignSegment can process a child manifest (HLS) and by that add and
+            // remove periods, refresh while updates are still blocked
```
CastagnaIT: need to fix the wrong sentence, and also i would like to have a generic comment so remove references to specific manifest type

```suggestion
            // OnStreamChange can process a manifest and by that add and
            // remove periods, refresh while updates are still blocked
```

src/Session.cpp
```diff
@@ -295,6 +295,13 @@ void SESSION::CSession::InitializePeriod()
     // Complete the transition into the new period
     m_adaptiveTree->m_currentPeriod = m_adaptiveTree->m_nextPeriod;
     m_adaptiveTree->OnPeriodChange();
+    {
+      // Block manifest updates so the snapshot is built from a stable m_periods.
+      // This runs once per period change, not on the player callback path.
+      std::lock_guard<adaptive::AdaptiveTree::TreeUpdateThread> lckUpdTree(
+          m_adaptiveTree->GetTreeUpdMutex());
+      m_adaptiveTree->RefreshChaptersSnapshot();
+    }
```
CastagnaIT, line 304: from what i see this refresh its not needed, since its already covered by others

src/Session.h
```diff
@@ -270,5 +270,13 @@ class ATTR_DLL_LOCAL CSession : public adaptive::AdaptiveStreamObserver
...
   double m_chapterSeekTime{0.0}; // In seconds
   uint8_t m_mediaTypeMask{0};
+
+  //! @todo: the InputStream API method GetChapterName returns a "const char*" whose
+  //! lifetime the addon cannot control, so the value has to be cached here to keep the
+  //! returned pointer valid after the method returns. Kodi core copies the string right
+  //! away (CInputStreamAddon::GetChapterName), and the only two callers are on the GUI
+  //! thread, so this is safe in practice but remains a workaround for a bad API.
+  //! The API should be reworked to return a std::string, then this member can be removed.
+  mutable std::string m_chapterName;
```
CastagnaIT, line 280: mutable is a leftover?

src/common/AdaptiveStream.cpp
ksooo: fixed

src/Session.cpp
ksooo: This is the only site that updates currentIndex when the current period changes. The other four sites all fire on manifest updates, and each is conditional:

* update thread worker — only when `m_updateInterval > 0`
* after `OnRequestSegments` — only when `HasManifestUpdatesSegs()`, i.e. live with interval 0
* after `OnStreamChange` — only on a quality switch
* after `PrepareRepresentation` in `PrepareStream` — only when `IsReqPrepareStream()`, and `m_isReqPrepareStream` is set in exactly one place: HLSTree.cpp:153

So for a multi-period DASH VOD — the classic chapters case — no refresh site can ever fire after `PostOpen()`. Remove the one in `InitializePeriod()` and `GetChapter()` keeps reporting the period that was current at open for the rest of playback, and `GetChapterStartTime()` returns the wrong offset. The old code computed both live from `m_periods`, so this would be a regression, not a cleanup.

src/Session.h
ksooo: No. `GetChapterName()` is declared `const` (Session.h:170), so assigning `m_chapterName` in it needs `mutable`. `const` + `mutable` is the normal idiom for caching in a const accessor, and the only caller is main.cpp:534 on a non-const session.

src/Session.cpp
CastagnaIT: you have right i forgot that need to read the index

PR #2105, opened 2026-08-23
[HLS] Support FairPlay-signaled CMAF with ClearKey

ldzhjn:
## Description

Preserve a 128-bit hexadecimal KID from a FairPlay `skd://` URI and expose the
track encryption KID as Common Encryption information when an encrypted CMAF
initialization segment contains no usable PSSH box.

This allows a user-configured ClearKey decrypter to open `SAMPLE-AES` fMP4
streams that use FairPlay HLS signaling. It does not implement FairPlay key
management or license acquisition.

## Motivation and context

Some CMAF HLS streams signal `KEYFORMAT="com.apple.streamingkeydelivery"` and
use `cbcs`, while carrying the default KID only in the `skd://` URI and/or the
track's `tenc` box. Without a PSSH box, the initialization segment previously
returned no DRM information, so the ClearKey session could not be matched to
the configured KID.

The parser now accepts only `skd://` values that normalize to exactly 16 bytes
of hexadecimal data. Other FairPlay asset identifiers remain unchanged.

## How has this been tested?

- All 62 native unit tests pass on macOS.
- Added an HLS regression test for a hyphenated, uppercase `skd://` KID and
  `SAMPLE-AES`/CBC mode.
- Built successfully for tvOS arm64 against Kodi 22 Addon API 2.0.3.
- Verified playback of a 3840x2160 50 fps HEVC 10-bit CBCS CMAF HLS stream on
  Apple TV using a configured ClearKey, including video and audio decryption.

Not applicable.

## Types of change

## Checklist:


src/samplereader/FragmentedSampleReader.cpp
```diff
@@ -210,6 +210,18 @@ std::vector<DRM::DRMInfo> CFragmentedSampleReader::GetInitDRMInfo()
 
       drmInfos.emplace_back(drmInfo);
     }
+
+    // Some encrypted CMAF streams carry the default KID only in the track's tenc box
+    // and do not include a usable PSSH box. Keep the Common Encryption information
+    // so a configured decrypter can create its session with that KID.
+    if (drmInfos.empty() && !defaultKid.empty() && cryptoMode != CryptoMode::NONE)
```
CastagnaIT: drop the crypto mode check, there is already a error log print and also above its not a blocking operation here
```suggestion
    if (drmInfos.empty() && !defaultKid.empty())
```

src/parser/HLSTree.cpp
```diff
@@ -1315,13 +1316,33 @@ void adaptive::CHLSTree::ProcessEncryption(
...
   else if (STRING::CompareNoCase(keyFormat, "com.apple.streamingkeydelivery"))
   {
     DRM::DRMInfo& drmInfo = drmInfos[DRM::KS_FAIRPLAY]; // Create or update
     drmInfo.keySystem = DRM::KS_FAIRPLAY;
 
-    // There is no DRM/Key management implementation
+    // FairPlay HLS can identify the content key with an skd:// URI. Although FairPlay
+    // key management is not implemented, preserve a valid KID so a configured
+    // ClearKey decrypter can handle SAMPLE-AES fMP4 streams.
+    constexpr std::string_view SKD_SCHEME{"skd://"};
```
CastagnaIT: better static, since you have kept it inside a method
```suggestion
    static constexpr std::string_view URI_SDK_SCHEME{"skd://"};
```

src/parser/HLSTree.cpp
```diff
@@ -1315,13 +1316,33 @@ void adaptive::CHLSTree::ProcessEncryption(
...
+    if (uriUrl.size() > SKD_SCHEME.size() &&
+        STRING::CompareNoCase(uriUrl.substr(0, SKD_SCHEME.size()), SKD_SCHEME))
```
CastagnaIT: there is no need to have this complex condition
```suggestion
    if (STRING::StartsWith(...))
```

src/parser/HLSTree.cpp
```diff
@@ -1315,13 +1316,33 @@ void adaptive::CHLSTree::ProcessEncryption(
...
+    {
+      std::string keyId = uriUrl.substr(SKD_SCHEME.size());
+      if (keyId.size() > 2 && STRING::CompareNoCase(keyId.substr(0, 2), "0x"))
```
CastagnaIT: also here STRING::StartsWith

src/parser/HLSTree.cpp
```diff
@@ -1315,13 +1316,33 @@ void adaptive::CHLSTree::ProcessEncryption(
...
+        keyId.erase(0, 2);
+      STRING::ReplaceAll(keyId, "-", "");
+
+      const bool isHexKeyId =
+          keyId.size() == 32 &&
+          std::all_of(keyId.cbegin(), keyId.cend(),
+                      [](unsigned char c) { return std::isxdigit(c) != 0; });
```
CastagnaIT: this check, should be made by add an appropriate method in the decrypters\Helpers.h that can be used to validate a KID e.g. `IsValidKID`

src/parser/HLSTree.cpp
```diff
@@ -1315,13 +1316,33 @@ void adaptive::CHLSTree::ProcessEncryption(
...
+      if (isHexKeyId)
+        drmInfo.defaultKid = STRING::ToLower(keyId);
+      else
+        LOG::LogF(LOGDEBUG, "FairPlay SKD URI does not contain a 128-bit hexadecimal KID");
```
CastagnaIT: i think this log can be removed, fairplay documentation dont specify any restriction about the format of the URI path, so it might report useless notifications

src/parser/HLSTree.cpp
```diff
@@ -1315,13 +1316,33 @@ void adaptive::CHLSTree::ProcessEncryption(
...
+    if (uriUrl.size() > SKD_SCHEME.size() &&
+        STRING::CompareNoCase(uriUrl.substr(0, SKD_SCHEME.size()), SKD_SCHEME))
+    {
+      std::string keyId = uriUrl.substr(SKD_SCHEME.size());
+      if (keyId.size() > 2 && STRING::CompareNoCase(keyId.substr(0, 2), "0x"))
+        keyId.erase(0, 2);
+      STRING::ReplaceAll(keyId, "-", "");
+
+      const bool isHexKeyId =
+          keyId.size() == 32 &&
```
CastagnaIT: make sure that this code is able to extract KID also from URI such as the following example:
`skd://11111111-1234-1234-1234-000000000000:11111111123412341234000000000000`
and add it on the test case

src/parser/HLSTree.cpp
```diff
@@ -21,6 +21,7 @@
 #include "utils/log.h"
 
 #include <algorithm>
+#include <cctype>
```
CastagnaIT: after code adaptations this include can be removed

CastagnaIT (changes requested):
not bad idea but needs a bit of more cleanups

src/samplereader/FragmentedSampleReader.cpp
ldzhjn: Fixed in 32b5c77: removed the  check as suggested.

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: simplified the scheme check to .

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: removed the diagnostic for unrecognized SKD URI path formats.

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: the  prefix check now uses .

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: extraction now takes the final colon/path component before normalization. The HLS regression fixture uses this exact two-value URI form.

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: made the method-local SKD scheme constant .

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: added reusable  in  and use it here.

src/parser/HLSTree.cpp
ldzhjn: Fixed in 32b5c77: removed the now-unused `<cctype>` include.

src/parser/HLSTree.cpp
```diff
@@ -1319,13 +1315,34 @@ void adaptive::CHLSTree::ProcessEncryption(
...
+    // FairPlay HLS can identify the content key with an skd:// URI. Although FairPlay
+    // key management is not implemented, preserve a valid KID so a configured
+    // ClearKey decrypter can handle SAMPLE-AES fMP4 streams.
+    static constexpr std::string_view URI_SKD_SCHEME{"skd://"};
+    if (STRING::StartsWith(uriUrl, URI_SKD_SCHEME))
+    {
+      std::string keyId = uriUrl.substr(URI_SKD_SCHEME.size());
+      const size_t queryPos = keyId.find_first_of("?#");
+      if (queryPos != std::string::npos)
+        keyId.erase(queryPos);
```
CastagnaIT, line 1334: its not clear this use case, please show an example of the URI

ldzhjn: Example (identifiers redacted):

    skd://linear/00112233445566778899AABBCCDDEEFF?mt=FOUR_K_H&vendorId=example&contentId=<UUID>

The final path component is the KID, while the query parameters carry stream metadata. They must be removed before IsValidKID validation. I added the concise example next to this logic in 3ad066b.

CastagnaIT (approved):
I made some tests i think its ok
thank you

PR #2107, opened 2026-08-27
Support Kodi VFS resources

garbear:
## Description

Allow adaptive resources to be read through Kodi's VFS in addition to HTTP(S). Because I want to play `file:://` URLs.

AI use:

Extend URL resolution to support Kodi VFS schemes, including `file://` URLs and URNs (such as potential v23 `magnet:` links in https://github.com/xbmc/xbmc/pull/23696) across nested playlist, init and segment URLs.

Extend `CUrl` so HTTP(S) keeps the existing CURL path while other schemes use `kodi::vfs::CFile`. Byte ranges use the existing HTTP `Range` header for HTTP(S), or seek and bounded reads for VFS resources.

Update adaptive segment downloads to use the transport-independent range handling.

## Motivation and context

InputStream Adaptive currently opens adaptive resources through Kodi's CURL interface, which limits playback to HTTP(S)-style resources.

This allows manifests and their referenced resources to be loaded from Kodi VFS URLs such as `file://`, and enables VFS providers such as Kodi's `magnet:` filesystem to be used by InputStream Adaptive without adding transport-specific code to the add-on.

## How has this been tested?

Tested on macOS ARM64 with Kodi 22 development builds.

- Added tests for generic URI detection and nested VFS URL resolution, including preservation of magnet query identity.
- Added tests for local VFS reads, bounded and open-ended byte ranges, EOF handling, failed seeks, and preservation of the existing HTTP path.
- Added adaptive-stream tests for VFS byte ranges and stopped downloads.
- Full InputStream Adaptive test suite passes: 83/83.
- Tested end-to-end with a local CMAF/HLS presentation using `file://`.
- Kodi loaded the modified InputStream Adaptive add-on, opened the master and child playlists, initialization fragments and media segments, decoded 3840x2160 H.264 with E-AC-3 Atmos audio, and played the approximately 20-second presentation to normal EOF.

## Types of change

### DIFF
--- a/src/utils/CurlUtils.cpp
+++ b/src/utils/CurlUtils.cpp
@@ -18,6 +18,9 @@
 #include "CompKodiProps.h"
 #include "SrvBroker.h"
 
+#include <algorithm>
+#include <limits>
+
 using namespace UTILS;
 using namespace UTILS::CURL;
 
@@ -186,7 +189,11 @@ void StoreCookies(const std::string& url, const std::vector<std::string>& cookie
 } // unnamed namespace
 
 UTILS::CURL::CUrl::CUrl(const std::string& url, const RequestType reqType /* = RequestType::AUTO */)
+  : m_url(url), m_isHttp(URL::IsHttpUrl(url))
 {
+  if (!m_isHttp)
+    return;
+
   if (m_file.CURLCreate(url))
   {
     auto& kodiProps = CSrvBroker::GetKodiProps();
@@ -224,7 +231,7 @@ UTILS::CURL::CUrl::CUrl(const std::string& url, const std::string& postData) : C
 
 UTILS::CURL::CUrl::~CUrl()
 {
-  if (CSrvBroker::GetKodiProps().GetConfig().internalCookies)
+  if (m_isHttp && CSrvBroker::GetKodiProps().GetConfig().internalCookies)
     StoreCookies(GetEffectiveUrl(), GetResponseHeaders("set-cookie"));
 
   m_file.Close();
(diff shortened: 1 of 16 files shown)


src/utils/CurlUtils.cpp
```diff
@@ -224,93 +231,154 @@ UTILS::CURL::CUrl::CUrl(const std::string& url, const std::string& postData) : C
...
+  if (m_isHttp && CSrvBroker::GetKodiProps().GetConfig().internalCookies)
     StoreCookies(GetEffectiveUrl(), GetResponseHeaders("set-cookie"));
 
   m_file.Close();
 }
 
 int UTILS::CURL::CUrl::Open()
 {
-  if (!m_file.CURLOpen(ADDON_READ_NO_CACHE | ADDON_READ_NO_BUFFER))
+  if (m_isHttp)
```
CastagnaIT, line 242: I would like to have something like `m_isVfs` instead of `m_isHttp` so that its more clear that code is for VFS protocols (here and in all other code changes)

src/utils/CurlUtils.cpp
```diff
@@ -224,93 +231,154 @@ UTILS::CURL::CUrl::CUrl(const std::string& url, const std::string& postData) : C
...
   m_bytesRead += static_cast<size_t>(ret);
+  if (m_bytesRemaining)
+    *m_bytesRemaining -= bytesRead;
   return ReadStatus::CHUNK_READ;
 }
 
 double UTILS::CURL::CUrl::GetDownloadSpeed()
 {
-  return m_file.GetFileDownloadSpeed();
+  return m_isHttp ? m_file.GetFileDownloadSpeed() : 0.0;
```
CastagnaIT, line 374: Of course, if we don't have information about the download speed,
this will make it impossible to manage automatically the stream quality in "adaptive" streams
this should be documented, more likely on Wiki

It will need to be documented that also headers cannot be handled either,
they are usually important for most streaming services
I’m not familiar with VFS protocols, but i imagine that VFS sources are custom streams (?) and don’t require http headers but it’s best to point this out also this one on Wiki

garbear: I'll see what we can do about download speed. I only tested with `file://` URLs, in which case download speed isn't really an issue. So VFS support was initially for testing. But I want to use it with `magnet:?xt=urn:btih:...` URNs as well. Later I'll research this use case more, and come back with changes.

